PR #94, opened 2017-06-14
extract stash explorer

rebornix:
Move stash explorer out of Gitlens explorer.

### DIFF
--- a/src/views/gitStashExplorer.ts
+++ b/src/views/gitStashExplorer.ts
@@ -1,0 +1,43 @@
+'use strict';
+import { Event, EventEmitter, ExtensionContext, TreeDataProvider, TreeItem, Uri, window } from 'vscode';
+import { ExplorerNode, StashNode } from './gitExplorerNodes';
+import { GitService, GitUri } from '../gitService';
+import { StashCommitNode } from './stashCommitNode';
+
+export * from './gitExplorerNodes';
+
+export class GitStashExplorer implements TreeDataProvider<ExplorerNode>  {
+    private _node: ExplorerNode;
+    private _onDidChangeTreeData = new EventEmitter<ExplorerNode>();
+    public get onDidChangeTreeData(): Event<StashCommitNode> {
+        return this._onDidChangeTreeData.event;
+    }
+
+    constructor(private context: ExtensionContext, private git: GitService) {
+        const editor = window.activeTextEditor;
+
+        const uri = (editor !== undefined && editor.document !== undefined)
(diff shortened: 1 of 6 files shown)


src/views/gitStashExplorer.ts
```diff
@@ -0,0 +1,43 @@
+'use strict';
+import { Event, EventEmitter, ExtensionContext, TreeDataProvider, TreeItem, Uri, window } from 'vscode';
+import { ExplorerNode, StashNode } from './gitExplorerNodes';
+import { GitService, GitUri } from '../gitService';
+import { StashCommitNode } from './stashCommitNode';
+
+export * from './gitExplorerNodes';
+
+export class GitStashExplorer implements TreeDataProvider<ExplorerNode>  {
```
eamodio: Let's just call it StashExplorer -- I know my naming has git everywhere, but I'm going to try to move away from it eventually ;)

src/views/gitStashExplorer.ts
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    constructor(private context: ExtensionContext, private git: GitService) {
+        const editor = window.activeTextEditor;
+
+        const uri = (editor !== undefined && editor.document !== undefined)
+            ? new GitUri(editor.document.uri, { repoPath: git.repoPath, fileName: editor.document.uri.fsPath })
+            : new GitUri(Uri.file(git.repoPath), { repoPath: git.repoPath, fileName: git.repoPath });
+
+        this._node = new StashNode(uri, this.context, this.git);
```
eamodio: What do you think of keeping this in a node or just nuking the node and dropping the logic in the explorer? I keep going back and forth -- eh keeping it in the node is nice for possible reuse later in another view or whatnot

src/views/gitStashExplorer.ts
rebornix: Addressed.

src/views/gitStashExplorer.ts
rebornix: The first time I saw the code, I found that it's not intuitive enough. For example, let's say we have this Stash Explorer, the tree node can only be `StashNode` or `FileChangeNode` so a union type `StashNode | FileChangeNode` is good enough. That's also how I implemented it.

But for your case, if the hierarchy can be more than 2, your solution is definitely smarter. The leaf node is always `FileChangeNode` in Commits, Stashes, Pull Request so I'll say let's go with this way, putting things into the node and reuse them with ease.

PR #97, opened 2017-06-18
Add link to remote commit

mtsmfm:
Hi
Thank you for creating awesome extension 😄 

BTW, I think it is useful if we can open in remote by clicking commit.

screen shot 2017-06-19 at 3 56 13

What do you think?

### DIFF
--- a/src/annotations/annotations.ts
+++ b/src/annotations/annotations.ts
@@ -63,7 +63,7 @@ export class Annotations {
                 .replace(/\n/g, '  \n');
             message = `\n\n> ${message}`;
         }
-        return `\`${commit.shortSha}\` &nbsp; __${commit.author}__, ${moment(commit.date).fromNow()} &nbsp; _(${moment(commit.date).format(dateFormat)})_${message}`;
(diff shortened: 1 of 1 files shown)


eamodio:
@mtsmfm I think this is a great idea and have been thinking about something similar. The issue is that vscode doesn't support `command:` in the hovers right now. But you prompted me to open https://github.com/Microsoft/vscode/issues/29076 and send a PR https://github.com/Microsoft/vscode/pull/29082 :) 

If this gets accepted I will look into adding a similar feature -- although I would have it open the commit details quickpick rather than open in remote as that is more flexible.

Thanks again for your PR and push!

mtsmfm:
@eamodio Thanks quick response!

Hmm, in my environment (vscode Version 1.13.1), this patch works well:

lens

eamodio:
@mtsmfm wow you are right -- I'm always running the Insiders build and it looks like it only doesn't work there -- I'll dig more into it. Thanks!

eamodio:
@mtsmfm Currently it looks like this feature has been pulled because of security concerns with malicious content in hovers -- see https://github.com/Microsoft/vscode/issues/29076

I hope they can be convinced to not pull the feature completely, but instead change it to mitigate the security concerns.

eamodio:
@mtsmfm FYI, I've added this feature to GitLens v5. Thanks again for the PR, but unfortunately given the vscode API changes and the fact that I wanted the hover commands to open the details quick pick -- I'm closing this PR.

PR #102, opened 2017-06-26
Add commands for stash explorer

rebornix:
The idea is 

* Bind existing command to stash commit node
* Check if the args is commit node in Stash command, if so, directly read stash commit info from it instead of fetching that again from local
* Run command
* Trigger the update of the tree

Right now I reference types from View, which is not good. @eamodio let me know your take on this.


src/commands/stashDelete.ts
```diff
@@ -19,8 +20,16 @@ export class StashDeleteCommand extends Command {
         super(Commands.StashDelete);
     }
 
-    async execute(args: StashDeleteCommandArgs = { confirm: true }) {
+    async execute(args: StashDeleteCommandArgs | StashCommitNode = { confirm: true }) {
```
eamodio: @rebornix Can we use the new `run` method like here: https://github.com/eamodio/vscode-gitlens/blob/feature-explorer/src/commands/diffWithPrevious.ts#L25

That way the dealing with the node and refreshing it could be specific for that context?

eamodio:
@rebornix sorry for the mess I made out of this PR with all the rebasing. I've "merged" this in to the `feature-explorer` branch, but also heavily modified it to fit with the new command structure. And changed the view refreshing to be decoupled from the commands themselves -- since they could be executed in other contexts (or even update via command line changes).

Let me know what you think.

eamodio:
This has been "merged" closing...

PR #121, opened 2017-07-20
mixin state of super type

jrieken:
As discussed, this should be ok for Insiders and Stable and accommodate for our URI changes.


vvs:
I applied the patch to my installed extension (adopted to js instead of ts), and it works with the latest insiders build. Many thanks!!!

jrieken:
@vss Thanks for testing. Should also be tested with stable, which I did but you never ;-)

PR #145, opened 2017-09-16
Fix GitLab integration's multi-line selection.

AmandaCameron:
It seems GitLab uses a slightly-different syntax for multi-line selection, this fixes the gitlab provider to Do The Right Thing™

### DIFF
--- a/src/git/remotes/gitlab.ts
+++ b/src/git/remotes/gitlab.ts
@@ -1,4 +1,5 @@
 'use strict';
+import { Range } from 'vscode';
 import { GitHubService } from './github';
 
 export class GitLabService extends GitHubService {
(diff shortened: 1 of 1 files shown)


eamodio:
@AmandaCameron Thanks for the PR! Given that subtle difference, I think it would probably be best if the `GitLabService` just doesn't derive from the `GitHubService`. Would you mind making that happen in this PR (copy/paste of the rest of the functions is perfect fine)?

AmandaCameron:
I'll do the code changes tomorrow morning, as I'm getting ready to unwind for the night.

eamodio:
Awesome thank you !

eamodio:
Thank you for the PR!

PR #154, opened 2017-09-27
Add "external diff" command to SCM groups and files

ckaczor:
Adds an "External Diff" command to launch the configured git difftool for various selections of files:

- Run command directly: All modified files
- "Staged Changes" or "Changes" group context menu: All modified files in that group
- Source Control file context menu: The selected files

Can handle a file having both staged and unstaged changes - only the selected diff (or two diffs - one for each file if both are selected) will be shown.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1069,6 +1069,11 @@
                 }
             },
             {
+                "command": "gitlens.externalDiff",
+                "title": "Open Changes (with difftool)",
+                "category": "GitLens"
+            },
+            {
                 "command": "gitlens.resetSuppressedWarnings",
                 "title": "Reset Suppressed Warnings",
                 "category": "GitLens"
@@ -1520,23 +1525,33 @@
                 {
                     "command": "gitlens.openChangedFiles",
                     "when": "gitlens:enabled",
-                    "group": "1_gitlens@1"
+                    "group": "2_gitlens@1"
                 },
                 {
                     "command": "gitlens.closeUnchangedFiles",
                     "when": "gitlens:enabled",
-                    "group": "1_gitlens@2"
+                    "group": "2_gitlens@2"
+                },
+                {
+                    "command": "gitlens.externalDiff",
+                    "when": "gitlens:enabled",
+                    "group": "2_gitlens@3"
                 },
                 {
                     "command": "gitlens.stashSave",
                     "when": "gitlens:enabled",
-                    "group": "2_gitlens@1"
+                    "group": "3_gitlens@1"
                 }
             ],
             "scm/resourceState/context": [
                 {
                     "command": "gitlens.openFileInRemote",
                     "when": "gitlens:enabled && gitlens:hasRemotes",
+                    "group": "navigation"
+                },
+                {
+                    "command": "gitlens.externalDiff",
+                    "when": "gitlens:enabled",
                     "group": "navigation"
                 },
                 {
--- a/src/commands/externalDiff.ts
+++ b/src/commands/externalDiff.ts
@@ -1,0 +1,115 @@
+'use strict';
+import { commands, SourceControlResourceState, Uri, window } from 'vscode';
+import { Command, Commands } from './common';
+import { BuiltInCommands } from '../constants';
+import { CommandContext } from '../commands';
+import { GitService } from '../gitService';
+import { Logger } from '../logger';
+import { Messages } from '../messages';
+
+enum Status {
+    INDEX_MODIFIED,
+    INDEX_ADDED,
+    INDEX_DELETED,
+    INDEX_RENAMED,
+    INDEX_COPIED,
+
+    MODIFIED,
+    DELETED,
+    UNTRACKED,
+    IGNORED,
+
+    ADDED_BY_US,
+    ADDED_BY_THEM,
+    DELETED_BY_US,
+    DELETED_BY_THEM,
+    BOTH_ADDED,
+    BOTH_DELETED,
+    BOTH_MODIFIED
+}
+
+enum ResourceGroupType {
+    Merge,
+    Index,
+    WorkingTree
+}
+
+interface Resource extends SourceControlResourceState {
+    readonly resourceGroupType: ResourceGroupType;
+    readonly type: Status;
+}
+
+class ExternalDiffFile {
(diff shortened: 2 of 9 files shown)


package.json
```diff
@@ -1531,6 +1536,11 @@
                     "command": "gitlens.stashSave",
                     "when": "gitlens:enabled",
                     "group": "2_gitlens@1"
+                },
+                {
+                    "command": "gitlens.externalDiff",
+                    "when": "gitlens:enabled",
+                    "group": "3_gitlens@1"
```
eamodio: I want move the `gitlens.openChangedFiles` and `gitlens.closedUnchangedFiles` into the `2_gitlens@1` and `2_gitlens@2` groups respectively, so can you change this one to `2_gitlens@3`. Also `gitlens.stashSave` should move to `3_gitlens@1`

package.json
```diff
@@ -1553,6 +1563,11 @@
...
```
eamodio: Please move this one to `navigation`

src/git/git.ts
```diff
@@ -276,6 +276,16 @@ export class Git {
         return gitCommand({ cwd: repoPath }, ...params);
     }
 
+    static difftool_fileDiff(repoPath: string, fileName: string, staged: boolean) {
+        const params = [`difftool`, `--no-prompt`];
+        if (staged) {
```
eamodio: Is `--staged` used? I can't find any documentation for this on `difftool`

src/git/git.ts
```diff
@@ -276,6 +276,16 @@ export class Git {
         return gitCommand({ cwd: repoPath }, ...params);
     }
 
+    static difftool_fileDiff(repoPath: string, fileName: string, staged: boolean) {
+        const params = [`difftool`, `--no-prompt`];
```
eamodio: I think the file name should be prefixed with `--`

src/commands/externalDiff.ts
```diff
@@ -0,0 +1,87 @@
...
+    Index,
+    WorkingTree
+}
+
+interface Resource extends SourceControlResourceState {
+    readonly resourceGroupType: ResourceGroupType;
+}
+
+class ExternalDiffFile {
+    constructor(public uri: Uri, public staged: boolean) {
```
eamodio: If staged isn't valuable with `difftool` we should remove this

src/commands/externalDiff.ts
```diff
@@ -0,0 +1,87 @@
...
+                if (_.workTreeStatus === 'M') {
+                    files.push(new ExternalDiffFile(_.Uri, false));
+                }
+            });
+        }
+
+        return this.execute(repoPath, files);
+    }
+
+    async execute(repoPath: string, files: ExternalDiffFile[]) {
```
eamodio: Can we wrap these args into a single interface?
```ts
export interface ExternalDiffCommandArgs {
    files: ExternalDiffFile[]; // or just uris: Uri[] if staged isn't needed anymore
}

...

    async execute(args: ExternalDiffCommandArgs = {}) {
```

To make things easier to deal with when executing different commands from different contexts, I try to keep them all a single set of args.

src/commands/externalDiff.ts
```diff
@@ -0,0 +1,87 @@
...
+
+            if (staged) {
+                files = status.files.filter(_ => _.indexStatus === 'M').map(_ => new ExternalDiffFile(_.Uri, true));
+            } else {
+                files = status.files.filter(_ => _.workTreeStatus === 'M').map(_ => new ExternalDiffFile(_.Uri, false));
+            }
+        } else {
+            files = [];
+
+            status.files.forEach(_ => {
```
eamodio: Nitpick , but unless it is a simple single line `forEach`, I favor `for (const foo of foos)` 😃

src/commands/externalDiff.ts
```diff
@@ -0,0 +1,87 @@
...
+                    files.push(new ExternalDiffFile(_.Uri, false));
+                }
+            });
+        }
+
+        return this.execute(repoPath, files);
+    }
+
+    async execute(repoPath: string, files: ExternalDiffFile[]) {
+        try {
```
eamodio: Can we copy in this same code here: https://github.com/eamodio/vscode-gitlens/blob/65a3b31ca58d70efe895cf385a3facbae41af05b/src/commands/diffDirectory.ts#L23-L30

src/commands/externalDiff.ts
```diff
@@ -0,0 +1,87 @@
...
+}
+
+export class ExternalDiffCommand extends Command {
+
+    constructor(private git: GitService) {
+        super(Commands.ExternalDiff);
+    }
+
+    protected async preExecute(context: CommandContext): Promise<any> {
+        const repoPath = await this.git.getRepoPathFromUri(undefined);
```
eamodio: Can we move the repoPath work into execute? I try to keep the `preExecute` as just massaging context specific data into the proper args for `execute`

src/commands/externalDiff.ts
```diff
@@ -0,0 +1,87 @@
...
+            files = [];
+
+            context.scmResourceStates.forEach((_: Resource) => {
+                files.push(new ExternalDiffFile(_.resourceUri, _.resourceGroupType === ResourceGroupType.Index));
+            });
+
+            return this.execute(repoPath, files);
+        }
+
+        const status = await this.git.getStatusForRepo(repoPath);
```
eamodio: Possibly the same here -- maybe move this in to `execute` for when there aren't any files specified in the args

src/commands/externalDiff.ts
```diff
@@ -0,0 +1,87 @@
...
+        if (status === undefined) return window.showWarningMessage(`Unable to open changed files`);
+
+        if (context.type === 'scm-groups') {
+            const staged = context.scmResourceGroups[0].id === 'index';
```
eamodio: I think instead of using the `status` here, you can just use `context.scmResourceGroups[0].resourceStates`

ckaczor:
I think I covered all the changes and switched the branches right. GitHub wouldn't let me change my end so its still master but the pull request should be against your develop now.

eamodio:
Thank you!!!

PR #158, opened 2017-10-02
Only show remote and details command and diff message where applicable

SpaceEEC:
Only show the `Open in Remote` and `Show Commit Details` when the change is committed.
Only show a diff message when there is a previous version available.

Achieved by: 
- The removal of the `Open in Remote` and `Show Commit Details` commands for uncomitted changes.
Clicking the left one tells you that the commit couldn't be found.
Clicking the right one links you to a 404 page as `0000000000000000000000000000000000000000` is not a valid hash.
grafik

- The removal of the diff message when there is no previous commit available (e.g. the line was never changed after comitting the file).
The previous commit hash is displayed as `undefined` and clicking it is the same as clicking the right one.
Clicking on Changes errors out as there is no old hash to compare.
The diff is also not quite right, it displays the "new" one for both.
grafik
This is similar to a completely uncomitted file which displays as the following:
grafik

### DIFF
--- a/src/annotations/annotations.ts
+++ b/src/annotations/annotations.ts
@@ -53,7 +53,11 @@ export class Annotations {
         }
 
         let message = '';
+        let openInRemoteCommand = '';
+        let showCommitDetailsCommand = '';
         if (!commit.isUncommitted) {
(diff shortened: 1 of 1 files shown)


eamodio:
Thank you! I've merged this into my develop branch and this will make in into the next GitLens release

SpaceEEC:
Awesome! Since you merged this into the develop branch, should I just close this PR now?
(Just realized I targetted the wrong branch.)

eamodio:
I'm assuming once I merge back to master for release this will auto-close but if not I will do it then.

SpaceEEC:
Okay

PR #192, opened 2017-11-08
support for non https remotes

hjanuschka:
hey, thx for this awesome plugin ❤️  

we have a gitlab instance running without https -> don't ask!

here is the required change to support the http protocol (and keep the default `https`; works for all remotes) 

config sample:

```
{
    "gitlens.remotes": [{ "domain": "gitlab.xxxx.at", "type": "GitLab", "protocol": "http" }]
}
```

let me know if there is something to change/adapt.


eamodio:
@hjanuschka Thanks for the contribution, I will get this merged into GitLens 6 which is likely coming out tomorrow!

PR #231, opened 2017-12-13
Tuned blame line highlighting

notmedia:
Added small fix to invert git blame line highlighting. When you select the line GitLens highlights all commits except that which you selected. I'm just inverted this.


eamodio:
@notmedia Thanks!

Issue #258, opened 2018-01-25
Add tree-like layout to branches


Yukaii:
- GitLens Version: 7.5.6
- VSCode Version: 1.19.2
- OS Version:

It' common to recognize branch name before `/` as folder name. In SourceTree:

![screen shot 2018-01-25 at 8 49 02 am](https://user-images.githubusercontent.com/4230968/35365475-c6a1daae-01af-11e8-833b-cc23cad6654d.png)

GitLen:

![screen shot 2018-01-25 at 8 48 14 am](https://user-images.githubusercontent.com/4230968/35365483-d0549d52-01af-11e8-82a0-f2b83e8e0d5d.png)

BTW, GitLen is absolute a great work!

eamodio:
Awesome suggestion! Thanks!

Yukaii:
I can't wait to see this feature, so I just made it 😂 

![screen shot 2018-01-25 at 7 03 52 pm](https://user-images.githubusercontent.com/4230968/35385247-a38d9dde-0202-11e8-8ace-3f4abf96853e.png)

It's a prototype now, and I would submit a PR when it's ok 😸

PR #260, opened 2018-01-26
Enhance branch display

Yukaii:
This implements #258 

Add `gitlens.gitExplorer.branches.layout` config with three available option: `list`, `tree`, and `mix-tree`

`list` is the current behavior, would display all branches with full branch name:

screen shot 2018-01-26 at 11 01 05 am

`tree` organize branch name with slashes as folder:
  
 
screen shot 2018-01-29 at 9 13 26 pm

`mix-tree` sort all branches alphabetically:

screen shot 2018-01-26 at 11 14 08 am

### TODOs

- [x] Unexpected ordering with `tree` layout
- [x] Current branch within branch folder should be at first in tree layout
	- [x] and should be expanded by default
- [x] Detach state branch name shouldn't be parse as tree


Yukaii:
@eamodio I think it's ready for review 😄  Ping me if any change is needed.

eamodio:
@Yukaii Awesome -- I'm starting to look at this. But first can you adapt this to target the `develop` branch? As that branch is what will become GitLen 8 -- which will be the next release. Thanks!

Yukaii:
@eamodio I've merged the changes from develop branch 😸

package.json
```diff
@@ -426,6 +426,17 @@
...
                 },
+                "gitlens.gitExplorer.branches.layout": {
+                    "type": "string",
+                    "default": "list",
+                    "enum": [
+                        "list",
+                        "tree",
+                        "mix-tree"
+                    ],
+                    "description": "Specifies how the `Branches` view will display branches\n `list` - display all branches \n`tree` - organize branch as folder if branch name contains slashes \"\/\"\n `mix-tree` - display branch folders along with normal branch alphabetically",
```
eamodio: Any objection to getting rid of `mix-tree` and just having `tree` behave this way (i.e. sorted naturally)?

src/views/branchNode.ts
```diff
@@ -24,6 +25,19 @@ export class BranchNode extends ExplorerRefNode {
         return this.branch.name;
     }
 
+    get label(): string {
+        const branchName = this.branch.getName();
+        if (this.explorer.config.branches.layout === ExplorerBranchesLayout.List) {
+            return branchName;
+        } else {
+            return !!branchName.match(/\s/) ? branchName : this.branch.getBasename();
```
eamodio: I'm not sure I understand the logic here. Why are you checking if the name contains any whitespace?

package.json
eamodio: Or if you'd like to keep the feature, can we change it to a different setting?

Like `gitlens.gitExplorer.branches.sort` = `'alphabetic' | 'foldersFirst' | 'foldersLast'?
or
`gitlens.gitExplorer.branches.folderGrouping` = `'none' | 'start' | 'end'?

Thoughts?

src/views/branchNode.ts
Yukaii: In some case like detached git state, the folder name will not be parsed correctly. So here I see any white space as invalid folder branch name, and choose not to handle them. Same logic appears in `branchesNode.ts:38` and `remoteNode.ts:39`.

package.json
Yukaii: I think having only two options (`list` and `tree`) is fine, actually current tree sort implementation (not the `mix-tree` one) calculates more. It runs a recursive function, marking the parent folder as current (`BranchFolderNode#findCurrent`) 

If you think it's okay, I can drop those commits 😄

eamodio: Ok let's drop it and just have the 2 options. Thanks!

src/views/branchNode.ts
eamodio: Ah ok, but why is it needed? A "branch" with spaces could be handled just the same as any other without '/' characters right? Was there an explicit case you were protecting against?

eamodio: Ah -- ok, thanks

eamodio:
@Yukaii I'm trying to finalize GitLens 8 for release and as I'd like to get this in, I'm planning on just getting this in and then remove the `mix-tree` option in a follow-on commit -- sound ok?

Yukaii:
@eamodio Sure 👍

eamodio:
This has been merged into develop and as long as GitLens 8 release candidate looks good this will ship in v8.  FYI, I made a bunch of changes to the behavior here: https://github.com/eamodio/vscode-gitlens/commit/112879dbc20c90c2fd89e5cb504742b699084572

I removed the `mix-tree` option and changed `tree` sorting to be similar to what `mix-tree` was. Although I pulled the current branch out of any nesting. Since I think it is important to keep the current branch at the top and stable.

I removed the folder icons in favor of using a new feature in vscode 1.20 (using the File Icon Theme). Unfortunately for vscode 1.19 there will be no icons at all, but since 1.20 is supposed to ship this week, I don't think it matter much.

I also added the new layout option to the new GitLens Settings page.

Thanks again for your awesome work! This is a great feature!

PR #278, opened 2018-02-08
Add new command to diff HEAD with branch using command palette

g3offrey:
Hi 👋 
First thank you for your awesome plugin, I felt in love with it ❤️ 

However, these days something tickled me a little bit. I thought the workflow to compare HEAD with another branch (like comparing feature branches with master for checking colleagues Pull Requests) was a bit unfriendly (especially for people without the GitLens panel in explorer) 😅 

I thought something like _Cmd + P > Compare HEAD with ... (master)_ would be very cool.
You already made that but to use yours we need to have an external difftool configured 😢 
That's sad because you already made one awesome difftool included in Code to compare branches : I name it : "GitLens result panel".

That's it ! Just have to mix **Command Palette** and **GitLens result panel** to have an awesome workflow. 
I tried to cook something on a branch, that's works very well. It would be great to share this with everyone don't you think ? 👍 

## Tech Part
I decided to create a new command not to modify directly `diffDirectory.ts` to avoid "spaghetti code". However I'm not a **genius**, neither the **owner of GitLens** (are these synonym ? 😄) so if you prefer another approach I'll be glad to modify my branch to satisfy your needs. 👍 
Just tell me what to modify to make this better than it is 🥇 
Once you ok with this feature, I'll update the README accordingly 😀

Have a nice day !

### DIFF
--- a/src/commands/diffHeadWithBranch.ts
+++ b/src/commands/diffHeadWithBranch.ts
@@ -1,0 +1,54 @@
+'use strict';
+import { CancellationTokenSource, TextEditor, Uri, window } from 'vscode';
+import { ActiveEditorCommand, Commands, getCommandUri } from './common';
+import { GlyphChars } from '../constants';
+import { Container } from '../container';
+import { Logger } from '../logger';
+import { Messages } from '../messages';
+import { BranchesAndTagsQuickPick, CommandQuickPickItem } from '../quickPicks';
+
+export class DiffHeadWithBranchCommand extends ActiveEditorCommand {
+
+    constructor() {
+        super([Commands.DiffHeadWithBranch]);
+    }
+
+    async execute(editor?: TextEditor, uri?: Uri): Promise<any> {
+        uri = getCommandUri(uri, editor);
+
+        let progressCancellation: CancellationTokenSource | undefined;
+
+        try {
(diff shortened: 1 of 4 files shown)


g3offrey:
Here's the workflow I got with my branch 😉 
I'm French that's why my editor is in 🇫🇷 

gitlens-workflow-compare

src/commands/diffHeadWithBranch.ts
```diff
@@ -0,0 +1,54 @@
...
+        super([Commands.DiffHeadWithBranch]);
+    }
+
+    async execute(editor?: TextEditor, uri?: Uri): Promise<any> {
+        uri = getCommandUri(uri, editor);
+
+        let progressCancellation: CancellationTokenSource | undefined;
+
+        try {
+            const repoPath = await Container.git.getRepoPath(uri);
```
g3offrey: There is some duplication right here with `diffDirectory.ts` (to open the modal to select a branch). 
Maybe we can do something here to mutualise code ?

eamodio:
This is awesome! Thank you for your contribution!

g3offrey:
It was a pleasure 😄 
Thank you so much for your time and for making an extension so gooood 👍

PR #390, opened 2018-06-01
add another unknownRevision error to detect

ryenus:
The following error also caused blame not working:

> git rev-parse --symbolic-full-name my_branch@{u}
fatal: upstream branch 'refs/heads/my_branch' not stored as a remote-tracking branch

Git version: 2.17.1


ryenus:
Tested locally and it works, with this change the inline blame and comparing with previous revision works as expected.

But I wonder if it's possible to skip resolving the remote tracking branch? Is it possible to add an option to disable this behavior?

eamodio:
Thank you for your contribution! 👍 

As for skipping, I use the information about the remote branch to enable/disable other GitLens features (and it is only done once), so I don't see a compelling reason to have an option to disable it.

PR #393, opened 2018-06-04
Fixes #392 - unable to contribute if default script shell is sh

### LINKED ISSUE #392: [contributing] "pushd: not found" with npm install
- GitLens Version: 8.3.3
- VSCode Version: /
- OS Version: Ubuntu 16.04

Steps to Reproduce:

1. clone the repo
2. npm i
```
> gitlens@8.3.3 postinstall /home/cedric/Projects/vscode-gitlens
> node ./node_modules/vscode/bin/install && pushd "./src/ui" && npm install --no-save && popd

Detected VS Code engine version: ^1.23.0
Found minimal version that qualifies engine range: 1.23.0
Fetching vscode.d.ts from: https://raw.githubusercontent.com/Microsoft/vscode/7c7da59c2333a1306c41e6e7b68d7f0caa7b3d45/src/vs/vscode.d.ts
vscode.d.ts successfully installed!

sh: 1: pushd: not found
```

Looks like it has something to do with this : https://stackoverflow.com/questions/45630223/how-to-change-shell-of-npm-install
But I don't really see the point of doing this.

Would it be ok to replace `pushd / popd` by `cd / cd -` ? It should have the same behavior and increase the compatibility with more systems.

cmalard:


eamodio:
Thank you for the contribution, but unfortunately I can't accept this as this doesn't work on Windows (my primary dev environment).

cmalard:
I replaced `cd -` by `cd ../..` (with all those stuff that should make it compatible with Windows (`cd \"../..\"`), could you give it a try ? :thinking:

eamodio:
Yup that works fine, sucks that `pushd`/`popd` isn't supported everywhere

PR #416, opened 2018-06-14
Fixes #415 - branch names mangled by color escapes

### LINKED ISSUE #415: branch names mangled by color escapes
- GitLens Version: 8.3
- VSCode Version: 1.24.0
- OS Version: Microsoft Windows [Version 10.0.16299.431]

## Description

branch names are displayed in mangled form with missing associated commits when git config contains "color.branch=always".

## Steps to Reproduce

1. enable branch colors on shell command line:

    ```
    git config --global color.branch always
    ```

1.  branch names displayed within the "Branches" section of the Gitlens explorer will contain color escape codes, eg, "[32mmaster[m", and will not show any contained commits.

## Fix

Using the git global option "-c color.branch=never" will definitively disable color for the `git branch` command, fixing both issues.

Addressing possible concerns of generality... by researching the `git` online documentation and Wikipedia, the "-c ..." option appears to have been introduced somewhere between the 1.7.0 - 1.7.12.4 versions (being released in the Feb-2010 to Oct-2012 time frame).

ref: https://en.wikipedia.org/wiki/Git#Releases
ref: https://git-scm.com/docs/git/1.7.0
ref: https://git-scm.com/docs/git/1.7.12.4

rivy:
- add "-c color.branch=never" option to git branch commands, definitively disabling color branch output

.# Discussion

Branch names are mis-parsed if `git branch` output is colored (eg, when git is configured
with "config.branch=always"). Using the base "-c color.branch=never" option overrides any
user configuration, displaying git branch output as uncolored in all cases and fixing the
issue.

Notably, option ordering is important in this instance. The "-c ..." option is a git
base option, not a git branch subcommand option. So, as done here, it must be inserted
before the "branch" subcommand.

ref: Issue #415

### DIFF
--- a/src/git/git.ts
+++ b/src/git/git.ts
@@ -359,7 +359,7 @@ export class Git {
     }
 
     static diff(repoPath: string, fileName: string, sha1?: string, sha2?: string, options: { encoding?: string } = {}) {
-        const params = ['diff', '--diff-filter=M', '-M', '--no-ext-diff', '--minimal'];
+        const params = ['-c', 'color.diff=false', 'diff', '--diff-filter=M', '-M', '--no-ext-diff', '--minimal'];
         if (sha1) {
             params.push(Git.isStagedUncommitted(sha1) ? '--staged' : sha1);
         }
@@ -372,7 +372,7 @@ export class Git {
     }
 
     static diff_nameStatus(repoPath: string, sha1?: string, sha2?: string, options: { filter?: string } = {}) {
-        const params = ['diff', '--name-status', '-M', '--no-ext-diff'];
+        const params = ['-c', 'color.diff=false', 'diff', '--name-status', '-M', '--no-ext-diff'];
         if (options && options.filter) {
             params.push(`--diff-filter=${options.filter}`);
         }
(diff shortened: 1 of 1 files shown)


eamodio:
Thanks for your contribution!

I'm surprised setting `color.ui=false` isn't enough as it is supposed to be the master switch:
https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration#_code_color_ui_code

Here is the relevant code
https://github.com/eamodio/vscode-gitlens/blob/develop/src/git/git.ts#L102

And if `color.ui` is not reliable, we will probably need to add the specific overrides to all the commands, branch, diff, status, etc

rivy:
Yes, I was surprised that "color.ui=false" doesn't turn off all color output, as well. It may well be that it actually does turn off all color, but then is reversed by more specific overrides (such as "color.branch=always") at other configuration levels.

I, personally, have "color.branch=always", "color.diff=always", and "color.status=always" set at the global level, so that I can, by default, get colored output even for redirections (eg, to `less`). This does cause parsing confusion for code instrumenting `git` and expecting non-colored output (eg, https://github.com/jisaacks/GitGutter/pull/475).

Although this is the only issue that I'm encountering here, I do think that, ultimately, for robustness, all the verb specific color settings (branch, diff, interactive, status) should be set to "false" or "never" when parsing the `git` output of those verbs.

rivy:
Done.

I found three uses of 'diff' and two uses of 'status', and I modified the config options of all of them. All are using "false". But, FYI, I only looked for uses within "src/git/git.ts". It appeared that that was where everything was localized. Let me know if there are other areas to change.

The final result is working correctly for me.

eamodio:
Looks great! Thank you!

PR #441, opened 2018-07-09
Add the gitlens.hovers.avatars configuration option

segevfiner:
This setting allows disabling/enabling showing avatars in hovers. We already have settings for the explorer and blame gutter, so it makes sense to also have one for the hovers.


eamodio:
Awesome -- thank you for the contribution! It looks great!

PR #445, opened 2018-07-16
Replace `--` with `—` in package.json

korigod:
Double hyphen is a poor em dash replacement in the monospace font. `--` were replaced with `—` in the configuration properties descriptions.


eamodio:
Thanks!

Issue #457, opened 2018-07-23
Displays the wrong username (You)


ashley:
If a contributor in a Git repository has the same first name as me, GitLens will display "You" as the author. The links and the avatar are for the right user, but when hovering over a LOC, it'll say something like, "You, a x time ago".

 GitLens Version:
- VSCode Version: 8.5.2
- OS Version: MacOS High Sierra 10.13.4

Steps to Reproduce:
1. Go to a repo that includes a contributor with the same first name as you.
2. Go to a line of code which that contributor committed.

speed-oss:
@ashley if this was helpful to you, consider giving a reward to the issue solver using [Speedoss.com](speedoss.com)

PR #460, opened 2018-07-24
Fix incorrect author display `You`(#457)

qzyse2017:
add email for further check

### DIFF
--- a/src/git/parsers/blameParser.ts
+++ b/src/git/parsers/blameParser.ts
@@ -163,8 +163,12 @@ export class GitBlameParser {
         let commit = commits.get(entry.sha);
         if (commit === undefined) {
             if (entry.author !== undefined) {
-                if (currentUser !== undefined &&
-                    currentUser.name === entry.author && currentUser.email === entry.authorEmail) {
+                if (
+                    currentUser !== undefined &&
+                    (currentUser.name !== undefined || currentUser.email !== undefined) &&
+                    (currentUser.name === undefined || currentUser.name === entry.author) &&
+                    (currentUser.email === undefined || currentUser.email === entry.authorEmail)
+                ) {
                     entry.author = 'You';
                 }
                 let author = authors.get(entry.author);
--- a/src/git/parsers/blameParser.ts
+++ b/src/git/parsers/blameParser.ts
@@ -28,7 +28,8 @@ export class GitBlameParser {
         data: string,
         repoPath: string | undefined,
         fileName: string,
-        currentUser: string | undefined
+        currentUser: string | undefined,
+        currentUserEmail: string | undefined
     ): GitBlame | undefined {
         if (!data) return undefined;
 
@@ -69,7 +70,8 @@ export class GitBlameParser {
                             .slice(1)
                             .join(' ')
                             .trim();
-                        if (currentUser !== undefined && currentUser === entry.author) {
+                        if (currentUser !== undefined && currentUser === entry.author &&
+                            currentUserEmail !== undefined && currentUserEmail === entry.authorEmail) {
                             entry.author = 'You';
                         }
                     }
--- a/src/git/parsers/blameParser.ts
+++ b/src/git/parsers/blameParser.ts
@@ -28,8 +28,7 @@ export class GitBlameParser {
         data: string,
         repoPath: string | undefined,
         fileName: string,
-        currentUser: string | undefined,
-        currentUserEmail: string | undefined
+        currentUser: { name?: string, email?: string } | undefined
     ): GitBlame | undefined {
         if (!data) return undefined;
 
@@ -70,10 +69,6 @@ export class GitBlameParser {
                             .slice(1)
                             .join(' ')
                             .trim();
-                        if (currentUser !== undefined && currentUser === entry.author &&
-                            currentUserEmail !== undefined && currentUserEmail === entry.authorEmail) {
-                            entry.author = 'You';
-                        }
                     }
                     break;
 
@@ -127,7 +122,7 @@ export class GitBlameParser {
                     }
                     first = false;
 
-                    GitBlameParser.parseEntry(entry, repoPath, relativeFileName, commits, authors, lines);
+                    GitBlameParser.parseEntry(entry, repoPath, relativeFileName, commits, authors, lines, currentUser);
 
                     entry = undefined;
                     break;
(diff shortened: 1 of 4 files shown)


src/git/parsers/blameParser.ts
```diff
@@ -28,7 +28,8 @@ export class GitBlameParser {
         data: string,
         repoPath: string | undefined,
         fileName: string,
-        currentUser: string | undefined
+        currentUser: string | undefined,
```
eamodio: Rather than separating `currentUser` and `currentEmail`, let's combine them into `currentUser: { name?: string; email?: string } | undefined`

src/git/parsers/blameParser.ts
```diff
@@ -69,7 +70,8 @@ export class GitBlameParser {
                             .slice(1)
                             .join(' ')
                             .trim();
-                        if (currentUser !== undefined && currentUser === entry.author) {
+                        if (currentUser !== undefined && currentUser === entry.author &&
```
eamodio: Rather than have these checks here, I think they should be moved to be at the end of the entry parsing -- probably in `parseEntry` https://github.com/eamodio/vscode-gitlens/blob/develop/src/git/parsers/blameParser.ts#L157

That way we don't need to have the check in 2 places, since we don't want to rely on the email being parsed before the author (which should actually be the other way around given the current structure)

src/git/parsers/blameParser.ts
eamodio: This applies to the other parsers as well

src/git/parsers/blameParser.ts
eamodio: This applies to the other parsers as well

src/gitService.ts
```diff
@@ -935,6 +937,17 @@ export class GitService extends Disposable {
         }
         return user;
     }
+    // TODO: Clear cache when git config changes
+    private _userEmailMapCache: Map<string, string | undefined> = new Map();
+
+    async getCurrentUserEmail(repoPath: string) {
```
eamodio: Instead of adding `getCurrentUserEmail`, let's rename `getCurrentUsername` to `getCurrentUser` and have it return `{ name?: string; email?: string } | undefined`. Then we can share the same cache too

eamodio: Probably would also be best to add a new method in `git.ts`:
```ts
    static async config_getRegex(pattern: string, repoPath?: string) {
        try {
            const data = await gitCommandCore({ cwd: repoPath || '' }, 'config', '--get-regex', pattern);
            return data.trim();
        }
        catch {
            return undefined;
        }
    }
```

That new method can be called from `getCurrentUser` using the pattern `user.(name|email)`.
It will then return output that looks like:
```
> git config --get-regex "user.(name|email)"
user.name Foo Bar
user.email foo@bar.com
```
which could then be easily parsed and returned

qzyse2017: It seems the returned results in ```config_getRegex``` is got by executing a command in command line and returning its results, but I tried in a test repo, and when the repo's username or email is reset, it may return multiple lines and it will make the judegment rather ambiguous. It seems using ```config_get``` would be a better choice

qzyse2017:
image
image
these screenshots showed the conditions talked above
the first two config is global, and the last ones are local to the test repo

src/git/parsers/blameParser.ts
```diff
@@ -160,11 +157,16 @@ export class GitBlameParser {
...
         commits: Map<string, GitBlameCommit>,
         authors: Map<string, GitAuthor>,
-        lines: GitCommitLine[]
+        lines: GitCommitLine[],
+        currentUser: {name?: string, email?: string} | undefined
     ) {
         let commit = commits.get(entry.sha);
         if (commit === undefined) {
             if (entry.author !== undefined) {
+                if (currentUser !== undefined &&
```
eamodio: I think this `if` should look like
```ts
if (
    currentUser !== undefined &&
    // Name or e-mail is configured
    (currentUser.name !== undefined || currentUser.email !== undefined) &&
    // Match on name if configured
    (currentUser.name === undefined || currentUser.name === entry.author) &&
    // Match on email if configured
    (currentUser.email === undefined || currentUser.email === entry.authorEmail)
) {
    entry.author = 'You';
}
```

src/git/parsers/logParser.ts
```diff
@@ -238,10 +236,15 @@ export class GitLogParser {
         relativeFileName: string,
         commits: Map<string, GitLogCommit>,
         authors: Map<string, GitAuthor>,
-        recentCommit: GitLogCommit | undefined
+        recentCommit: GitLogCommit | undefined,
+        currentUser: {name?: string, email?: string} | undefined
     ): GitLogCommit | undefined {
         if (commit === undefined) {
             if (entry.author !== undefined) {
+                if (currentUser !== undefined &&
```
eamodio: Same here
```ts
if (
    currentUser !== undefined &&
    // Name or e-mail is configured
    (currentUser.name !== undefined || currentUser.email !== undefined) &&
    // Match on name if configured
    (currentUser.name === undefined || currentUser.name === entry.author) &&
    // Match on email if configured
    (currentUser.email === undefined || currentUser.email === entry.authorEmail)
) {
    entry.author = 'You';
}
```

src/gitService.ts
```diff
@@ -925,13 +925,17 @@ export class GitService extends Disposable {
     }
 
     // TODO: Clear cache when git config changes
-    private _userNameMapCache: Map<string, string | undefined> = new Map();
+    private _userMapCache: Map<string, { name: string, email: string } | undefined> = new Map();
```
eamodio: I think this should be adapted like so
```ts
    private _userMapCache = new Map<string, { name?: string; email?: string } | null>();

    async getCurrentUser(repoPath: string) {
        let user = this._userMapCache.get(repoPath);
        if (user != null) return user;
        // If we found the repo, but no user data was found just return
        if (user === null) return undefined;

        const data = await Git.config_getRegex('user.(name|email)', repoPath);
        if (!data) {
            // If we found no user data, mark it so we won't bother trying again
            this._userMapCache.set(repoPath, null);
            return undefined;
        }

        user = { name: undefined, email: undefined };

        let match: RegExpExecArray | null = null;
        do {
            match = userConfigRegex.exec(data);
            if (match == null) break;

            // Stops excessive memory usage -- https://bugs.chromium.org/p/v8/issues/detail?id=2869
            user[match[1] as 'name' | 'email'] = (' ' + match[2]).substr(1);
        } while (match != null);

        this._userMapCache.set(repoPath, user);
        return user;
    }
```

Allows user name or email to be missing, and avoids calling config each time if we don't find data. It also allows for the last entry to win (since that is "closest" to the repo)

eamodio: The regex
```ts
const userConfigRegex = /^user\.(name|email) (.*)$/gm;
```

Can be placed on line 70 in `gitService.ts`

eamodio:
As for your question about clearing the cache, I think the easiest solution is to add
```ts
        if (reason === RepositoryChange.Config) {
            this._userMapCache.delete(repo.path);
        }
```

Here: https://github.com/eamodio/vscode-gitlens/blob/develop/src/gitService.ts#L128

That should take care of any repo-level changes to the config. Now if the user changes the global config all bets are off -- but I'm ok not bothering with that at this point 😄

eamodio:
Almost there! Looking forward to getting this merged in!

PR #465, opened 2018-07-26
Fix incorrect author display

qzyse2017:
tested it manually and it seemed to perform correctly.  Removed some simple comments from your suggested code to follow your original style which seems to have very few comments, thanks again for your great suggestions !


qzyse2017:
sorry I chosed a incorrect base😥

PR #579, opened 2018-12-06
Gravartars not showing if email contained upper case characters

sgtwilko:
The email address was being correctly manipulated for the lookup of the Gravatar from local cache, but when the call to Gravatar was made for one that wasn't in the local cache the email was not being run via `trim()` and `toLowerCase()`.

This has been causing the avatars to fail to show for valid accounts.


eamodio:
Thank you!

Issue #626, opened 2019-01-14
Branch names with only digits always appear first


mlasson:
Issue Type: <b>Bug</b>

To reproduce, make a branch with a name consisting only of digits (eg. "123") in a repository containing a starred/favorite branch (eg. "master"). Normally, you'll expect to have master appearing first in the list because it is starred but you'll see that the "123" is first:

![image](https://user-images.githubusercontent.com/5814258/51110376-cbcc4180-17f8-11e9-82a9-c47e20828ace.png)

Extension version: 9.4.1
VS Code version: Code - Insiders 1.31.0-insider (e02e44464c02d617b98d0a3265bf5efef7794436, 2018-12-21T06:20:35.852Z)
OS version: Windows_NT x64 10.0.17134

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|Intel(R) Core(TM) i7-4790 CPU @ 3.60GHz (8 x 3592)|
|GPU Status|2d_canvas: enabled<br>checker_imaging: disabled_off<br>flash_3d: enabled<br>flash_stage3d: enabled<br>flash_stage3d_baseline: enabled<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>native_gpu_memory_buffers: disabled_software<br>rasterization: enabled<br>surface_synchronization: enabled_on<br>video_decode: enabled<br>webgl: enabled<br>webgl2: enabled|
|Memory (System)|15.89GB (6.76GB free)|
|Process Argv||
|Screen Reader|no|
|VM|0%|

</details>

speed-oss:
@mlasson if this was helpful to you, consider giving a reward to the issue solver using [Speedoss.com](speedoss.com)

PR #627, opened 2019-01-14
Avoid using objects to implement Map

mlasson:
This commit occurring with branch names which are parsable integers. 

They were always shown first in the list of branches (even if they are
not starred) because the chrome implementation does not respect the
insertion order for keys that happen to be parseable integers.

It fixes Issue #626. 

# Checklist

- [X] My changes are based off of the `develop` branch
- [X] My changes include any required corresponding changes to the documentation

### DIFF
--- a/src/system/array.ts
+++ b/src/system/array.ts
@@ -85,7 +85,7 @@ export namespace Arrays {
         value?: T;
 
         // parent?: IHierarchicalItem<T>;
-        children: { [key: string]: IHierarchicalItem<T> } | undefined;
+        children: Map<string, IHierarchicalItem<T>> | undefined;
         descendants: T[] | undefined;
     }
 
@@ -98,7 +98,7 @@ export namespace Arrays {
         const seed = {
             name: '',
             relativePath: '',
-            children: Object.create(null),
+            children: new Map(),
             descendants: []
         };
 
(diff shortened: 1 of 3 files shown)


src/system/array.ts
```diff
@@ -85,7 +85,7 @@ export namespace Arrays {
         value?: T;
 
         // parent?: IHierarchicalItem<T>;
-        children: { [key: string]: IHierarchicalItem<T> } | undefined;
+        children: Map<string, IHierarchicalItem<T>> | undefined;
```
mlasson: I kept the "`| undefined`" here, but it seems to me that we could rewrite the code easily to use an empty map instead (or an empty object before). Same for `descendants`.
Is there a good reason ?

mlasson:
Note that running "prettier" induce some modification on files that appear to be unrelated to this PR:
```
     modified:   CHANGELOG.md
     modified:   src/annotations/blameAnnotationProvider.ts
     modified:   src/commands/diffWithRevision.ts
     modified:   src/commands/openFileRevision.ts
```
Same for lint, that returns the following warning: 
```
WARNING: C:/Users/mlasson/Documents/Git/vscode-gitlens/src/vsls/host.ts:78:9 - Promises must be handled appropriately
```

src/system/array.ts
eamodio: Mainly for performance, to avoid creating objects that will need to be GC'd that we won't need.

eamodio:
@mlasson Awesome work! Thanks so much for your contribution! This will be in the next release!

mlasson:
@eamodio: Thank you !

PR #781, opened 2019-07-05
Fix bug when repoPath & fileName differ in suffix

ampresent:
# Description

This bug can be simply reproduced :
"/x" is a symlink to "/y",  repoPath is "/y" and fileName is somehow /x/file1.
Joining them will give the wrong path "/y/x/file1".

# Checklist

- [x] My changes are based off of the `develop` branch

### DIFF
--- a/src/git/gitUri.ts
+++ b/src/git/gitUri.ts
@@ -1,5 +1,6 @@
 'use strict';
 /* eslint-disable constructor-super */
+import * as url from 'url';
 import * as paths from 'path';
 import { Uri } from 'vscode';
 import { UriComparer } from '../comparers';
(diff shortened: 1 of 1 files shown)


src/git/gitUri.ts
```diff
@@ -365,7 +366,7 @@ export class GitUri extends ((Uri as any) as UriEx) {
 
         if (normalizedFileName.startsWith(normalizedRepoPath)) return normalizedFileName;
 
-        return Strings.normalizePath(paths.join(normalizedRepoPath, normalizedFileName));
+        return Strings.normalizePath(url.resolve(repoPath, fileName));
```
eamodio: Need to ensure that the paths all have the same `/` slashes

```suggestion
        return Strings.normalizePath(url.resolve(normalizedRepoPath, normalizedFileName));
```

eamodio:
@wuyihao14 Thank you so much for your contribution, and I'm really sorry for letting this sit for so long. The reason is that change feels quite risky given how it will affect a lot of the Urls in GitLens. Can you provide more details on the scenario this fixes and what testing you have done to verify that this won't break things -- e.g. paths with `\` rather than `/`, what if the repoPath doesn't end with a `/` (it looks like `url.resolve` will end up replacing it, etc. Thank you!

PR #1128, opened 2020-10-14
Support language pack

Jotting89:


eamodio:
@Jotting89 Unfortunately from me switching to `main` as the default branch and nuking `master` it closed this PR. Can you re-open it against `main` please?

Also, while this is a great start, there are unfortunately a LOT of strings internally to GitLens that this doesn't yet cover.

PR #1171, opened 2020-11-18
Closes #1165 - Adds options to modify file annotation button behaviour

### LINKED ISSUE #1165: Is there any way to set the default behavior of the File Annotations button?
I'd like to do Toggle File Blame with a single click, like the previous version do.

But in this version, I've to select a toggle mode in each click.

image

Is there any way to set the default behavior of that button?

arrpee:
# Description

Adds two options to modify the behaviour of the file annotations button:
- "gitlens.menus.annotations.behaviour" changes the behaviour between a menu (default) and a toggle button
- "gitlens.menus.annotations.toggleType" changes what is toggled when behaviour is set to toggle

I wasn't too sure of what to name the settings so if you can suggest more appropriate ones that would be great.

# Checklist

- [X] My changes include any required corresponding changes to the documentation
- [X] My changes have a descriptive commit message with a short title, including a `Fixes $XXX -` or `Closes #XXX -` prefix to auto-close the issue that your PR addresses

### DIFF
--- a/README.md
+++ b/README.md
@@ -835,6 +835,12 @@ See also [View Settings](#view-settings- 'Jump to the View settings')
 | `gitlens.heatmap.hotColor`     | Specifies the base color of the gutter heatmap annotations when the most recent change is newer (hot) than the `gitlens.heatmap.ageThreshold` value                                                             |
 | `gitlens.heatmap.toggleMode`   | Specifies how the gutter heatmap annotations will be toggled<br /><br />`file` - toggles each file individually<br />`window` - toggles the window, i.e. all files at once                                      |
 
+## File Annotations Settings [#](#file-annotation-settings- 'File Changes Settings')
+
+| Name                           | Description                                                                                                                                                                                                     |
+| ------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------- |
+| `gitlens.fileAnnotations.command` | Specifies whether the file annotations button in the editor title toggles blame, heatmap, or change annotations. Setting this to `null` (default) makes the button act as a menu |
+
 ## Git Command Palette Settings [#](#git-command-palette-settings- 'Git Command Palette Settings')
 
 | Name                                              | Description                                                                                                                           |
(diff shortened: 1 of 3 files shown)


MFredrich (approved):

eamodio:
Thanks for the contribution! Unfortunately we can't use the `gitlens.menus` setting as you have here, since it will break how that setting is used elsewhere (since its set as an object). So we'll need to come up with another setting name -- maybe add a new `fileAnnotations.command` setting, that defaults to `null` (which would mean show the menu), or be an enum of `blame`, `changes`, `heatmap`. We probably will also need to add a `fileAnnotations.diffCommand` setting, which is the same as  `fileAnnotations.command`, but can't be set to `changes`

eamodio (changes requested):
Thanks for the contribution! 

Unfortunately we can't use the `gitlens.menus` setting as you have here, since it will break how that setting is used elsewhere (since its set as an object). So we'll need to come up with another setting name -- maybe add a new `fileAnnotations.command` setting, that defaults to `null` (which would mean show the menu), or be an enum of `blame`, `changes`, `heatmap`. We probably will also need to add a `fileAnnotations.diffCommand` setting, which is the same as `fileAnnotations.command`, but can't be set to `changes`.

arrpee:
I've ocombined and replaced the settings with `fileAnnotations.command`. What would `fileAnnotations.diffCommand` be used for?

eamodio:
@arrpee Thanks! Looking great!

I currently have the _Changes_ file annotations disabled when you are in a diff editor -- but thinking about that more I don't think that is needed anymore. So I think we can just move ahead without anything like `fileAnnotations.diffCommand`

README.md
```diff
@@ -857,9 +857,11 @@ See also [View Settings](#view-settings- 'Jump to the View settings')
 
 ## Menu & Toolbar Settings [#](#menu--toolbar-settings- 'Menu & Toolbar Settings')
 
-| Name            | Description                                           |
-| --------------- | ----------------------------------------------------- |
-| `gitlens.menus` | Specifies which commands will be added to which menus |
+| Name                                   | Description                                                                                                          |
+| -------------------------------------- | -------------------------------------------------------------------------------------------------------------------- |
+| `gitlens.menus`                        | Specifies which commands will be added to which menus                                                                |
+| `gitlens.menus.annotations.behaviour`  | Specifies whether the file annotations button in the editor title acts as a menu (default) or a toggle button        |
```
eamodio, line 863: Just need to remove this and document the new `gitlens.fileAnnotations.command`

arrpee: I've made the changes

eamodio:

Thank you!

PR #1307, opened 2021-01-05
Adds Feature for Copy Current Branch Name

kh0m:
# Description

### Summary of Changes
- Registers a command to bind a command id under the namespace `gitlens.views.branches`
- Adds a command contribution to make it available in the Command Palette
- Registers an onCommand activationEvent
- Uses GitService and env to copy the name to clipboard

### Background and Motivation
I love using GitLens and it works great when managing multiple repositories and multiple branches. Given the large number of branches in my repos, I would find it extremely convenient to have a command from the Command Palette that copies the name of the currently checked out branch. I do this so many times in a day in order to paste it in trackers and pull requests.

While this can be done using the copy command in the branchesView, it will save having to open that view and scroll through the hundreds of branches to find the one with the checkmark next to it.

### Related Issue: https://github.com/eamodio/vscode-gitlens/issues/1306

# Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -141,6 +141,7 @@
 		"onCommand:gitlens.addAuthors",
 		"onCommand:gitlens.connectRemoteProvider",
 		"onCommand:gitlens.disconnectRemoteProvider",
+		"oncommand:gitlens.copyCurrentBranch",
 		"onCommand:gitlens.copyMessageToClipboard",
 		"onCommand:gitlens.copyShaToClipboard",
 		"onCommand:gitlens.closeUnchangedFiles",
@@ -185,7 +186,6 @@
 		"onCommand:gitlens.fetchRepositories",
 		"onCommand:gitlens.pullRepositories",
 		"onCommand:gitlens.pushRepositories",
-		"oncommand:gitlens.copyCurrentBranch",
 		"onStartupFinished"
 	],
 	"contributes": {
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -4658,6 +4659,11 @@
 				"command": "gitlens.views.tags.setShowAvatarsOff",
 				"title": "Hide Avatars",
 				"category": "GitLens"
+			},
+			{
+				"command": "gitlens.views.branches.copyCurrentBranchName",
```
eamodio, line 4664: I would recommend changing this to `gitlens.copyCurrentBranch` -- since this shouldn't live inside the views, it is a global command in the command palette.

src/views/branchesView.ts
```diff
@@ -192,6 +194,11 @@ export class BranchesView extends ViewBase<BranchesViewNode, BranchesViewConfig>
 			() => this.setShowBranchPullRequest(false),
 			this,
 		);
+		commands.registerCommand(
+			this.getQualifiedCommand('copyCurrentBranchName'),
```
eamodio, line 198: This should live as a global command rather than inside this view (especially because this won't even get run until the view is visible for the first time).

I would recommend following this pattern: https://github.com/eamodio/vscode-gitlens/blob/main/src/commands/openBranchOnRemote.ts

You can delete the `preExecute` stuff. And like that command, you have to account for there being more than 1 repository open at a time (so the `getRepoPathOrActiveOrPrompt` call will get you the "best" repository.

eamodio (changes requested):
Thanks for the contribution! Although we will need to make some changes before this can be merged in. 😄

package.json
kh0m: 👍 Thanks, that's a good point--I agree this should be global. Fixed the naming to `gitlens.copyCurrentBranch`.

src/views/branchesView.ts
kh0m: Thanks for the guidance on this! `getRepoPathOrActiveOrPrompt` is working much better and I can now select the repository I want when there is no active editor.

eamodio:
I made some minor changes and merged it. Thank you so much for your contribution!

Thank you!

kh0m:
Wow! Thank _you_ for accepting the change and for doing it so quickly too! I've verified it on GitLens insiders--it's so cool to see it on there! 👍

PR #1342, opened 2021-01-24
Fixes #1335 - improve contributors view. Order by recent commit and use all refs

### LINKED ISSUE #1335: Contributors view doesn't update after fetch
- GitLens Version: 11.1.3
- Git Version: Run `git --version` 2.14.1

- VSCode Version: Run `code --version`: 1.53
- OS Version: Windows 10

In this image I'm using Git Graph on the right and GitLens on the left. Git Graph seems to use data from the upstream I fetched. However on the left contributors is using my local master (which is a few months behind). The left shows Andrew Clark's last commit as 4 months ago but on the right he committed today (19th Jan) 

image

Now that autofetch [fetches all remotes](https://github.com/microsoft/vscode/issues/110811) routinely is it possible for contributors to reflect the latest changes remotely, without me needing to pull? Similar to how Git Graph is working.
This would be useful to see what changes contributors have added without needing to switch branches to master then pulling from upstream.

jasonwilliams:
## Ordering
Currently the contributors view hasn’t been very useful. The ordering was the most commits at the top. The problem that brings is you usually get people who don’t work on the project anymore at the top of the list and those you want to collaborate with half way down. This PR changes the ordering by most recent commit.

There is a limit on commits searched (the last 200). So this may show less contributors than before. A setting could be made to adjust this.

## Search all refs
The contributors view only checked the active remote. So if you added someone’s remote (like in the github pull request extension when you checkout someone’s PR) it wasn’t reflected in the contributors view. This PR fixes that by searching all refs in git log

- [ ] Show `You` correctly in results of Authors
- [ ] Add setting to adjust how many commits to look back (default 200)
- [ ] Update documentation

This could be made to be more efficient if we have a separate parser and formatter for contributors (which uses formatting to only get contributors like `git log --format="%an %ae"`). Then commits won't be fetched and thrown away.
This would also allow us to show more contributors than now because the command would return quicker

I don't see it as a blocker though, and is something that can be improved on.

fixes #1335

### DIFF
--- a/package.json
+++ b/package.json
@@ -1961,6 +1961,12 @@
 					"markdownDescription": "Specifies whether to show pull requests (if any) associated with commits in the _Contributors_ view. Requires a connection to a supported remote service (e.g. GitHub)",
 					"scope": "window"
 				},
+				"gitlens.views.contributors.preferUpstream": {
+					"type": "boolean",
+					"default": false,
+					"markdownDescription": "Specifies whether to use the upstream remote for contributors details",
+					"scope": "window"
+				},
 				"gitlens.views.defaultItemLimit": {
 					"type": "number",
 					"default": 10,
--- a/src/git/git.ts
+++ b/src/git/git.ts
@@ -748,6 +748,7 @@ export namespace Git {
 			reverse,
 			similarityThreshold,
 			since,
+			all
 		}: {
 			authors?: string[];
 			format?: 'refs' | 'default';
@@ -756,6 +757,7 @@ export namespace Git {
 			reverse?: boolean;
 			similarityThreshold?: number | null;
 			since?: string;
+			all?: boolean;
 		},
 	) {
 		const params = [
@@ -784,6 +786,10 @@ export namespace Git {
 
 		if (authors != null && authors.length !== 0) {
 			params.push('--use-mailmap', ...authors.map(a => `--author=${a}`));
+		}
+
+		if (all) {
+			params.push('--all');
 		}
 
 		if (ref && !GitRevision.isUncommittedStaged(ref)) {
--- a/src/git/gitService.ts
+++ b/src/git/gitService.ts
@@ -1865,6 +1865,7 @@ export class GitService implements Disposable {
 			ref?: string;
 			reverse?: boolean;
 			since?: string;
+			all?: boolean;
 		} = {},
 	): Promise<GitLog | undefined> {
 		const limit = options.limit ?? Container.config.advanced.maxListItems ?? 0;
@@ -1877,6 +1878,7 @@ export class GitService implements Disposable {
 				reverse: options.reverse,
 				similarityThreshold: Container.config.advanced.similarityThreshold,
 				since: options.since,
+				all: options.all
 			});
 			const log = GitLogParser.parse(
 				data,
--- a/src/git/gitService.ts
+++ b/src/git/gitService.ts
@@ -1503,26 +1503,30 @@ export class GitService implements Disposable {
 		let contributors = this.useCaching ? this._contributorsCache.get(repoPath) : undefined;
 		if (contributors == null) {
 			try {
-				const data = await Git.shortlog(repoPath);
-				const shortlog = GitShortLogParser.parse(data, repoPath);
-				if (shortlog != null) {
+				const log = await this.getLog(repoPath, {all: true});
+				if (log != null) {
 					// Mark the current user
 					const currentUser = await this.getCurrentUser(repoPath);
-					if (currentUser != null) {
-						const index = shortlog.contributors.findIndex(
-							c => currentUser.email === c.email && currentUser.name === c.name,
-						);
-						if (index !== -1) {
-							const c = shortlog.contributors[index];
-							shortlog.contributors.splice(
-								index,
-								1,
-								new GitContributor(c.repoPath, c.name, c.email, c.count, true),
-							);
-						}
-					}
+					// if (currentUser != null) {
+					// 	const index = log.authors.has(
+					// 		c => currentUser.email === c.email && currentUser.name === c.name,
+					// 	);
+					// 	if (index !== -1) {
+					// 		const c = shortlog.contributors[index];
+					// 		shortlog.contributors.splice(
+					// 			index,
+					// 			1,
+					// 			new GitContributor(c.repoPath, c.name, c.email, c.count, true),
+					// 		);
+					// 	}
+					// }
+					const authorsList: GitContributor[] = [];
+					log.authors.forEach((v) => {
+						const isCurrentUser = currentUser?.email === v.email && currentUser?.name === v.name;
+						authorsList.push(new GitContributor(repoPath, v.name, v.email || '', v.lineCount, isCurrentUser));
+					});
 
-					contributors = shortlog.contributors;
+					contributors = authorsList;
 				} else {
 					contributors = [];
 				}
--- a/src/git/models/commit.ts
+++ b/src/git/models/commit.ts
@@ -11,6 +11,7 @@ import { GitReference, GitRevision, GitRevisionReference, PullRequest } from './
 export interface GitAuthor {
 	name: string;
 	lineCount: number;
+	email?: string;
 }
 
 export interface GitCommitLine {
--- a/src/git/parsers/logParser.ts
+++ b/src/git/parsers/logParser.ts
@@ -420,6 +420,7 @@ export class GitLogParser {
 					author = {
 						name: entry.author,
 						lineCount: 0,
+						email: entry.email
 					};
 					authors.set(entry.author, author);
 				}
--- a/src/views/nodes/contributorNode.ts
+++ b/src/views/nodes/contributorNode.ts
@@ -102,6 +102,7 @@ export class ContributorNode extends ViewNode<ContributorsView | RepositoriesVie
 			this._log = await Container.git.getLog(this.uri.repoPath!, {
 				limit: this.limit ?? this.view.config.defaultItemLimit,
 				authors: [`^${this.contributor.name} <${this.contributor.email}>$`],
+				all: true
 			});
 		}
 


jasonwilliams:
pings @eamodio 

Also, how is Contibutors ordered right now? It would be nice if it orders by recent commit, so the more recent contributors are at the top.

jasonwilliams:
Ok some good news and bad news.

Good news:
I noticed that `git log` takes an `-all` option which will check across all refs. So I've added this as an option and now i can see commits from all remotes.

Before:
This was only checking HEAD ref, even though i have halidODat's remote added
halidBefore

After:
I can now see his commits from his own remote
halidAfter

Bad News:
Because commits are fetched lazily It doesn't seem possible or easy to order by most recent commit.

Question:
Do we need the config for this? It seems like `--all` should be a good default

jasonwilliams:
@eamodio I've managed to get it ordered by recent commit.
I achieved this by using `git.log` instead of `git.shortLog` then took the `authors` out of the response.

It's basically working as I would expect now.
Would appreciate any advice on if this is efficient or can be improved (as im sure reading git.log can be intensive)

eamodio:
@jasonwilliams Sorry for the delay on this one. Unfortunately, I'm really not a fan of using `git log` instead of `git shortlog`, as there are many implications (performance, limited results, etc).

I opened a discussion about adding ordering support to `shortlog` here: https://github.com/git-for-windows/git/discussions/3067

The next steps are to create a proposal for this to be added to Git, as suggested here: https://github.com/git-for-windows/git/discussions/3067#discussioncomment-428770. I haven't had a chance to do that (and probably won't for a while still), so it would be great if you could submit such a proposal.

jasonwilliams:
Hi @eamodio thanks for getting back to me on this.

> Unfortunately, I'm really not a fan of using git log instead of git shortlog, as there are many implications (performance, limited results, etc).

Would there still be performance implications with a specific filter? Other parts of Gitlens use git log with a filter attached. Doing this operation manually makes a difference in speed of returned results. This also solves limited results.

Even if the shortlog feature is added I’m not sure how you’ll get around git client compatibility.

>  I haven't had a chance to do that (and probably won't for a while still), so it would be great if you could submit such a proposal.

yeah I took a look at https://github.com/git-for-windows/git/discussions/3067#discussioncomment-547038

As much as I want to help I don’t think I’m in a position to do that. it seems a proposal would need at least some understanding of the Git codebase and a patch provided to work off and I can’t offer that.

eamodio:
So the main issue with `log` vs `shortlog` is that `log` just lists commits, whereas `shortlog` can be a rollup. So to get all the contributors to a repository, with `shortlog` it is a single (fast) git call, whereas with `log`, you'd have to get ALL the commits in the repository, and then find all the unique authors -- any limiting of the commits could miss authors. And even if you used `shortlog` to get the list of all the authors, there isn't a way of just getting the most recent commit for each of them, without resorting to either getting ALL commits again, or making N git calls (where N is the number of authors).

And lets say you wanted to combine the two approaches -- use `shortlog` to get the full list, but then use `log` to get say the most recent 3mo of commits (or some time-frame), and then use that more limited list to raise the sorting of authors that are in the recent set. The issue is still that the `log` call could return a VERY large number of commits (depending on the repository), and if you bounded it by say only 200 commits or something, the sorting could end up so unreliable that it would imo be almost worthless.

I, unfortunately, don't see a better option. :cry:

jasonwilliams:
Yeah, that’s a shame. I think it’s not enough to just ask for the feature in their mailing list, it looks like a patch is also required and some insight into how it would work.

> and if you bounded it by say only 200 commits or something, the sorting could end up so unreliable that it would imo be almost worthless.

I’m not sure I understand this. Why does the sorting become unreliable? In my patch the sorting is most recent commit to oldest by author (deduped). On quite a few different repositories the sorting worked as expected.

Personally if I had to choose between bounding of 200 to not having the feature at all I’d still believe the bounding is better until shortlog can be used. When collaborating with someone on an old codebase it takes a while to find them and then I can’t even see their remote branch anyway. This improves the UX for collaboration. It’s true the downside would be not seeing all contributors but why would someone need this?

eamodio:
You're PR will get ALL commits in the repo, so while it will work, it's extremely expensive (of course depending on the repository). And bounding it to only the X most recent commits, can be unreliable because there might only be 1 or 2 authors in those recent commits and now their shorting is different than all the other authors.

jasonwilliams:
> You're PR will get ALL commits in the repo, so while it will work, it's extremely expensive (of course depending on the repository).

My understanding here was [passing limit](https://github.com/eamodio/vscode-gitlens/pull/1342/files#diff-7581395e5e99772a736828baac71cb778dc35561c89c49135dd7266ac6b56286R103) to getLog was the equivalent of calling `git log -n x`. 

Limit set [here](https://github.com/eamodio/vscode-gitlens/blob/1b6feab73551b907388b1eb98330251ce502219d/src/git/gitService.ts#L1928) which adds to the command line here: https://github.com/eamodio/vscode-gitlens/blob/1b6feab73551b907388b1eb98330251ce502219d/src/git/git.ts#L779

Am i missing something?

eamodio:
> I would only include authors from those commits

That's the issue for me -- if you truncate/limit the commits, you are limiting the authors detected. And on a busy repo (like VS Code), a hundred commits or so could only capture a day or two. So depending on how often an individual commits, they could pop in and out of the _Contributors_ view. The _Contributors_ view should always show all contributors. 

Personally, I would love to see that Git proposal happen (since I think it's a valuable feature for Git), but at the same time I completely understand that it's out of your expertise (it is out of mine as well, I haven't done C in MANY years). But regardless of that, I would also like to better understand your use-case/workflow that is driving your desired changes here. Maybe there is another way of accomplishing the behavior you want.

jasonwilliams:
Hey @eamodio, after doing some research I reached out to @peff (maintainer of Git and implementer of some shortlog features) about this and what options we have. I sent him a link to this PR and the git-for-windows one for context. This was his reply, its very insightful.


> Thanks for those links. My first response would have been: you should
> probably use git-log, but I see you already tried that. :)
> 
> I do think that the concern about looking at all commits may be overly
> conservative there. After all, that is exactly what git-shortlog is
> doing (it is just running git-log internally and collating the results).
> Sending the output to another program to collate incurs some extra
> reading/writing overhead, but should be in the same ballpark.
> 
> Running:
> 
>   \# collate and count by name, but sort the output by recency
> ```
>   git log --format='%at %aN' --no-merges |
>   perl -lne '
>     /^(\d+) (.*)/; # <time> <author>
>     $count{$2}++;
>     $time{$2} = $1 if $1 > $time{$2};
>     END {
>       for my $author (sort { $time{$b} <=> $time{$a} } keys(%count)) {
>         print "$count{$author} $author";
>       }
>     }
>   '
> ```
> 
> takes roughly the same amount of time as:
> 
>   git shortlog -ns --no-merges
> 
> (I get 380ms and 330ms respectively).
> 
> That said, two other options which might make sense to me are:
> 
>   - tell shortlog not to traverse so far back. There was some discussion
>     of "-n" there, but probably "--since=1.year.ago" might be a better
>     option, as that scales with time independent of how busy the
>     repository is. A nice side effect here is that it actually makes
>     shortlog _cheaper_, since it can stop traversing past the cutoff.
> 
>   - you can ask shortlog to tell you the time of each commit. Like:
> 
>       git shortlog -n --date=short --format='%ad %s'
> 
>     though now we are back to having one line per commit, like the
>     git-log above. The %s there could also feed the "show commits from
>     each author" that GitLens supports, but it is loading all at once
>     (whereas it sounds like the current code does it on-demand when
>     click the author).
> 
> > It seems git shortlog does not support ordering by commit date, only by the
> > number of commits per author. I noticed you've worked in this area and was
> > wondering how easily can this be added? C is not my strong point, and it
> > seems i can't just ask for a feature request on the mailing list.
> 
> I doubt it would be that hard to list the authors by commit date. The
> code internal to shortlog is conceptually similar to the perl I showed
> above (counting up each name and then sorting by count). So we'd just be
> remembering the max date for each author and then sorting by that at the
> end.
> 
> I'm not entirely sure it solves your problem, though (and likewise for
> the perl I showed above). You still have to decide when you've gone far
> enough into the list and are showing old, non-recent authors. If you go
> too far, then you're showing names that aren't valuable. If you don't go
> far enough, you may omit authors that are just inactive lately, but
> still have a huge count of commits.
> 
> So I think do think the "--since" option is better if you're going to
> have a cutoff. It's simpler and more efficient.
> 
> If you're not going to have a cutoff, then you probably need some more
> complex formula to give each author a "score": how recent were they
> active, and how active were they? And that's getting into something that
> we probably wouldn't do in shortlog (but you could easily do by piping
> git-log as shown above).
> 
> I hope that helps. We aren't generally interested in raw feature
> requests on the list, but it's not that they're not _allowed_. It's just
> that nobody is likely to work on it unless your suggestion really
> tickles someone's fancy. But we are happy to help guide people who
> aren't sure where to get started, and even discussion like we're having
> above might be valuable to have in the public list archive.
> 
> -Peff

eamodio:
@jasonwilliams Thank you for the follow up. I've run some perf tests with using `shortlog` vs `log` and I was able to get things to be pretty close on the VS Code repo -- `shortlog` was around 640ms and `log` was ~70ms more. Git returned in almost the same amount of time, but my parsing of the `log` data took ~60-70ms.

Given that, I've checked in a set of new changes to the _Contributors_ view using `log`, see https://github.com/eamodio/vscode-gitlens/commit/b60637f00e245fd23c84ceebf7a4aeb637c64d96.

It will now use the upstream branch (if one exists), or all branches (if specified by the new `gitlens.views.contributors.showAllBranches` setting). And you can also control the sorting of the _Contributors_ view via the new `gitlens.sortContributorsBy` setting. The default is still by commit count (`count:desc`), but can be easily changed to `date:desc`.

Let me know what you think, and thanks again for your work on this -- and pushing it (and me) forward!

Thank you!

jasonwilliams:
@eamodio is this now in the main version of the plugin or do i still need the nightly for this feature?

eamodio:
It's in stable now.

PR #1396, opened 2021-02-23
Closes #1379 - adds Gitea remote provider

### LINKED ISSUE #1379: Add support for gitea as a remote
Gitea is a popular, lightweight alternative to GitHub or GitLab and becoming increasingly popular. It would be nice to add native support for it as a remote. Currently one has to set up a custom remote and specify all the urls to fully use it.

septatrix:
# Description

Add support for Gitea as a remote (fixes #1379).

# Checklist

<hr>

The branch parsing logic in particular is a bit different from GitHub and GitLab (shortest branch vs longest branch). This is to match the parsing logic used by gitea for URLs. Also the method to generate URLs from files uses a deprecated route (which is kept for backwards compatibility) as there is no alternative for branchless, SHA-less file paths.

I still need to test the URL parsing for more ambiguous inputs (which even Gitea sometimes handles weird). Also I need to add an icon. Based on the existing icons you prefer black & white variants. Is there anything else I have to watch out for?


eamodio:
@septatrix This is awesome! Thanks! Yeah, for the icons, its best to keep it b&w. I can't think of anything you've missed or need to consider. Great work!

septatrix:
FYI: I am writing an exam at the end of this week so I will probably not get to adding the icons until next week.

septatrix:
I have now added the icons. As the gitea logo normally has a fully white piece which did not work too well in dark mode with a simple conversion to b/w I borrowed the approach from the `match-*-selected` icons and used a slightly transparent fill for those areas. All the URL work fine on https://try.gitea.io/ with the exception of tags. As they are treated the same as branches by most git providers except gitea the is no `RemoteResourceType.Tag`. Gitea however differentiates between these two which leaves those links broken...

septatrix:
@eamodio How would you like to go forward with this? I am not familiar with the inner workings of this extension so this is probably as far as I am able to implement Gitea support.

The best solution would probably be to merge this with the current functionality as most things now work. Support for providers which differentiate between tags and branches can then be tackled in a separate issue.

eamodio:

Thank you!

PR #1575, opened 2021-07-10
Fixes #1131 - "Search Commits" allows free term at search end

### LINKED ISSUE #1131: Order matters for search filters in 'search commits'
Issue Type: <b>Bug</b>

**Repro**
1. Run the `Search Commits` command
1. Enter `search author:AUTHOR`
     - See search results
1. Now run the `Search Commits` command again
1. This time, swap the order of the two search terms: `author:AUTHOR search`
 
**Bug**
No results are returned in the second case

Extension version: 10.2.2
VS Code version: Code - Insiders 1.51.0-insider (74ef0a92fe8be7034d7d5f513e31262ab77d3bf1, 2020-10-19T05:48:39.773Z)
OS version: Darwin x64 19.6.0

Git-Lior:
# Description

Fixes #1131

Before this change, it was assumed (in `searchMessageOperationRegex`) that only the text *before* the search operations can be a free-text term.

With this change, we look for the free-text term in two places:
* before the first search operation (if none - the entire search is free-text)
   i.e:  `hello world author: abc def` is interpreted as `message:"hello world", author: "abc def"`
* after the first message value (respecting double-quotes) of the last search operation.
   i.e:  `author: abc def hello world` is interpreted as `author: "abc", message: "def hello world"`
   to get the same result as before with this format, the search should be: `author: "abc def" hello world`
   (this rule only applies if there is no free-text at the start of the search)

# Checklist

### DIFF
--- a/src/git/search.ts
+++ b/src/git/search.ts
@@ -39,10 +39,9 @@ export interface SearchPattern {
 export namespace SearchPattern {
 	const emptyStr = '';
 
-	const searchMessageOperationRegex = /(?=(.*?)\s?(?:(?:=:|message:|@:|author:|#:|commit:|\?:|file:|~:|change:)|$))/;
 	const searchMessageValuesRegex = /(".+"|[^\b\s]+)/g;
 	const searchOperationRegex =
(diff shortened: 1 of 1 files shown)


eamodio:
Awesome! Thank you so much for your contribution!

Thank you!

Issue #1617, opened 2021-08-27
better keybinding/shortcut to show scmView in order to deal with conflict


heartacker:
gitlen hijack the key <kbd>ctrl+shift+g</kbd>  which to show scm view, maybe some of user notice that it could use <kbd>ctrl+shift+g g</kbd> in stead and other user would not.

after few days experience of gitlens and my muscle Memory tells me that we could use  <kbd>ctrl+shift+g</kbd>+ <kbd>ctrl+shift+g</kbd> to show the scm.view. that will be great to have it.

what in my mind is:
one <kbd>ctrl+shift+g</kbd>, scmview wouldn't show, Intuitively press again <kbd>ctrl+shift+g</kbd>.

and notice that that will be an better experience.
![image](https://user-images.githubusercontent.com/1876302/131118076-2810a640-f0ed-439d-abe8-8cda81ff4dac.png)

```jsonc
{
  "key": "ctrl+shift+g g",
  "command": "workbench.view.scm",
  "when": "workbench.scm.active && !gitlens:disabled && config.gitlens.keymap == 'chorded'"
},
{
  "key": "ctrl+shift+g ctrl+shift+g", // add another key
  "command": "workbench.view.scm",
  "when": "workbench.scm.active && !gitlens:disabled && config.gitlens.keymap == 'chorded'"
}
```

heartacker:
can anybody see this:

https://github.com/gitkraken/vscode-gitlens/pull/1618

heartacker:
？

PR #1618, opened 2021-08-27
add another key `ctrl+shift+g ctrl+shift+g`  to show scm view

heartacker:
this will fix the issue #1617

add another key <kbd>ctrl+shift+g </kbd><kbd>ctrl+shift+g</kbd> to deal with the conflict with chinese ime (and other cjk user?) to show scm view


heartacker:
hi @eamodio  will you please cost some time to review this.
that is useful for not-Latin language IME

Issue #1729, opened 2021-11-17
Allow user to control the --first-parent flag when showing file history


alessioprestileo:
I have merged two repos using this commands:

```
git remote add project-a /path/to/project-a
git fetch project-a --tags
git merge --allow-unrelated-histories project-a/master # or whichever branch you want to merge
git remote remove project-a
```

Before merging `project-a` into my target repo, I moved all the files in `project-a` to a `packages/project-a` folder, so that they would fit in the structure of the target repo (which is a monorepo).
Now when I look at the file history in gitlens (`gitlens.advanced.fileHistoryFollowsRenames` option set to true), it only shows 1 commit for all files coming from `project-a`, as if they were created from scratch.
When I run this in the terminal:

```
git log --follow <filePath-from-project-a>
```

it shows me the complete commit history.

I looked into the extension code and found that, when displaying the file history, the the `--first-parent` flag is always added to the `git log` command if the `gitlens.advanced.fileHistoryFollowsRenames` option is set to true.
I would like to add a `gitlens.advanced.fileHistoryStopsAtFirstParent` option, so that the user can control the use of the `--first-parent` flag.
I have already prepared a PR for this.

alessioprestileo:
@eamodio 
I haven't received any feedback on this PR, who should I ask?

Kamori:
I also am in a situation where I cannot alter the version of git. But git segfaults using `git log -L1,2:/file.txt --first-parent` together. It would greatly improve my situation (as a stop gap) to be able to force not include this flag.

alessioprestileo:
@bantic @brettcannon @rivy 
Any feedback on this PR?
Any suggestions on who I could ask?

brettcannon:
@alessioprestileo I think you at @ mentioned the wrong person as I have nothing to do with this repo.

garywheeler69:
A

eamodio:
Can you please see if this issue is still reproducible in the pre-release edition of GitLens? Thanks!

You can switch to the pre-release edition of GitLens, by clicking on the "Switch to Pre-Release version of this extension" from the Extensions view.

![image](https://user-images.githubusercontent.com/641685/228608740-6071bc19-7b1a-44c8-843f-d2af44d6114b.png)

alessioprestileo:
@eamodio 

It seems the issue is now fixed, even without switching to the pre-release edition :-)

I'm not sure when it started working, I stopped relying on that feature and haven't used it in a long time.

PR #1730, opened 2021-11-17
Allows user to control the --first-parent flag

alessioprestileo:
Allows user to control the --first-parent flag when displaying file history.

I explain my need for this feature in
https://github.com/Axosoft/vscode-gitlens/issues/1729


d13:
Issue reported as fixed https://github.com/gitkraken/vscode-gitlens/issues/1729#issuecomment-1782036650

PR #1738, opened 2021-12-03
GPL v3.0 license

SteveJobs91:
I think that as this software is so useful, it should be strongly protected with a GPL v3.0 license instead of an MIT license.
My fork of this project has a GPL v3.0 license instead of an MIT license. I hope you do the same.

By the way, thanks for the great software.
pls pls pls


eamodio:
Thank you, but we aren't looking to change the license.

SteveJobs91:
ok no problem.
By the way, any way I can contribute to the software? Sounds amazing...

PR #1792, opened 2022-01-31
Updated README for Marketing Department

TristanTheLamb:
Updated copy based on suggestions from Lauren and Dwayne.


eamodio:
Thanks!

PR #1817, opened 2022-02-10
Fixes #1805 to prevent tracking remote branches if the local branch has not the same name

### LINKED ISSUE #1805: GitLens should only auto-track remote branches if the name matches
Currently when you create a new branch via GitLens from a remote branch, it automatically sets up tracking on the new branch. GitLens should only setup that tracking if the branch names match -- otherwise it can lead to unexpected tracking.

Alternatively, we could avoid tracking all together or prompt, but I feel like the above is the best of both.

qadram:
# Description

When creating a branch, if the ref is remote, then passes along the --no-track option to be added to the command line, both for the branch command, or for the checkout -b command.

### DIFF
--- a/src/commands/git/branch.ts
+++ b/src/commands/git/branch.ts
@@ -30,7 +30,7 @@ interface Context {
 	title: string;
 }
 
-type CreateFlags = '--switch';
+type CreateFlags = '--switch' | '--no-track';
 
 interface CreateState {
 	subcommand: 'create';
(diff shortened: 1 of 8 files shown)


src/commands/git/branch.ts
```diff
@@ -324,8 +324,16 @@ export class BranchGitCommand extends QuickCommand<State> {
 			}
 
 			QuickCommand.endSteps(state);
+
+			if ((state.reference as GitBranchReference).remote) {
```
eamodio, line 328: Here we need to do a check that we are dealing with a branch (because the `reference` could be a commit or tag or something else.

src/commands/git/switch.ts
```diff
@@ -73,7 +73,7 @@ export class SwitchGitCommand extends QuickCommand<State> {
 			() =>
 				Promise.all(
 					state.repos.map(r =>
-						r.switch(state.reference.ref, { createBranch: state.createBranch, progress: false }),
+						r.switch(state.reference.ref, { createBranch: state.createBranch, progress: false, track: false }),
```
eamodio, line 76: Is this an intentional change in behavior? Because `switch` will now always add the `--no-track` flag, which I'm not sure is desired.

eamodio (changes requested):
@qadram sorry for the delay on this -- added some comments and can you rebase this?

WarDroid23 (commented):
.

d13:
Resolved with a88f80ef05ce5d62166a72b28af0f59003c0d47d

PR #1947, opened 2022-04-05
Fixes #969 adds copy relative path to context menu

### LINKED ISSUE #969: Show copy relative path in the contextual menu
Hello, 

is it possible to show the command "copy relative path" in the contextual menu like in the explorer ?

image

The point here is that I use git along with another versioning tool at work. Git offers me some flexibility I don't have with the other tool, but when I have to use the other tool, I need at a moment or another the names of the files I have modified between two git branches. GitLens is truly awesome for this, since I can have this information easily, but getting the files paths is quite bothersome (especially when the list of file is huge...) since the only way I found until now is to right click on the file to open it, and then on the editor name, to right click again to get the relative path name. 

Thanks

d13:
# Description

This change adds a `Copy Relative Path` item to the context menu and the command palette. This copies the relative path, of the active file in the editor, to the clipboard.

Fixes #969

Screen Shot 2022-04-04 at 11 51 16 PM

# Checklist

### DIFF
--- a/src/commands/copyRelativePathToClipboard.ts
+++ b/src/commands/copyRelativePathToClipboard.ts
@@ -1,0 +1,26 @@
+import { env, TextEditor, Uri } from 'vscode';
+import { Commands } from '../constants';
+import type { Container } from '../container';
+import { command } from '../system/command';
+import { ActiveEditorCommand, getCommandUri } from './base';
+
+@command()
+export class CopyRelativePathToClipboardCommand extends ActiveEditorCommand {
+	constructor(private readonly container: Container) {
+		super(Commands.CopyRelativePathToClipboard);
+	}
+
+	async execute(editor?: TextEditor, uri?: Uri) {
+		uri = getCommandUri(uri, editor);
+		let relativePath = '';
+		if (uri != null) {
+			const repoPath = this.container.git.getBestRepository(editor)?.uri;
+			if (repoPath != null) {
+				relativePath = this.container.git.getRelativePath(uri, repoPath);
+			}
+		}
+
+		void (await env.clipboard.writeText(relativePath));
+		return undefined;
+	}
+}
--- a/src/commands/copyRelativePathToClipboard.ts
+++ b/src/commands/copyRelativePathToClipboard.ts
@@ -1,13 +1,23 @@
-import { env, TextEditor, Uri } from 'vscode';
+import type { TextEditor, Uri } from 'vscode';
+import { env } from 'vscode';
 import { Commands } from '../constants';
 import type { Container } from '../container';
 import { command } from '../system/command';
-import { ActiveEditorCommand, getCommandUri } from './base';
+import type { CommandContext } from './base';
+import { ActiveEditorCommand, getCommandUri, isCommandContextViewNodeHasFileCommit } from './base';
 
 @command()
 export class CopyRelativePathToClipboardCommand extends ActiveEditorCommand {
 	constructor(private readonly container: Container) {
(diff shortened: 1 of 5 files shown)


src/commands/copyRelativePathToClipboard.ts
```diff
@@ -0,0 +1,26 @@
...
+import type { Container } from '../container';
+import { command } from '../system/command';
+import { ActiveEditorCommand, getCommandUri } from './base';
+
+@command()
+export class CopyRelativePathToClipboardCommand extends ActiveEditorCommand {
+	constructor(private readonly container: Container) {
+		super(Commands.CopyRelativePathToClipboard);
+	}
+
```
eamodio, line 12: @d13 To ensure this copies the correct path when you right-click on a commit/etc, we need to handle the `preExecute` call, similar to
https://github.com/gitkraken/vscode-gitlens/blob/79f0d0b2971e506e7fbf5cbaa928288a07e823b7/src/commands/copyShaToClipboard.ts#L28-L50

Without that, we will be trying to use the repository for the open editor, rather than the context provided.

src/commands/copyRelativePathToClipboard.ts
```diff
@@ -0,0 +1,26 @@
...
+	async execute(editor?: TextEditor, uri?: Uri) {
+		uri = getCommandUri(uri, editor);
+		const repoPath = this.container.git.getBestRepository(editor)?.path;
+		let relativePath = '';
+		if (uri != null && repoPath != null) {
```
eamodio, line 17: Instead of messing with the path here, you should be able to call `this.container.git.getRelativePath(uri, repoPath)`

src/commands/copyRelativePathToClipboard.ts
```diff
@@ -0,0 +1,26 @@
...
+
+@command()
+export class CopyRelativePathToClipboardCommand extends ActiveEditorCommand {
+	constructor(private readonly container: Container) {
+		super(Commands.CopyRelativePathToClipboard);
+	}
+
+	async execute(editor?: TextEditor, uri?: Uri) {
+		uri = getCommandUri(uri, editor);
+		const repoPath = this.container.git.getBestRepository(editor)?.path;
```
eamodio, line 15: Using the `getRelativePath` below, get the `.uri` prop rather than path here. Basically want to transition to using Uris everywhere as possible.

eamodio:
Looking good! Added review comments.

eamodio (changes requested):
Looking good, added some comments.

marcoapcaldas:
Hi @d13 ! I'm try to give my first contribution to gitlens and I think I could finish this. I'm planning to take the files you modified and complete the changes requested by @eamodio. Is that ok?

src/commands/copyRelativePathToClipboard.ts
d13: When testing in the different sidebar views, `isCommandContextViewNodeHasFileCommit` covered each scenario.

eamodio:
My only question now is ordering/placement of the menu entry. Not sure it should be between the SHA/Message, maybe after? Should it be in its own section?

d13:
> My only question now is ordering/placement of the menu entry. Not sure it should be between the SHA/Message, maybe after? Should it be in its own section?

That's an old screenshot. I currently have it grouped with the `Copy Remote File` actions although I'm not really sold on it: 

Screen Shot 2023-03-13 at 7 05 27 PM

I think it would be better as its own group:

Screen Shot 2023-03-13 at 7 19 42 PM

eamodio:
Perfect

PR #1967, opened 2022-04-13
The bug is fixed

jasonleonhard:
Everything works again


eamodio:
@jasonleonhard Can you please add details as to what you are looking to accomplish with this PR? On GitHub, those trailing `-` are required (as far as I can tell) to have proper navigation. While the VS Code marketplace STILL has bugs dealing with that appropriately.

Issue #1964, opened 2022-04-13
Support Git remote providers running behind a reverse proxy


adambadura:
In our use case, we have Gerrit set up with the main (push) server being on `gerrit.ext.net.nokia.com`, however, Gitiles are on `https://gerrit.ext.net.nokia.com/gerrit/` rather than `https://gerrit.ext.net.nokia.com/`. It seems there is no way to configure this. While without proper configuration for example links from `Change-Id:` in the commit message don't work since they use `https://gerrit.ext.net.nokia.com/q/<change-id>` instead of `https://gerrit.ext.net.nokia.com/gerrit/q/<change-id>`.

For a more detailed investigation and analysis see #1954 and [the comment](https://github.com/gitkraken/vscode-gitlens/pull/1954#issuecomment-1095275182) by @felipecrs in particular.

adambadura:
Why only a `"pathPrefix"`? Why not provide possibility to specify domain/URL for each individual type of operation? Someone might come up with a setup where push is done to the main server but both fetch and query are done on secondary servers.

felipecrs:
> but both fetch and query are done on secondary servers.

I don't think this is realistic... as I mentioned at some point, in my company we also have such a master/mirror setup. We should clone from mirror and push to master, but any HTTPS operation (like querying) should be done against master.

adambadura:
Still, I consider the benefit of providing only the `pathPrefix` to be smaller than the flexibility of a full URL. However, if the "framework" allows it perhaps a more flexible configuration could be made. Like:
```json
{"domain": "gerrit.ext.net.nokia.com", "type": "Gerrit", "query": {"pathPrefix": "gerrit/"}}
```
or
```json
{"domain": "gerrit.ext.net.nokia.com", "type": "Gerrit", "query": {"url": "why.not.gerrit.ext.net.nokia.com/gerrit/"}}
```

felipecrs:
I always prefer to give users less things to tinker with rather than providing more options unless the options are really needed. Which hasn't been the case so far.

But anyway, that's only my opinion.

felipecrs:
By the way, if you enter in the clone page of any project, could you please confirm that the URL matches the following format when using HTTPS protocol?

```
git clone https://<user>@gerrit.ext.net.nokia.com/gerrit/a/<project>
```

adambadura:
I have the following options:

[...]

PR #1984, opened 2022-04-28
Smartly guess the protocol and port for the remote provider

### LINKED ISSUE #632: How to add port number to GitLab remote?
Our GitLab server runs on port 8000, so when you right-click on a file and select 'Open File on Remote', or 'Copy Remote Url to Clipboard', the generated URL doesn't work. It looks like this:

```
https://gitlab.example.com/example/foo/blob/3146b57606b1e727b4cd70e108df2dd51e3a332f/path/to/file
```

whereas it should be:

```
https://gitlab.example.com:8000/example/foo/blob/3146b57606b1e727b4cd70e108df2dd51e3a332f/path/to/file
```

I can't figure out how to configure this in the GitLens settings. I've tried adding this:

```json
    "gitlens.remotes": [
        {
            "domain": "gitlab.example.com",
            "type": "GitLab",
            "urls": {
                "file": "https://gitlab.example.com:8000/..."
            }
        }
    ],
```

But this seems to have no effect at all on the generated URL. What am I doing wrong?

### LINKED ISSUE #1918: git remote get-url is http but open branch on remote url is always https
- GitLens Version: v12.0.5
- Git Version: git version 2.35.1.windows.2

- VSCode Version: 1.65.2
- OS Version: windows11 x64

Steps to Reproduce:
1. git clone httt://anyurl, is http not https 
2. open this folder in vscode
3. click `Source Control`
image
4.click `BRANCHES` then click `Open Branch on Remote`
image
5. browser will auto open this url, but this url is https instead of http

log gif

how to fix it ?

felipecrs:
# Description

Currently the remote provider uses https by default and allow users to set a different protocol if they want in the configuration.

This is kept, but now, we try to infer the best protocol by looking at the git remote URL.

And, previously, no custom ports were allowed to be set. This is still true, but now, we try to infer the best port by looking at the git remote URL.

Fixes #632
Fixes #1918

Maybe more.

# Checklist

- [ ] My changes include any required corresponding changes to the documentation

### DIFF
--- a/package.json
+++ b/package.json
@@ -11070,6 +11070,7 @@
 		"ansi-regex": "6.0.1",
 		"billboard.js": "3.3.2",
 		"chroma-js": "2.3.0",
+		"git-url-parse": "11.6.0",
 		"https-proxy-agent": "5.0.1",
 		"iconv-lite": "0.6.3",
 		"lodash-es": "4.17.21",
@@ -11082,6 +11083,7 @@
 	"devDependencies": {
 		"@squoosh/lib": "0.4.0",
 		"@types/chroma-js": "2.1.3",
+		"@types/git-url-parse": "9.0.1",
 		"@types/glob": "7.2.0",
 		"@types/lodash-es": "4.17.6",
 		"@types/mocha": "9.1.0",
--- a/src/git/parsers/remoteParser.ts
+++ b/src/git/parsers/remoteParser.ts
@@ -1,13 +1,10 @@
+import gitUrlParse from 'git-url-parse';
 import { debug } from '../../system/decorators/log';
 import { GitRemote } from '../models';
 import { GitRemoteType } from '../models/remote';
 import { RemoteProvider } from '../remotes/provider';
 
-const emptyStr = '';
-
 const remoteRegex = /^(.*)\t(.*)\s\((.*)\)$/gm;
-const urlRegex =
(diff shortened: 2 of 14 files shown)


felipecrs:
@eamodio, @oliverdunk this is a draft for what we have been discussing in #1964. It would be a requirement for implementing support for `urls.base` (if we go by the [approach I mentioned](https://github.com/gitkraken/vscode-gitlens/issues/1964#issuecomment-1100339957)).

Do not worry about reviewing this in details (I expect things to be missing, which I will caught yet), but I would like to have your +1 in the route I'm taking with this PR before spending more time on it.

Any comments are appreciated.

oliverdunk:
Hey @felipecrs, thanks for poking at this! Two things that crossed my mind looking at the diff:

- Introducing a new library to parse URLs and new data structure to store them in feels like quite a large change. I'm fine with that if everyone agrees it's the best approach, but if we want to keep the changes minimal I feel like there are other ways to get the protocol using the existing code.
- Is changing the constructor of these providers a breaking change and is that a concern? Presumably other extensions that integrate with GitLens could be instantiating them.

I've never contributed to GitLens so am just replying here since I'm keen to see something land! All of this should be taken with a grain of salt.

felipecrs:
Very good points!

kumekay:
@felipecrs Do you have any plans to continue work on this PR?

felipecrs:
> @felipecrs Do you have any plans to continue work on this PR?

Sorry, no. Feel free to take it over.

oleksandr-vonahi:
@gooroo-ai please review

oleksandr-vonahi:
@gooroo-dev please review

PR #1994, opened 2022-05-12
Fixes #1981 by having the view always enabled

### LINKED ISSUE #1981: Add GitLens Sidebar view after install, even if VSCode does not have a repo open
When installing GitLens, if VSCode does not have a repo open, then the GitLens sidebar does not show up currently.
Once you open a repo, all works as expected, but it is not obvious that you need a repo open to see the GitLens sidebar, as the Source Control, File Explorer and other stock items do not require it

Request is to have the GL sidebar appear regardless of if a repo is open or not. 

Background:
Upon installing Fedora 35, I was setting up my dev tools but had not yet cloned any repos down.  
I installed VSCode and next installed GitLens (as I can't imagine working without it now) and thought I hit an error as the sidebar did not show up. 
Once I cloned a repo down and opened it in VSCode then GL started working as expected, but this seriously threw me off. 
I think is a realistic, but likely less often encountered scenario that would be run into, especially by people that change OS installs like socks ;)
Even if it just had the Welcome screen allowing me to connect to my GL+ account, I think that would be enough to let me know it is working. It would make sense to hide all the other sections until it can see a repo IMHO though.

joseleon-axosoft:
---

# Description

Fixes the issue #1981 by simply allowing the view to be always enabled. At this moment, the content shown on GitLens : Home view looks great to me.

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -10739,7 +10739,7 @@
 					"type": "webview",
 					"id": "gitlens.views.home",
 					"name": "Home",
-					"when": "!gitlens:disabled && config.gitlens.plusFeatures.enabled",
+					"when": "config.gitlens.plusFeatures.enabled",
 					"contextualTitle": "GitLens",
 					"icon": "images/gitlens-activitybar.svg",
 					"visibility": "visible"


eamodio:
I'm not ready to merge this -- as I still feel there is more UI work that is needed to guide users why other GitLens UI will be missing because they haven't opened a folder

eamodio:
This has been resolved in the upcoming GL 13.1

d13:
Fixed with a312aa4bd81e5ab99acb56bcae985f21e7b45fb8

PR #2090, opened 2022-07-14
fix centering of text in headers

miggy-e:
# Description

base.scss contains a css style for the body:
'font-size: 100% !important'
that causes the header of the graph to be misaligned and overrides some font sizing in the graph, making it inconsistent with the original graph.

This PR contains code which will override that style specifically in the graph

### DIFF
--- a/src/webviews/apps/plus/graph/graph.scss
+++ b/src/webviews/apps/plus/graph/graph.scss
@@ -7,3 +7,8 @@
 @import '../../shared/codicons';
 @import '../../shared/utils';
 @import '../../../../../node_modules/@axosoft/gitkraken-components/lib/css/GKGraphStyles.css';
+
+// override important font size from base.scss
+#graph-container-helmet {
+    font-size: 14px;
+}


axosoft-ramint (approved):
LGTM, thanks for fixing it!

src/webviews/apps/plus/graph/graph.scss
```diff
@@ -7,3 +7,8 @@
 @import '../../shared/codicons';
 @import '../../shared/utils';
 @import '../../../../../node_modules/@axosoft/gitkraken-components/lib/css/GKGraphStyles.css';
+
+// override important font size from base.scss
+#graph-container-helmet {
+    font-size: 14px;
```
ericf-axosoft, line 13: I think this is not the correct way to fix that. The fix should be on the library side.

And then, the css will be included on our generated `GKGraphStyles.css` file.

ericf-axosoft (changes requested):

src/webviews/apps/plus/graph/graph.scss
miggy-e: My thought was that CSS specific to the gitlens app was messing with our CSS styling, so adding the CSS rule inside of  gitlens seemed fitting (this rule won't necessarily be needed when we use the graph in other places). But, I'm willing to move this CSS to the component.

ericf-axosoft:
This change will be fixed on the library side in PR: https://github.com/gitkraken/GitKrakenComponents/pull/7

PR #2102, opened 2022-07-19
GITLENS-198 Add support for stash nodes to the graph

axosoft-ramint:
Addresses GITLENS-198: https://gitkraken.atlassian.net/browse/GITLENS-198

Important notes:
1. Our local git provider regex for forming summary and message of a stash can sometimes generate blank messages and summaries when the summary of the stash from the git log does not include a WIP section or a summary section. For example, on stashes created within GitKraken we often get a description like `On expand-graph-container:`. Running `stashSummaryRegex` on this yields an `onRef` group, but a blank `wip` group and a blank `summary` group which causes `getStash` to yield a blank summery and message for that stash commit. In this case, I updated the logic to default to `WIP on ${onRef}`.
2. The local git provider was not previously parsing or storing the parent of a stash commit. This logic was added to `getStash`.
3. Based on the way we currently load commit rows into the graph, we can have orphaned stashes. These are stashes were the parent commit is not loaded into the graph but the stash commit is. You may notice these stashes in your local test, where the edge goes all the way down to the bottom of the graph. In follow-up work, we need a solution for this. One idea is to check if each stash's parent sha exists in the graph and not load the stash if it doesn't, but I'm open to other suggestions.

### DIFF
--- a/src/env/node/git/localGitProvider.ts
+++ b/src/env/node/git/localGitProvider.ts
@@ -3299,12 +3299,14 @@ export class LocalGitProvider implements GitProvider, Disposable {
 				sha: string;
 				date: string;
 				committedDate: string;
+				parents: string;
 				stashName: string;
 				summary: string;
 			}>({
 				sha: '%H',
 				date: '%at',
 				committedDate: '%ct',
+				parents: '%P',
 				stashName: '%gd',
 				summary: '%B',
 			});
@@ -3320,20 +3322,27 @@ export class LocalGitProvider implements GitProvider, Disposable {
 				let onRef;
 				let summary;
 				let message;
+				let wipSummary;
+				let matchSummary;
+
 				const match = stashSummaryRegex.exec(s.summary);
 				if (match?.groups != null) {
+					matchSummary = match.groups.summary.trim();
 					onRef = match.groups.onref;
+					wipSummary = `WIP on ${onRef}`;
 					if (match.groups.wip) {
-						message = `WIP: ${match.groups.summary.trim()}`;
-						summary = `WIP on ${onRef}`;
+						message = `WIP: ${matchSummary}`;
+						summary = wipSummary;
 					} else {
-						message = match.groups.summary.trim();
-						summary = message.split('\n', 1)[0] ?? '';
+						message = matchSummary;
+						summary = message ? message.split('\n', 1)[0] : wipSummary;
 					}
 				} else {
 					message = s.summary.trim();
 					summary = message.split('\n', 1)[0] ?? '';
 				}
+
+				const baseParentSha = s.parents.split(' ')[0];
 
 				commits.set(
 					s.sha,
@@ -3344,7 +3353,7 @@ export class LocalGitProvider implements GitProvider, Disposable {
 						new GitCommitIdentity('You', undefined, new Date((s.date as any) * 1000)),
 						new GitCommitIdentity('You', undefined, new Date((s.committedDate as any) * 1000)),
 						summary,
-						[],
+						baseParentSha ? [baseParentSha] : [],
 						message,
 						s.files?.map(
 							f => new GitFileChange(repoPath, f.path, f.status as GitFileStatus, f.originalPath),
--- a/src/env/node/git/localGitProvider.ts
+++ b/src/env/node/git/localGitProvider.ts
@@ -3334,7 +3334,7 @@ export class LocalGitProvider implements GitProvider, Disposable {
 						message = `WIP: ${matchSummary}`;
 						summary = wipSummary;
 					} else {
-						message = matchSummary;
+						message = matchSummary.length ? matchSummary : wipSummary;
 						summary = message ? message.split('\n', 1)[0] : wipSummary;
 					}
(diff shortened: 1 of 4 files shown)


src/plus/webviews/graph/graphWebview.ts
```diff
@@ -257,16 +279,48 @@ export class GraphWebview extends WebviewWithConfigBase<State> {
...
+		committer: commit.committer,
+		type: getCommitType(commit)
 	}));
 }
 
+// TODO: Move constant to a better home
+const enum CommitType {
+	CommitNode = 'commit-node',
+	StashNode = 'stash-node',
+}
```
ericf-axosoft, line 297: We could modify our `GraphContainer.d.ts` to to have those constants and then import them here.
We can do something similar as what we have done with `GraphZoneType`.

Then we will be able to remove this enum definition here (in 294 line).

ericf-axosoft (changes requested):

axosoft-ramint:
Now waiting on [this PR to land first](https://github.com/gitkraken/GitKrakenComponents/pull/17).

Note: importing from the graph library directly in `graphWebView` causes weird, unrelated build errors, so I had to import it in the wrapper and do a mapping there instead. Feels like a hacky workaround, so advice on a better way is appreciated.

src/plus/webviews/graph/graphWebview.ts
axosoft-ramint: Addressed in commit: `Map to types in library`, but, I don't like the way I had to do it. I could not import from `gitkraken-components` in `graphWebView` without getting weird build errors that broke the build and the webview, so I had to do it in the wrapper and do a mapping there.

We will need to review/merge this PR first: https://github.com/gitkraken/GitKrakenComponents/pull/17

ericf-axosoft: We don't use enums for now in our library. Once we will convert our code to typescript we will be able to change that.
You don't need to do any mapping here. 

My idea was: 

- In `GraphContainer.d.ts` add following code instead of your changes:

```
import {
  commitNodeType,
  mergeConflictNodeType,
  mergeNodeType,
  stashNodeType,
  unsupportedRebaseWarningNodeType,
  workDirType
} from '../../domain/commit/CommitConstants';

...

export type WorkDirType = typeof mergeConflictNodeType
  | typeof unsupportedRebaseWarningNodeType
  | typeof workDirType;

export type CommitType = typeof commitNodeType
  | typeof mergeNodeType
  | typeof stashNodeType
  | WorkDirType;
```

- Create a new file `/src/domain/commit/CommitConstants.d.ts` with following code:
```
export const commitNodeType = 'commit-node';
export const mergeConflictNodeType = 'merge-conflict-node';
export const mergeNodeType = 'merge-node';
export const unsupportedRebaseWarningNodeType = 'unsupported-rebase-warning-node';
export const stashNodeType = 'stash-node';
export const workDirType = 'work-dir-changes';

```

- Then in `Gitlens` side, use directly the constant. For example in `graphWebView.ts`:
```
import { commitNodeType } from '@gitkraken/gitkraken-components/lib/domain/commit/CommitConstants';
...
if (GitCommitModel.isStash(commit)) {
   type = commitNodeType;
}
...
```

ericf-axosoft (approved):
I have tested the changes and they work. I do not know if there may be cases that are not contemplated but for now it seems fine to me. If we find a case, we can create another JIRA issue to fix it.

@d13 could you please take a look and let us know if it is ok for you?

axosoft-ramint:
> Now waiting on [this PR to land first](https://github.com/gitkraken/GitKrakenComponents/pull/17).
> 
> Note: importing from the graph library directly in `graphWebView` causes weird, unrelated build errors, so I had to import it in the wrapper and do a mapping there instead. Feels like a hacky workaround, so advice on a better way is appreciated.

Dependency PR has landed. This is once again ready for review.

src/env/node/git/localGitProvider.ts
```diff
@@ -3320,21 +3322,28 @@ export class LocalGitProvider implements GitProvider, Disposable {
...
 					if (match.groups.wip) {
-						message = `WIP: ${match.groups.summary.trim()}`;
-						summary = `WIP on ${onRef}`;
+						message = `WIP: ${matchSummary}`;
+						summary = wipSummary;
 					} else {
-						message = match.groups.summary.trim();
-						summary = message.split('\n', 1)[0] ?? '';
+						message = matchSummary;
+						summary = message ? message.split('\n', 1)[0] : wipSummary;
```
miggy-e, line 3338: if the message is empty should we also set it to the wipSummary? In this case, _message is `undefined` and _summary is `WIP on development`
image

miggy-e (changes requested):

src/env/node/git/localGitProvider.ts
axosoft-ramint: Nice catch! I had logic set up for this but it looks like it may have gotten lost when I rebased today. Mind trying it again?

miggy-e (approved):

PR #2105, opened 2022-07-20
gitlens-258 graph colors

miggy-e:
https://gitkraken.atlassian.net/browse/GITLENS-258
changes:
- add mechanism to update the graph colors when the theme is updated in vscode by using a mutation observer. This might be redundant with the current theme.js mutation observer for the gitlens app, but I think this application of it is different enough that justifies a separate instance
- gets the correct font colors
- calculates graph and ref mixed colors that originally was done in the theme compiler.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3010,7 +3010,8 @@
 					"gitlens.graph.columnColors": {
 						"type": "array",
 						"items": {
-							"type": "string"
+							"type": "string",
+							"pattern": "^#(?:[0-9a-fA-F]{3,4}){1,2}$"
 						},
 						"default": ["#15a0bf", "#0669f7", "#8e00c2", "#c517b6", "#d90171", "#cd0101", "#f25d2e", "#f2ca33", "#7bd938", "#2ece9d"],
 						"markdownDescription": "Specifies the colors used for the different columns in the graph",
--- a/package.json
+++ b/package.json
@@ -3006,6 +3006,18 @@
 						"markdownDescription": "Specifies the number of items to show in a each page when paginating a graph list. Use 0 to specify no limit",
 						"scope": "window",
 						"order": 20
+					},
+					"gitlens.graph.columnColors": {
+						"type": "array",
+						"items": {
+							"type": "string"
+						},
+						"default": ["#15a0bf", "#0669f7", "#8e00c2", "#c517b6", "#d90171", "#cd0101", "#f25d2e", "#f2ca33", "#7bd938", "#2ece9d"],
+						"markdownDescription": "Specifies the colors used for the different columns in the graph",
+						"scope": "window",
+						"order": 30,
+						"maxItems": 10,
+						"minItems": 10
 					}
 				}
 			},
--- a/src/webviews/apps/plus/graph/GraphWrapper.tsx
+++ b/src/webviews/apps/plus/graph/GraphWrapper.tsx
@@ -35,11 +35,10 @@ export interface GraphWrapperProps extends State {
 const getCssVariables = (): CssVariables => {
     const body = document.body;
     const computedStyle = window.getComputedStyle(body);
-	const isLightTheme = body.className.includes('vscode-light') || body.className.includes('vscode-high-contrast-light');
 
 	return {
         '--app__bg0': computedStyle.getPropertyValue('--color-background'),
-        '--panel__bg0': isLightTheme ? computedStyle.getPropertyValue('--color-background--darken-05') : computedStyle.getPropertyValue('--color-background--lighten-05'),
+        '--panel__bg0': computedStyle.getPropertyValue('--graph-panel-bg'),
 		'--text-selected': computedStyle.getPropertyValue('--color-foreground'),
 		'--text-normal': computedStyle.getPropertyValue('--color-foreground-85'),
 		'--text-secondary': computedStyle.getPropertyValue('--color-foreground-65'),
@@ -47,21 +46,18 @@ const getCssVariables = (): CssVariables => {
 		'--text-accent': computedStyle.getPropertyValue('--color-link-foreground'),
 		'--text-inverse': computedStyle.getPropertyValue('--vscode-input-background'),
 		'--text-bright': computedStyle.getPropertyValue('--vscode-input-background'),
-		... getGraphColors(computedStyle.getPropertyValue('--color-background')),
+		... getGraphColors(computedStyle),
     };
 };
 
-const getGraphColors = (bgColor: string): CssVariables => {
+const getGraphColors = (computedStyle: CSSStyleDeclaration): CssVariables => {
 	// mixed colors for ref labels (ref label color + graph bg color)
 	// making the ref label color transparent is not an option since ref labels overlap the ref line
 	// and can potentially overlap each other, e.g., when a truncated ref is expanded.
 	const mixedGraphColors: CssVariables = {};
-	const graphColors = [
-		'#15a0bf', '#0669f7', '#8e00c2', '#c517b6', '#d90171', '#cd0101', '#f25d2e', '#f2ca33', '#7bd938', '#2ece9d'
-	];
-	for (let i = 0; i < graphColors.length; i++) {
+	for (let i = 0; i < 10; i++) {
 		for (const mixInt of [15,25,45,50]) {
-			mixedGraphColors[`--graph-color-${i}-bg${mixInt}`] = mix(bgColor, graphColors[i], mixInt / 100).hex();
+			mixedGraphColors[`--graph-color-${i}-bg${mixInt}`] = computedStyle.getPropertyValue(`--graph-color-${i}-bg${mixInt}`);
 		}
 	}
 	return mixedGraphColors;
@@ -263,7 +259,7 @@ export function GraphWrapper({
 						{mainWidth !== undefined && mainHeight !== undefined && (
 							<GraphContainer
 								columnsSettings={graphColSettings}
-								cssVariables={cssVariables()}
+								cssVariables={cssVariables}
 								graphRows={graphList}
 								height={mainHeight}
 								hasMoreCommits={logState?.hasMore}
--- a/src/webviews/apps/plus/graph/GraphWrapper.tsx
+++ b/src/webviews/apps/plus/graph/GraphWrapper.tsx
@@ -32,14 +32,39 @@ export interface GraphWrapperProps extends State {
 // Copied from original pushed code of Miggy E.
 // TODO: review that code as I'm not sure if it is the correct way to do that in Gitlens side.
 // I suppose we need to use the GitLens themes here instead.
-export const getCssVariables = (): CssVariables => {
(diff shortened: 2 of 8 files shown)


ericf-axosoft (changes requested):
@d13 please, take a look at this.

src/webviews/apps/plus/graph/GraphWrapper.tsx
```diff
@@ -32,16 +33,41 @@ export interface GraphWrapperProps extends State {
 // Copied from original pushed code of Miggy E.
 // TODO: review that code as I'm not sure if it is the correct way to do that in Gitlens side.
 // I suppose we need to use the GitLens themes here instead.
-export const getCssVariables = (): CssVariables => {
+const getCssVariables = (): CssVariables => {
     const body = document.body;
     const computedStyle = window.getComputedStyle(body);
-    return {
+	const isLightTheme = body.className.includes('vscode-light') || body.className.includes('vscode-high-contrast-light');
```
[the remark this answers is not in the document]
miggy-e: Should I move the computeGraphColors there as well then?

miggy-e:
moved graph logic to theme.ts and am making use of app theme-watcher instead of graph specific. still todo: moving graph colors to user settings

src/webviews/apps/plus/graph/GraphWrapper.tsx
```diff
@@ -223,7 +251,7 @@ export function GraphWrapper({
 					<h2>Repository: {currentRepository}</h2>
 					<GraphContainer
 						columnsSettings={graphColSettings}
-						cssVariables={getCssVariables()}
+						cssVariables={cssVariables}
 						graphRows={graphList}
```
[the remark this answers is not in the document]
miggy-e: Ahh, the theme Opacity factor isn't related to cssVariables but is a separate prop. It seems like we probably will need to keep isLightTheme logic in the graph then?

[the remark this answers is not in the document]
ericf-axosoft: The opacity factor is a number and the graph is using a defined `dark` value by default which is in our case `1`. See constant in `GraphConstants.js`:
```
export const OPACITY_FACTOR_BY_THEME = {
  dark: 1,
  light: 0.5
};
```

I'm not saying to move the `isLightTheme` logic to the graph but just that we need to pass a value from Gitlens for that property.

src/webviews/apps/plus/graph/GraphWrapper.tsx
```diff
@@ -32,13 +32,43 @@ export interface GraphWrapperProps extends State {
...
+const getStyleProps = ():{cssVariables: CssVariables, themeOpacityFactor: number } => {
 	const body = document.body;
-	const computedStyle = window.getComputedStyle(body);
+    const computedStyle = window.getComputedStyle(body);
 	return {
-		'--app__bg0': computedStyle.getPropertyValue('--color-background'),
-		// note that we should probably do something theme-related here, (dark theme we lighten, light theme we darken)
-		'--panel__bg0': computedStyle.getPropertyValue('--color-background--lighten-05'),
+		cssVariables: getCssVariables(),
+		themeOpacityFactor: computedStyle.getPropertyValue('--graph-theme') === 'dark' ? 1 : 0.5,
```
ericf-axosoft, line 71: Why not add a new specific variable in `theme.ts` with that logic?

It could be something like in `theme.ts`: 
`bodyStyle.setProperty('--graph-theme-opacity-factor', isLightTheme ? 0.5 : 0);`

And then in `GraphWrapper.jsx`:
`themeOpacityFactor: computedStyle.getPropertyValue('--graph-theme-opacity-factor'),`

src/webviews/apps/plus/graph/graph.tsx
```diff
@@ -96,6 +96,12 @@ export class GraphApp extends App<State> {
 		}
 	}
 
+	protected override onThemeUpdated() {
+		console.log('zz theme updated, updating');
+		console.log('zz this', this);
```
ericf-axosoft, line 101: You forget to remove the console logs

ericf-axosoft (changes requested):

src/webviews/apps/plus/graph/GraphWrapper.tsx
miggy-e: I thought that it was a little weird if we put something in the CSS that wasn't ever planned to be used by CSS, and thought graph-theme dark/light would be more clean, but can do it this way as well

src/webviews/apps/plus/graph/graph.tsx
miggy-e: thanks

src/webviews/apps/plus/graph/graph.tsx
```diff
@@ -91,11 +94,40 @@ export class GraphApp extends App<State> {
...
 				super.onMessageReceived?.(e);
 		}
 	}
 
+	private getGraphColors(config: GraphConfig | undefined): CssVariables {
+		// this will be called on theme updated as well as on config updated since it is dependent on the column colors from config changes and the background color from the theme
+		const body = document.body;
+    	const computedStyle = window.getComputedStyle(body);
+		const bgColor = computedStyle.getPropertyValue('--color-background');
+		const columnColors = ((config?.columnColors) != null) ? config.columnColors : ['#00bcd4', '#ff9800', '#9c27b0', '#2196f3', '#009688', '#ffeb3b', '#ff5722', '#795548'];
```
ericf-axosoft, line 116: It's ok to load colors from the config but we are not setting the all the `cssVariables` needed to paint all edge columns. See my comment: https://github.com/gitkraken/vscode-gitlens/pull/2105#discussion_r926643618

Now, if we change one color from the settings, only the color of the branch is changed and not the entire edge column:
https://user-images.githubusercontent.com/90025366/180764253-0bbe6bc3-d432-4c79-822e-e4a5348141fe.mov

ericf-axosoft (changes requested):

src/webviews/apps/plus/graph/graph.tsx
ericf-axosoft: @miggy-e  in `base.jsonc` (GKC side) and `columnColors.less` (library side) files we use also `--graph-color-0-f10`, `--graph-color-0-f50`, `--graph-color-1-f10`, etc. 

For example in base.jsonc we have:
```
// graph color variations
"graph-color-5-f10": "fade(@graph-color-5, 10%)",
"graph-color-6-f10": "fade(@graph-color-6, 10%)",
"graph-color-7-f10": "fade(@graph-color-7, 10%)",
"graph-color-8-f10": "fade(@graph-color-8, 10%)",
"graph-color-9-f10": "fade(@graph-color-9, 10%)",
"graph-color-0-f10": "fade(@graph-color-0, 10%)",
"graph-color-1-f10": "fade(@graph-color-1, 10%)",
"graph-color-2-f10": "fade(@graph-color-2, 10%)",
"graph-color-3-f10": "fade(@graph-color-3, 10%)",
"graph-color-4-f10": "fade(@graph-color-4, 10%)",
"graph-color-5-f50": "fade(@graph-color-5, 50%)",
"graph-color-6-f50": "fade(@graph-color-6, 50%)",
"graph-color-7-f50": "fade(@graph-color-7, 50%)",
"graph-color-8-f50": "fade(@graph-color-8, 50%)",
"graph-color-9-f50": "fade(@graph-color-9, 50%)",
"graph-color-0-f50": "fade(@graph-color-0, 50%)",
"graph-color-1-f50": "fade(@graph-color-1, 50%)",
"graph-color-2-f50": "fade(@graph-color-2, 50%)",
"graph-color-3-f50": "fade(@graph-color-3, 50%)",
"graph-color-4-f50": "fade(@graph-color-4, 50%)",
...
```

Do you think we should include them too or it is not necessary?

src/webviews/apps/plus/graph/graph.tsx
```diff
@@ -116,9 +116,14 @@ export class GraphApp extends App<State> {
...
 		const mixedGraphColors: CssVariables = {};
 		for (let i = 0; i < columnColors.length; i++) {
+			mixedGraphColors[`--graph-color-${i}`] = columnColors[i];
+			mixedGraphColors[`--column-${i}-color`] = columnColors[i];
 			for (const mixInt of [15,25,45,50]) {
 				mixedGraphColors[`--graph-color-${i}-bg${mixInt}`] = mix(bgColor, columnColors[i], mixInt);
 			}
+			for (const mixInt of [10,50]) {
+				mixedGraphColors[`--graph-color-${i}-f${mixInt}`] = opacity(columnColors[i], mixInt);
+			}
```
miggy-e, line 126: @ericf-axosoft I added them here

src/webviews/apps/plus/graph/graph.tsx
miggy-e: I added them in line 124

ericf-axosoft: Ok, I have not seen it.

src/webviews/apps/plus/graph/graph.tsx
ericf-axosoft: Ok, I have not seen it.

package.json
```diff
@@ -3006,6 +3006,18 @@
 						"markdownDescription": "Specifies the number of items to show in a each page when paginating a graph list. Use 0 to specify no limit",
 						"scope": "window",
 						"order": 20
+					},
+					"gitlens.graph.columnColors": {
+						"type": "array",
+						"items": {
+							"type": "string"
+						},
+						"default": ["#15a0bf", "#0669f7", "#8e00c2", "#c517b6", "#d90171", "#cd0101", "#f25d2e", "#f2ca33", "#7bd938", "#2ece9d"],
```
ericf-axosoft, line 3015: 🤔 What would happen if the user enters an incorrect color here?  

I think the component doesn't check if the color is valid or not. We need to take a look at this and if it is the case, add code to protect this on the library side (on a new JIRA task).

miggy-e: the graph breaks, I'll investigate forcing formatting the columns as colors or checking the colors before I add them to the CSS variables as well as creating a new issue to validate the css variables.

miggy-e: https://gitkraken.atlassian.net/browse/GITLENS-283

miggy-e: @ericf-axosoft I added a regex on this side to ensure the column color array has an rgb string. We should still do the library-side fix as well though

ericf-axosoft:
Apart of the error commented https://github.com/gitkraken/vscode-gitlens/pull/2105#discussion_r929151546 that will be address in another issue and the above small thing, it seems that is working fine

ericf-axosoft (approved):

axosoft-ramint (approved):
Not sure if you're waiting on other reviewers. Feel free to merge if not. 👍🏼

PR #2107, opened 2022-07-21
GITLENS-276 - Performance issue when resizing graph columns when using "onColumnResized" event in GitLens (Gitlens side fix)

ericf-axosoft:
# Summary of changes
- Removed not needed debounce in `onColumnChange` of `GraphWrapper` component in `graph.tsx` (as we fixed that in the `GitkrakenComponents` library side).
- Removed the call to `notifyDidChangeState` in `changeColumn` procedure of `graphWebView.ts` as we don't need to retrieve again all stashes, tags and branches to refresh the view in that case. We only need to store the column sizes.

Notes:
- This PR works with https://github.com/gitkraken/GitKrakenComponents/pull/16 (`GitkrakenComponents` library side fix).

# JIRA issue
[GITLENS-276](https://gitkraken.atlassian.net/browse/GITLENS-276)


ericf-axosoft:
@d13 please take a look at this

d13:
While I 100% agree that we shouldn't be sending the entire state for this change, removing the update entirely leaves the state out of sync which is a bug.

I don't see any harm in leaving the debounce.

d13:
I've created an issue to address the state updates: https://github.com/gitkraken/vscode-gitlens/issues/2109

ericf-axosoft:
@d13 ok, so I will close this PR without merging changes. We can leave the `debounce` if you want and regarding the issue of the state, you have already created an issue 👍

PR #2116, opened 2022-07-27
GITLENS-263 Review the logic of retrieving and set the input properties "remotes", "heads" and "tags" to the "GraphContainer" component in GitLens side

axosoft-ramint:
Fixes: [GITLENS-263](https://gitkraken.atlassian.net/browse/GITLENS-263)

Issues addressed:
 - Loading all commits rather than just the checked-out ref and its history
 - Calculating remote branch names correctly so that they combine in ref groups (like in GKC)
 - Identifying annotated tags
 - Getting remote urls correctly
 - Remote avatarUrl support for GitHub remotes (others still needed)
 - Displaying merge commit nodes correctly in the graph node zone of the graph
 - Fixes stashes disappearing when choosing to "load more" commits

Notes:
- When using `--all` with `git log`, using a ref SHA as a cursor does not work. It will always load in the commits leading up to the ref. This breaks the "get more" functionality of the log. Therefore, when this option is used, currently we are opting to use `--skip` instead and just skip a number of commits based on the `count` of the previous log.
- When using the `--all` option, index commits for stashes, and stashes themselves i.e. everything in `refs/stashes`, are loaded in. This causes duplicates when merging with stash commits from `getStashCommits`, so we include a de-duplication/filtering step to ensure one commit per sha.

### DIFF
--- a/src/env/node/git/git.ts
+++ b/src/env/node/git/git.ts
@@ -690,7 +690,7 @@ export class Git {
 			ordering,
 			similarityThreshold,
 			since,
-			skip,
+			until,
 		}: {
 			all?: boolean;
 			argsOrFormat?: string | string[];
@@ -700,7 +700,7 @@ export class Git {
 			ordering?: string | null;
 			similarityThreshold?: number | null;
 			since?: string;
-			skip?: number;
+			until?: string;
 		},
 	) {
 		if (argsOrFormat == null) {
@@ -731,6 +731,10 @@ export class Git {
 			params.push(`--since="${since}"`);
 		}
 
+		if (until) {
+			params.push(`--until="${until}"`);
+		}
+
 		if (!merges) {
 			params.push('--first-parent');
 		}
@@ -748,10 +752,6 @@ export class Git {
 
 		if (ref && !GitRevision.isUncommittedStaged(ref)) {
 			params.push(ref);
-		}
-
-		if (skip) {
-			params.push(`--skip=${skip}`);
 		}
 
 		return this.git<string>(
--- a/src/env/node/git/git.ts
+++ b/src/env/node/git/git.ts
@@ -690,6 +690,7 @@ export class Git {
 			ordering,
 			similarityThreshold,
 			since,
+			skip,
 		}: {
 			all?: boolean;
 			argsOrFormat?: string | string[];
@@ -699,6 +700,7 @@ export class Git {
 			ordering?: string | null;
 			similarityThreshold?: number | null;
 			since?: string;
+			skip?: number;
 		},
 	) {
 		if (argsOrFormat == null) {
@@ -746,6 +748,10 @@ export class Git {
 
 		if (ref && !GitRevision.isUncommittedStaged(ref)) {
 			params.push(ref);
+		}
+
+		if (skip) {
+			params.push(`--skip=${skip}`);
 		}
 
 		return this.git<string>(
--- a/src/env/node/git/localGitProvider.ts
+++ b/src/env/node/git/localGitProvider.ts
@@ -2090,7 +2090,7 @@ export class LocalGitProvider implements GitProvider, Disposable {
 			ordering?: string | null;
 			ref?: string;
 			since?: string;
-			skip?: number;
+			until?: string;
 		},
 	): Promise<GitLog | undefined> {
 		const cc = Logger.getCorrelationContext();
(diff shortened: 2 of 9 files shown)


axosoft-ramint:
Getting some alerts pointing to files I didn't touch in this PR. Not sure what I broke. Will have a look tomorrow.

ericf-axosoft (commented):
Tried with different repositories and compared with the GKC version and it seems that are displayed correctly now. Great work 👍 

The only think I have noticed is following warnings related with duplicated keys:
Captura de pantalla 2022-07-27 a las 11 48 52

axosoft-ramint:
Definitely could use a review from @eamodio and @d13 on this one as I am touching core Git, provider logic, and remote models and there may be a simpler approach for getting more commits or repo avatars that I'm not aware of.

axosoft-ramint:
> Tried with different repositories and compared with the GKC version and it seems that are displayed correctly now. Great work 👍
> 
> The only think I have noticed is following warnings related with duplicated keys: Captura de pantalla 2022-07-27 a las 11 48 52

@ericf-axosoft This is a bug on the component side with key assignment. I'll put together a PR to fix it momentarily.

eamodio:
@axosoft-ramint IMO, this PR is going to need some more thought and work:

 - Since `log --all` return all refs, we need to fix up the parsing to ensure that stashes get properly identified. We probably also need to allow for exluding refs (at a min remotes) too as we will need to add filtering soon.
 - We need to think about/figure out how to make the equivalent work for virtual repositories (e.g. GitHub)
 - Also, not a fan of the `skip` option, but not sure if we have a better alternative for paging -- it really worries me because if the set of commits change, then the skip is wrong.
- Could you go in more detail on the `avatarUrl` for a remote? Remotes in GitLens already have an icon (but not related to the "owner", but related to type. I can kind of see the desire to have a remote-specific (e.g. owner) version of that, but it needs to be supported across all services that we support (or at least fallback to the service icon)

We can talk more about this.

axosoft-ramint:
> @axosoft-ramint IMO, this PR is going to need some more thought and work:
>

Thanks for the feedback! Some initial thoughts:
 
> * Since `log --all` return all refs, we need to fix up the parsing to ensure that stashes get properly identified. We probably also need to allow for exluding refs (at a min remotes) too as we will need to add filtering soon.

Yep, I've noticed it even brings in intermediary "merge" commits when stashes are created. I'm currently doing all of the cleanup in `combineAndFilterStashCommits`, but it would be nicer to have a more clever log command for this stuff if we can think of one.

> * We need to think about/figure out how to make the equivalent work for virtual repositories (e.g. GitHub)

Could definitely use some discussion on how virtual repos might break the assumptions here i.e. how stashes and the git log in general work for them.

> * Also, not a fan of the `skip` option, but not sure if we have a better alternative for paging -- it really worries me because if the set of commits change, then the skip is wrong.

I'm not a fan of it either. It feels really "brittle". But I could not find an alternative cursor to work with the --all option, since using the sha of the last commit didn't really work.

> * Could you go in more detail on the `avatarUrl` for a remote? Remotes in GitLens already have an icon (but not related to the "owner", but related to type. I can kind of see the desire to have a remote-specific (e.g. owner) version of that, but it needs to be supported across all services that we support (or at least fallback to the service icon)

Yep, I did notice that we have icons already for remotes based on type/hosting service. I was trying in this case to bring in an "owner"-specific one but it really only seems to work with GitHub remotes, so I'm not sure if it adds a lot of value other than matching GK Client behavior.

I could add in the GitLens icons for remotes as a minimum, but on top of that, do we want this GitHub-specific owner avatar feature in or should we hold off on it?

axosoft-ramint:
**Updates:**
- Removed the avatar url logic from remotes for now (will approach in a separate PR)
- Updated paging mechanism to use `--until` with UNIX timestamp of the last commit (+1 second for buffering)
   - Seeking a more elegant approach here: https://github.com/git-for-windows/git/discussions/3969
- Fixed a bug (happens in remote repos) where no commits would show if the stash of a repo is empty

axosoft-ramint:
**Note:** This should work for remote repositories (try testing with the GitHub repositories extension and opening a repo from GitHub), but currently only loads commits as if you were soloing. I believe this is an API limitation and will set up a separate task to bring in other branches/refs for those repos.

axosoft-ramint:
@eamodio Let me know if we should hold off on merging this until we hear back from the discussion group, or if our date strategy is sufficient for now.

eamodio:
I think we should move forward

axosoft-ramint:
> I think we should move forward

Sounds good, thanks! Feel free to give it another review pass before merge.

axosoft-ramint:
@ericf-axosoft @miggy-e If you wouldn't mind giving this a review/test today, I'd appreciate it!

ericf-axosoft:
> @ericf-axosoft @miggy-e If you wouldn't mind giving this a review/test today, I'd appreciate it!

Sure. I was just taking a look at your changes and doing some tests with different repositories.

ericf-axosoft (approved):

miggy-e (approved):

axosoft-ramint:
Going to merge this so we can move forward. If we encounter issues with the date paging strategy, we can update it in another task.

PR #2135, opened 2022-08-04
Uses VS Code as core.editor

### LINKED ISSUE #2134: Use VS Code as core.editor
A superset of #2084, currently git's global `core.editor` is used to edit commit messages instead of VSCode in scenarios merge, pull, cherry-pick etc. I would like to be able to use VSCode to edit commit messages without changing global config.

ShafinKhadem:
# Description

Closes #2134.

When using git commands from terminal, one may prefer to use terminal editors to edit commit messages. But when using VS Code, it's better to use VS Code to edit the commit messages. This PR ensures the usage of VS Code as `core.editor` whenever a terminal command is used.

# Checklist


eamodio (changes requested):
Thanks! I like this change, but I am concerned about changing the existing behavior -- as users might want to use their git config. We should make this an opt-in boolean setting -- `gitlens.terminal.overrideGitEditor`. Please add it in the `terminal` section of the config.

ShafinKhadem:
@eamodio I have added `gitlens.terminal.overrideGitEditor` config in a separate commit. However, please consider not adding this config because:
- We aren't letting user configure `sequence.editor` for rebase. `core.editor` shouldn't be different.
- A VS Code extension doesn't need to (or shouldn't) open another editor. If user wants to use another editor, running a terminal command is always an option.

eamodio:
That's a good point, let's make it on by default then, but users can still opt-out.

ShafinKhadem:
> That's a good point, let's make it on by default then, but users can still opt-out.

I have kept it on by default.

eamodio:

Thank you!

gutwrinch (approved):

PR #2150, opened 2022-08-17
Add avatarUrl to remote model, send to graph for GitHub remotes

axosoft-ramint:
This is intended to be the start of sending owner avatar urls for remotes to the graph when one is available. The avatar url is added to the remote model and implemented in GitHub as a first hosting service for support:

image

**Note:** This does not include fallback to hosting service icons/SVGs, which will be implemented in a separate PR.


eamodio:
@axosoft-ramint Thanks!

PR #2154, opened 2022-08-19
Addition of Work In Progress row

axosoft-ramint:
This PR attaches a "WIP" row to the top of the graph when there are changes in the working directory.

**Note:** To support this, I have utilized `git-ls-files`: https://git-scm.com/docs/git-ls-files/2.9.5

I also generated a new "work dir stats" model which represents the number of added, deleted and modified files.

**Additional notes:**
- We currently aren't sending the work dir stats into the graph as we cannot handle renames just yet. They are incredibly tricky to detect with unstaged files using git and it needs to be thought out more.
- We are using ls-files instead of diff shortstat because git diff does not pick up unstaged stuff by itself (which is a shame, because git diff DOES have rename detection).

### DIFF
--- a/src/plus/webviews/graph/graphWebview.ts
+++ b/src/plus/webviews/graph/graphWebview.ts
@@ -783,7 +783,7 @@ export class GraphWebview extends WebviewBase<State> {
 		if (!this.isReady || !this.visible) return false;
 
 		return this.notify(DidChangeWorkDirStatsNotificationType, {
-			workDirStats: await this.getWorkDirStats() || {added: 0, deleted: 0, modified: 0 },
+			workDirStats: await this.getWorkDirStats() ?? {added: 0, deleted: 0, modified: 0 },
 		});
 	}
 
@@ -947,11 +947,11 @@ export class GraphWebview extends WebviewBase<State> {
 		}
 
 		const status: GitStatus | undefined = await this.container.git.getStatusForRepo(this.repository.path);
-		const workingTreeStatus = status?.getDiffStatus() || undefined;
+		const workingTreeStatus = status?.getDiffStatus();
 		return {
-			added: workingTreeStatus?.added || 0,
-			deleted: workingTreeStatus?.deleted || 0,
-			modified: workingTreeStatus?.changed || 0,
+			added: workingTreeStatus?.added ?? 0,
+			deleted: workingTreeStatus?.deleted ?? 0,
+			modified: workingTreeStatus?.changed ?? 0,
 		};
 	}
 
@@ -1046,7 +1046,7 @@ export class GraphWebview extends WebviewBase<State> {
 				header: this.getColumnHeaderContext(columns),
 			},
 			nonce: this.cspNonce,
-			dirStats: await this.getWorkDirStats() || {added: 0, deleted: 0, modified: 0 },
+			dirStats: await this.getWorkDirStats() ?? {added: 0, deleted: 0, modified: 0 },
 		};
 	}
 
--- a/src/plus/webviews/graph/graphWebview.ts
+++ b/src/plus/webviews/graph/graphWebview.ts
@@ -1,13 +1,20 @@
 import type {
 	ColorTheme,
 	ConfigurationChangeEvent,
-	Disposable,
 	Event,
 	StatusBarItem,
 	WebviewOptions,
 	WebviewPanelOptions,
 } from 'vscode';
-import { CancellationTokenSource, EventEmitter, MarkdownString, StatusBarAlignment, ViewColumn, window } from 'vscode';
+import {
+	CancellationTokenSource,
+	Disposable,
+	EventEmitter,
+	MarkdownString,
+	StatusBarAlignment,
+	ViewColumn,
+	window,
+} from 'vscode';
 import type { CreatePullRequestActionContext } from '../../../api/gitlens';
 import { getAvatarUri } from '../../../avatars';
 import type {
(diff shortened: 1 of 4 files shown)


src/plus/webviews/graph/graphWebview.ts
```diff
@@ -140,9 +141,13 @@ export class GraphWebview extends WebviewWithConfigBase<State> {
 
 		let commits: GitCommit[] | undefined;
 		if (ref != null) {
-			const commit = await this.selectedRepository?.getCommit(ref);
-			if (commit != null) {
-				commits = [commit];
+			if (ref === 'work-dir-changes') {
+				// TODO: Add action for selecting WIP row.
+			} else {
```
axosoft-ramint, line 146: @eamodio @d13 I left the WIP row selection logic as TODO because I'm not sure yet how to interact with the SCM view and for the sake of speed, I was wondering if either of you could pick this part up when you have a chance. 🙏🏼

src/env/node/git/localGitProvider.ts
```diff
@@ -4060,4 +4062,45 @@ export class LocalGitProvider implements GitProvider, Disposable {
...
+			const [status, file] = line.split(' ');
+			let incrementCount = false;
+			if (statusByFile[file] === undefined) {
+				statusByFile[file] = status;
+				incrementCount = true;
+			} else if (statusByFile[file] === 'C' && status !== 'C') {
+				statusByFile[file] = status;
+				stats.modified = stats.modified - 1;
+				incrementCount = true;
+			}
```
axosoft-ramint, line 4091: Note on this part: I noticed in git-ls-files that if a file is deleted, it shows up as both "modified" and "deleted". In this case we want to remove it from "modified" and count it as "deleted" instead.

axosoft-ramint:
@eamodio Need a review pass on this since I'm touching core git logic.

miggy-e:
@axosoft-ramint Heads up you'll want to bump the current version of the component-library and change the package.json to use the new version, probably v1.0.0-rc.4

eamodio:
@axosoft-ramint Since this isn't a requirement for 12.2 and I'm still swamped in 12.2 task, I won't get to review this until later in the week.

axosoft-ramint:
@eamodio This PR is updated. Would you mind giving it a look when you have the chance? Thanks!

axosoft-ramint:
Looks like our new paging strategy broke this after rebasing it. When paging, we get a WIP node for every page of commits and the edges break on it. Having a look.

src/env/node/git/localGitProvider.ts
```diff
@@ -1839,6 +1843,23 @@ export class LocalGitProvider implements GitProvider, Disposable {
...
+					email: '',
+					date: new Date().getTime(),
+					message: '',
+					type: GitGraphRowType.Working,
+					heads: [],
+					remotes: [],
+					tags: [],
+				});
+			}
+
```
eamodio, line 1862: This needs to be moved into the outer function -- since we only want this called once when we get the original set of rows, not on each page.

This is also a "refresh"/performance challenge -- since it is much more likely that the WIP will change than all the rows. So I think we might actually have to move this out the provider altogether, and instead add it into `graphWebview` so that it could not only be included in the original rows payload, but we could have a specific single row update when it changes

eamodio (changes requested):
See inline comment. If you want to chat about this let me know.

src/env/node/git/localGitProvider.ts
axosoft-ramint: @eamodio What if I just checked (in the `if` clause on 1848) if no cursor has been set?

i.e...
```
if (!cursor && (workingTreeStatus?.staged || workingTreeStatus?.unstaged)) {
...
```

That would at least fix the "only on the first set" issue, but the performance issue you mentioned would still be there I suppose.

In that case we may need separate state update logic that adds/removes/updates the date on the WIP row.

eamodio: Yeah -- we'd have to plumb through a new way to splice in just that row (at least from the GitLens -> webview side -- we could just pass the full set of rows to the component again

axosoft-ramint: I think I have an idea. What if I just add in the WIP node on the component side? We already have a prop for work dir stats that can be passed in from GitLens. We could just listen for workdir changes and update that prop and pass it through from this side, and on the component side, staple a WIP node to the top whenever the workdirchanges prop is non-empty, and update that node whenever the prop changes. WDYT?

eamodio: Sounds great! We'd still need to move the workdir calc out of the provider and into the `grapWebview` or add a new wrapper method on the provider to give out the data (and events) that we need -- though I think that might be overkill

axosoft-ramint:
**NOTE:** Updated this to just send `workDirStats` as a prop and update it when needed. We will take care of adding the WIP node row on the component side: https://github.com/gitkraken/GitKrakenComponents/pull/89

eamodio:
Quality gate Quality gate: **_Failed_** 
 


            
 #### 9 Issues 
* Code issue 9 Code issues 

 [Details](https://app.embold.io/organization/gh/gitkraken/repositories/e23a25993521c5b78d5e00d144892422/pull-request/2154)

axosoft-ramint:
> Quality gate Quality gate: **_Failed_**
> 
> #### 9 Issues
> * Code issue 9 Code issues
> 
> [Details](https://app.embold.io/organization/gh/gitkraken/repositories/e23a25993521c5b78d5e00d144892422/pull-request/2154)

@eamodio Some of the warnings from the Embold analysis make sense, and I've updated the code for those, but...

image

...all I did was move an import from the type list to the non-type list. Why would this trigger such a warning?

image

We do this everywhere already in the existing code? I counted 131 other instances, including one in `getState` right above.

image

We also did this everywhere already in the existing code.

Should I change these anyway?

eamodio:
Please ignore any Embold checks -- I was trying an evaluation and it apparently added all this without asking.

axosoft-ramint:
Component-side logic has landed. Once this merges I can bump the library version and it should enable the feature.

PR #2190, opened 2022-09-02
Send platform to the graph

miggy-e:
Retrieves platform and translate it into one of the expected platforms for the container.

Note: tested on a windows machine, will need to be tested on a Mac prior to approval

https://gitkraken.atlassian.net/browse/GITLENS-289

### DIFF
--- a/src/webviews/apps/plus/graph/GraphWrapper.tsx
+++ b/src/webviews/apps/plus/graph/GraphWrapper.tsx
@@ -2,11 +2,13 @@ import GraphContainer, {
 	type CssVariables,
 	type GraphColumnSetting as GKGraphColumnSetting,
 	type GraphColumnsSettings as GKGraphColumnsSettings,
+	type GraphPlatform,
 	type GraphRow,
 	type GraphZoneType,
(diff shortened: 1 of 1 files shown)


src/webviews/apps/plus/graph/GraphWrapper.tsx
```diff
@@ -114,6 +116,20 @@ const getIconElementLibrary = (iconKey: string) => {
 	return iconElementLibrary[iconKey];
 };
 
+const getClientPlatform = (): GraphPlatform => {
+	switch (getPlatform()) {
+		case 'web-macOS':
+			return 'darwin';
+		case 'web-windows':
+			return 'win32';
```
d13, line 124: do we want to imply 32 bit vs 64 bit?

miggy-e: There is currently no need to specify.

eamodio (approved):
Looks good! 
:shipit:

axosoft-ramint (approved):

PR #2200, opened 2022-09-07
Send detached HEAD to graph

axosoft-ramint:
In the local git provider, we skip over HEAD in tips. This causes it not to populate in a commit row's Heads collection and therefore the graph doesn't display it when you're on a detached head. This change causes the detached HEAD to be added as a ref, so it makes it to the graph rows and ultimately is displayed as a local ref in the graph:

image

**Note:** In GK Client, we display the detached HEAD as a tag, rather than a local ref. I'm not sure why, but if we wish to do it that way, I can modify the logic here accordingly.  We would need to update the component-side graph logic in that case because Tag types don't include an "isCurrentHeadRef" property.

### DIFF
--- a/src/env/node/git/localGitProvider.ts
+++ b/src/env/node/git/localGitProvider.ts
@@ -1756,7 +1756,7 @@ export class LocalGitProvider implements GitProvider, Disposable {
 
 				if (commit.tips) {
 					for (let tip of commit.tips.split(', ')) {
-						if (tip === 'refs/stash' || tip === 'HEAD') continue;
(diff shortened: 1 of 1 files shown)


axosoft-ramint:
@eamodio Touching git provider logic here and could use your 👀

axosoft-ramint:
@eamodio Resurrected and updated this branch. Please have a look when you're free and let me know if my provider modifications seem safe. 👍🏼

eamodio (approved):
Looks great!

PR #2248, opened 2022-10-05
Send remote ids and handle ref metadata requests when remote connected

axosoft-ramint:
Sends PR metadata through to the graph in order to show PR icons next to refs when they are applicable.

image

Uses a dynamic system where an empty object is sent and remotes, when they render, ask for ref metadata. If we find something for them, we send it over. Otherwise, we send `null` indicating that this remote should stop asking for PR metadata.

To see PRs for a remote appear, you must connect to that remote first:

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -12336,7 +12336,7 @@
 		"vscode:prepublish": "yarn run bundle"
 	},
 	"dependencies": {
-		"@gitkraken/gitkraken-components": "1.0.0-rc.20",
+		"@gitkraken/gitkraken-components": "1.0.0-rc.21",
 		"@microsoft/fast-element": "1.10.5",
 		"@microsoft/fast-react-wrapper": "0.3.14",
 		"@octokit/core": "4.0.5",
--- a/package.json
+++ b/package.json
@@ -6431,6 +6431,18 @@
 				"icon": "$(git-pull-request-create)"
 			},
 			{
+				"command": "gitlens.graph.openPullRequestOnRemote",
+				"title": "Open Pull Request on Remote",
+				"category": "GitLens",
+				"icon": "$(globe)"
+			},
+			{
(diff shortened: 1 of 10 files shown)


axosoft-ramint:
@eamodio Opened this up as a draft because:

1. I need help, when you have time, sending the context over for a pull request. No idea how to set that up, but it can be included in pull request metadata and will allow us to open context menu on the PR icon.
2. Currently, we don't update state when connecting/disconnecting to a remote. You have to reload the graph to see the PR icons appear or disappear. Not sure yet how to set that up.
3. Not sure if I did the ID part correctly. The ID part I'm sending for a remote ref is the id found on the GitBranch object for that remote ref, which is different than in @ericf-axosoft 's PR which generates an ID instead using repo path and branch names.

ericf-axosoft:
@axosoft-ramint I'm using the `GitBranch.id` as well in my PR:

Captura de pantalla 2022-10-05 a las 14 29 42

axosoft-ramint:
> @axosoft-ramint I'm using the `GitBranch.id` as well in my PR:

> But using `getGitBranchId` function instead.

Ahhh ok, thanks! In that case, I can update mine to match yours (or can just rebase on top of it if it gets merged first). 👍🏼

ericf-axosoft:
> > @axosoft-ramint I'm using the `GitBranch.id` as well in my PR:
> 
> > But using `getGitBranchId` function instead.
> 
> Ahhh ok, thanks! In that case, I can update mine to match yours (or can just rebase on top of it if it gets merged first). 👍🏼

🤔 uhm, if you are using the GitBranch object, no rebase on top or code modifications are need. You have just to use the `id` property.

My change is just to expose a function to generate the same id that we generate in the GitBranch constructor 🤷

PR #2258, opened 2022-10-10
Graph: double-click on a ref to check it out (similar to GKC)

ericf-axosoft:
# Summary of changes
- Used the `onDoubleClickGraphRef` input event property of the Graph component to do the check out of the branch.

https://user-images.githubusercontent.com/90025366/194901856-ec889122-34e9-42bf-9675-096ca55a45e4.mov

### DIFF
--- a/src/plus/webviews/graph/graphWebview.ts
+++ b/src/plus/webviews/graph/graphWebview.ts
@@ -1,3 +1,5 @@
+import type { Head } from '@gitkraken/gitkraken-components';
+import { stringify } from 'json5';
 import type {
 	ColorTheme,
 	ConfigurationChangeEvent,
@@ -567,13 +569,13 @@ export class GraphWebview extends WebviewBase<State> {
 	}
 
 	private onDoubleClickRef(e: DoubleClickedRefParams) {
-		const refContext: string | undefined = (e.ref as any).context;
+		if (e.ref.context) {
+			const item: GraphItemContext = typeof e.ref.context === 'string'
+				? JSON.parse(e.ref.context)
+				: e.ref.context as GraphItemContext;
 
-		if (refContext) {
-			const item: GraphItemContext = JSON.parse(refContext);
-			const { ref } = (item.webviewItemValue as any);
-
-			if ((e.ref as any).isCurrentHead) {
+			const { ref } = item.webviewItemValue as GraphItemRefContextValue;
+			if (e.ref.refType === 'head' && (e.ref as Head).isCurrentHead) {
 				return GitActions.switchTo(ref.repoPath);
 			}
 			return GitActions.switchTo(ref.repoPath, ref);
--- a/src/plus/webviews/graph/graphWebview.ts
+++ b/src/plus/webviews/graph/graphWebview.ts
@@ -101,10 +101,11 @@ import type {
 	SearchOpenInViewParams,
 	SearchParams,
 	State,
+	SwitchToRefParams,
 	UpdateColumnParams,
 	UpdateRefsVisibilityParams,
 	UpdateSelectedRepositoryParams,
-	UpdateSelectionParams,
+	UpdateSelectionParams
 } from './protocol';
 import {
 	DidChangeAvatarsNotificationType,
@@ -126,6 +127,7 @@ import {
 	GetMoreRowsCommandType,
 	SearchCommandType,
 	SearchOpenInViewCommandType,
+	SwitchToRefCommandType,
 	UpdateColumnCommandType,
 	UpdateRefsVisibilityCommandType,
 	UpdateSelectedRepositoryCommandType,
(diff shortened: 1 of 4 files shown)


src/plus/webviews/graph/graphWebview.ts
```diff
@@ -539,6 +543,16 @@ export class GraphWebview extends WebviewBase<State> {
 		this.updateColumn(e.name, e.config);
 	}
 
+	private onSwitchToRef(e: SwitchToRefParams) {
+		const item: GraphItemContext = JSON.parse((e.ref as any).context);
```
ericf-axosoft, line 547: @eamodio @d13 I'm not sure if it is the best way of doing that (using the `context` string property). 

What do you guys think?

axosoft-ramint (changes requested):
This feature should definitely be a togglable setting in the graph settings. Not sure if it should be enabled or disabled by default though (probably disabled).

ericf-axosoft:
> This feature should definitely be a togglable setting in the graph settings. Not sure if it should be enabled or disabled by default though (probably disabled).

I don't see that requirements in our task. I guess it's something Eric A. said we should do.

@eamodio @d13 this new option should be disabled by default as Ramin said? how this should work?

eamodio:
I do agree it should be configurable ultimately, but I think I'd ideally want more than just a bool, e.g. allow you to pick the double click behavior. But that's out of scope for this release and since we do prompt before actually switching (checking out) I think we are OK for this release. Unless there are other concerns/considerations I'm missing.

ericf-axosoft:
@eamodio @axosoft-ramint @d13 I have just resolved the conflicts

axosoft-ramint (approved):
I notice that when I double-click a remote ref that either I don't have yet locally or my local branch is behind the remote one, it puts me in a detached HEAD state:

https://user-images.githubusercontent.com/67011668/195223676-1258bfee-ee72-4706-897b-910a7eda5cb3.mp4

This is fine for now, because I'm guessing we don't have a flow yet to ask the user if they want to catch their local version up to the remote (or any other intervention that GKC would do), but as this produces a different behavior than GKC, folks might ask about it.

LGTM

src/plus/webviews/graph/graphWebview.ts
```diff
@@ -539,6 +543,22 @@ export class GraphWebview extends WebviewBase<State> {
 		this.updateColumn(e.name, e.config);
 	}
 
+	private onSwitchToRef(e: SwitchToRefParams) {
+		const refContext: string | undefined = (e.ref as any).context;
```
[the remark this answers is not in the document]
eamodio: We should definitely provide some valid type for `e.ref` so we don't use an `as any` cast. For using the context, I think that works.

FYI, @axosoft-ramint is working on allowing a string or object for the contexts, which once it lands I'll switch over to using.

src/webviews/apps/plus/graph/GraphWrapper.tsx
```diff
@@ -57,6 +59,7 @@ export interface GraphWrapperProps {
 	subscriber: (callback: UpdateStateCallback) => () => void;
 	onSelectRepository?: (repository: GraphRepository) => void;
 	onColumnChange?: (name: GraphColumnName, settings: GraphColumnConfig) => void;
+	onSwitchToRef?: (ref: GraphRef) => void;
```
eamodio, line 62: I think we should reframe this to be `onDoubleClickRef`

src/plus/webviews/graph/protocol.ts
```diff
@@ -174,6 +175,11 @@ export const UpdateRefsVisibilityCommandType = new IpcCommandType<UpdateRefsVisi
 	'graph/refs/update/visibility',
 );
 
+export interface SwitchToRefParams {
+	ref: GraphRef
+}
+export const SwitchToRefCommandType = new IpcCommandType<SwitchToRefParams>('graph/ref/checkout');
```
eamodio, line 181: Reframe this to be `DoubleClickedRefCommandType`, so that why on the GL side we can decide the behavior of the double click

eamodio (changes requested):
See comments in line

src/webviews/apps/plus/graph/GraphWrapper.tsx
ericf-axosoft: Modified in commit: `Renamed command type and function as requested in PR`.

src/plus/webviews/graph/graphWebview.ts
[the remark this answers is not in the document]
axosoft-ramint: @eamodio This change has landed, so feel free to switch over to it here.

axosoft-ramint (approved):
No need to re-request my review if you only renamed a few things since I last approved. 👍🏼

src/plus/webviews/graph/graphWebview.ts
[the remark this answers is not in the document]
ericf-axosoft: @eamodio @axosoft-ramint just removed the cast to any and provided a valid type (I have modified the d.ts file of the Graph component).

For now, I have modified that function to be able to use string or object for the context.

PR #2282, opened 2022-10-20
116 - Remove "height" and "width" input properties of the "GraphContainer" (Gitlens side)

ericf-axosoft:
# Description
Right now, Gitlens is calculating the `height` and `width` of the graph and passing its values via input properties to the component.

This calculation should be done internally in the Graph component instead (Graph component should be expanded to the container automatically).

That is, "height" and "width" input properties should be removed as they will no longer be needed.

# Summary of changes
- Removed logic to resize graph component (in `GraphWrapper.jsx` file) when webview is resized. Now, the graph component is able to do these calculations internally
- Removed `height` and `width` input properties of the `GraphContainer` component (in `GraphWrapper.jsx` file).

# Notes
- This PR needs the changes of `Graph component` side as well: https://github.com/gitkraken/GitKrakenComponents/pull/117
- Once we proceed to merge this PR, we will need to do a bump version.

# Task
https://github.com/gitkraken/GitKrakenComponents/issues/116

### DIFF
--- a/src/webviews/apps/plus/graph/GraphWrapper.tsx
+++ b/src/webviews/apps/plus/graph/GraphWrapper.tsx
@@ -10,7 +10,7 @@ import type {
 	OnFormatCommitDateTime,
 } from '@gitkraken/gitkraken-components';
 import type { ReactElement } from 'react';
-import React, { createElement, useEffect, useLayoutEffect, useMemo, useRef, useState } from 'react';
+import React, { createElement, useEffect, useMemo, useRef, useState } from 'react';
 import { getPlatform } from '@env/platform';
 import { DateStyle } from '../../../../config';
 import { RepositoryVisibility } from '../../../../git/gitProvider';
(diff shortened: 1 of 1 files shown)


src/webviews/apps/plus/graph/GraphWrapper.tsx
```diff
@@ -155,10 +143,6 @@ export function GraphWrapper({
 	onSelectionChange,
 	onDismissBanner,
 }: GraphWrapperProps) {
-	// TODO: application shouldn't know about the graph component's header
-	const graphHeaderOffset = 24;
-	const [mainWidth, setMainWidth] = useState<number>();
-	const [mainHeight, setMainHeight] = useState<number>();
 	const mainRef = useRef<HTMLElement>(null);
```
d13, line 146: `mainRef` was only needed for the `ResizeObserver`. This can be remove too.

ericf-axosoft: Done in commit: `Removed unneeded "mainRef" variable`.

d13 (approved):

Issue #2284, opened 2022-10-21
Ability to debug our Graph code by adding a breakpoint.


ericf-axosoft:
Ability to debug our Graph code by adding a breakpoint.

PR #2286, opened 2022-10-21
2284 - Enable GK Graph debug

ericf-axosoft:
# Description
- Ability to debug our Graph code by adding a breakpoint.

# Summary of changes
- Modified `webpack.config` to enable inline-source-map and avoid using terser when we are in development mode.
- This works with commit `Enable GK Graph debug` of Graph component side.

# Task
https://github.com/gitkraken/vscode-gitlens/issues/2284

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -344,7 +344,7 @@ function getWebviewsConfig(mode, env) {
 		},
 		mode: mode,
 		target: 'web',
-		devtool: mode === 'production' ? false : 'inline-source-map',
+		devtool: mode === 'production' ? false : 'source-map',
 		output: {
 			filename: '[name].js',
 			path: path.join(__dirname, 'dist', 'webviews'),
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -344,14 +344,14 @@ function getWebviewsConfig(mode, env) {
 		},
 		mode: mode,
 		target: 'web',
-		devtool: mode === 'production' ? false : 'source-map',
+		devtool: mode === 'production' ? false : 'inline-source-map',
 		output: {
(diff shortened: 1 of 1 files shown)


axosoft-ramint:
Question: once this is merged, is there any special command we would have to use when publishing to ensure the published version continues to be minified?

axosoft-ramint (approved):
Approved with a question (see above).

webpack.config.js
```diff
@@ -344,14 +344,14 @@ function getWebviewsConfig(mode, env) {
 		},
 		mode: mode,
 		target: 'web',
-		devtool: mode === 'production' ? false : 'source-map',
+		devtool: mode === 'production' ? false : 'inline-source-map',
```
eamodio, line 347: I haven't tried this (and won't be able to), but please make sure this doesn't negatively impact normal webview debugging.

ericf-axosoft:
> Question: once this is merged, is there any special command we would have to use when publishing to ensure the published version continues to be minified?

@ramin @d13 there is no need to do anything. Webpack will do the work by taking into account if we are in production mode or not. To be sure you can always verify the generated output file and see if it is minified. 

On the library side, `npm` must be running the `yarn build` script which is setting the production mode (`yarn compile --mode production`) (see `package.json` scripts of the library).

webpack.config.js
ericf-axosoft: I don't see any impact on the webview debugging. As it works using 'source-map' as well, I can revert that line if you prefer.

ericf-axosoft: Changed in commit: `Changed "inline-source-map" to "source-map"`.

PR #2330, opened 2022-11-08
84 - Graph: Allow column re-ordering (Gitlens side)

ericf-axosoft:
# Summary of changes
- Added handler to use the new graph input property event to be able to save the column ordering.
- Modified `onColumnChange` to be able to be able to notify more than one column change at the same time.

https://user-images.githubusercontent.com/90025366/200541279-5fd219d5-b397-4248-b2e0-69fdd9c41b1f.mov

# To take into account
- Works with PR https://github.com/gitkraken/GitKrakenComponents/pull/138 (`Graph component` side). 
- Once we get both PR merged, we need to bump the graph library version (in `Gitlens` and `graph component` side).

# Task
https://github.com/gitkraken/GitKrakenComponents/issues/84

### DIFF
--- a/src/plus/webviews/graph/graphWebview.ts
+++ b/src/plus/webviews/graph/graphWebview.ts
@@ -1439,10 +1439,9 @@ export class GraphWebview extends WebviewBase<State> {
 
 	private updateColumns(columnsCfg: GraphColumnsConfig) {
 		let columns = this.container.storage.getWorkspace('graph:columns');
-		Object.keys(columnsCfg).forEach((key: string) => {
-			const name = key as GraphColumnName;
-			columns = updateRecordValue(columns, name, columnsCfg[name]);
-		});
+		for (const [key, value] of Object.entries(columnsCfg)) {
+			columns = updateRecordValue(columns, key, value);
+		}
 		void this.container.storage.storeWorkspace('graph:columns', columns);
 		void this.notifyDidChangeColumns();
 	}
--- a/src/plus/webviews/graph/graphWebview.ts
+++ b/src/plus/webviews/graph/graphWebview.ts
@@ -87,6 +87,7 @@ import type {
 	GetMoreRowsParams,
 	GraphColumnConfig,
 	GraphColumnName,
+	GraphColumnsConfig,
 	GraphColumnsSettings,
 	GraphComponentConfig,
 	GraphHiddenRef,
@@ -101,7 +102,7 @@ import type {
 	SearchOpenInViewParams,
 	SearchParams,
 	State,
-	UpdateColumnParams,
+	UpdateColumnsParams,
 	UpdateRefsVisibilityParams,
 	UpdateSelectedRepositoryParams,
 	UpdateSelectionParams,
@@ -127,7 +128,7 @@ import {
 	GetMoreRowsCommandType,
 	SearchCommandType,
 	SearchOpenInViewCommandType,
-	UpdateColumnCommandType,
+	UpdateColumnsCommandType,
 	UpdateRefsVisibilityCommandType,
 	UpdateSelectedRepositoryCommandType,
 	UpdateSelectionCommandType,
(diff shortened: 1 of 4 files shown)


src/webviews/apps/plus/graph/GraphWrapper.tsx
```diff
@@ -1,14 +1,14 @@
...
+	GraphColumnsSettings,
 	GraphContainerProps,
 	GraphPlatform,
 	GraphRef,
 	GraphRefGroup,
 	GraphRefOptData,
 	GraphRow,
-	OnFormatCommitDateTime,
-} from '@gitkraken/gitkraken-components';
+ OnFormatCommitDateTime } from '@gitkraken/gitkraken-components';
```
axosoft-ramint, line 11: Formatting issue here

ericf-axosoft: Fixed in commit: `Fixed formatting issue`.

src/plus/webviews/graph/graphWebview.ts
```diff
@@ -1410,9 +1411,12 @@ export class GraphWebview extends WebviewBase<State> {
...
 	}
 
-	private updateColumn(name: GraphColumnName, cfg: GraphColumnConfig) {
+	private updateColumns(columnsCfg: GraphColumnsConfig) {
 		let columns = this.container.storage.getWorkspace('graph:columns');
-		columns = updateRecordValue(columns, name, cfg);
+		Object.keys(columnsCfg).forEach((key: string) => {
+			const name = key as GraphColumnName;
+			columns = updateRecordValue(columns, name, columnsCfg[name]);
+		});
```
eamodio, line 1419: In general, I avoid `forEach` loops unless we aren't creating a new closure. As they are harder for the runtime to optimize (mainly because a new closure is created) and harder to step into vs a `for...of` loop.

I'd also rewrite this to avoid an additional lookup on each loop iteration.
```suggestion
		for (const [key, value] of Object.entries(columnsCfg)) {
			columns = updateRecordValue(columns, key, value);
		});
```

axosoft-ramint (approved):
Seems ok to me, other than the feedback already provided.

axosoft-ramint (changes requested):
**Edit:** Spotted a bug when using the commit search. Not sure if it's related to the component side or GitLens side:

The graph search is broken. When trying to navigate search results it fails and we get an error in console. To repro:
1. Open graph on a repo.
2. Search for something in the search bar of the graph (with more than one result).
3. Use the arrow key on the right side of the search box to move to the second result.
4. **Expect:** the graph highlights the next result. **Actual:** no change and receive an error in console:
image

When I try the above steps while on the `main` branches of both, there seems to be no issue.

ericf-axosoft:
> **Edit:** Spotted a bug when using the commit search. Not sure if it's related to the component side or GitLens side:
> 
> The graph search is broken. When trying to navigate search results it fails and we get an error in console. To repro:
> 
> 1. Open graph on a repo.
> 2. Search for something in the search bar of the graph (with more than one result).
> 3. Use the arrow key on the right side of the search box to move to the second result.
> 4. **Expect:** the graph highlights the next result. **Actual:** no change and receive an error in console:
>    image
> 
> When I try the above steps while on the `main` branches of both, there seems to be no issue.

@axosoft-ramint Good catch! the problem was produced because `Gitlens` uses directly `selectCommits` function of the `GraphContainer` component (`graphRef.current as any)?.selectCommits(...)`). So, when I have added the `react-dnd` context, this caused the function to not point to the correct object (instead of `graph Ref.current`, it has to point to `graph Ref.current.child`).

So, the problem was on the library side.

I fixed it by changing the way of creating the `DragDropContext` in our component in commit:
`Fixed problem when searching commits by using directly "selectCommits" function outside our component`

src/plus/webviews/graph/graphWebview.ts
ericf-axosoft: Done in commit: `Avoid "forEach" loops `

axosoft-ramint (approved):
Regression was fixed in the latest test, and did not notice any other issues.

axosoft-ramint:
@eamodio Hoping we can land this and [its component-side partner](https://github.com/gitkraken/GitKrakenComponents/pull/138) today - would you have time to give it a quick look today?

src/webviews/apps/plus/graph/GraphWrapper.tsx
```diff
@@ -464,13 +465,28 @@ export function GraphWrapper({
...
 
+	const handleOnGraphColumnsReOrdered = (columnsSettings: GraphColumnsSettings) => {
+		const graphColumnsConfig: GraphColumnsConfig = {};
+		Object.keys(columnsSettings).forEach((columnName: string) => {
+			graphColumnsConfig[columnName] = {
+				width: columnsSettings[columnName].width,
+				isHidden: columnsSettings[columnName].isHidden,
+				order: columnsSettings[columnName].order
+			};
+		});
```
d13, line 486: `GraphColumnsSettings` and `GraphColumnsConfig` appear to be the same structure. Is the loop just to copy the objects? If so , this could be simplified to:

```suggestion
		for (const [columnName, config] of Object.entries(columnsSettings as GraphColumnsConfig)) {
			graphColumnsConfig[columnName] = { ...config };
		});
```

d13 (approved):

PR #2378, opened 2022-11-28
Add option to modify the scroll padding for the graph

axosoft-ramint:
This is the GitLens-side changes needed for configurable graph scroll padding to work.

[Component-side PR here](https://github.com/gitkraken/GitKrakenComponents/pull/149).

**Note** Must immediately bump library version once this and the component-side changes are merged, or there will be issues with the changes to the graph ref when jumping to commits during commit search.

Adds configurable scroll padding to the graph. This option is shown in the graph options and allows you to configure the number of rows away from the edge before the graph starts to scroll.

image

scroll padding with three rows

**Caveat 1:** This padding is only applied when auto-scrolling to a row (when the graph needs to select and jump to a commit you chose from outside the graph itself, i.e. in the search) or using keyboard scrolling. This padding should NOT be applied when selecting a row with your mouse.

**Caveat 2:** Internal (component-side) restrictions have been placed on the value of the setting. Any padding value below 0 rows will be defaulted to 0 rows. Any value more than half the graph height, in rows, will be defaulted down to that maximum. Note that this isn't easily enforcable in the graph settings on GitLens because you can put in a value that satisfies the constraint and then shrink your graph height.

### DIFF
--- a/src/webviews/apps/settings/partials/commit-graph.html
+++ b/src/webviews/apps/settings/partials/commit-graph.html
@@ -65,6 +65,21 @@
 					</div>
 
 					<div class="setting">
+						<div class="setting__input">
+							<label for="graph.scrollRowPadding">Start scrolling at </label>
+							<input
+								id="graph.scrollRowPadding"
+								name="graph.scrollRowPadding"
+								type="number"
+								placeholder="0"
+								data-setting
+								data-default-value="0"
+							/>
+							<label for="graph.searchItemLimit"> rows from the edge</label>
+						</div>
+					</div>
+
+					<div class="setting">
 						<div class="setting__input setting__input--inner-select">
 							<input
 								id="graph.showDetailsView"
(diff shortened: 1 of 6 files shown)


src/webviews/apps/settings/partials/commit-graph.html
```diff
@@ -64,6 +64,21 @@ <h2>
...
+							<label for="graph.scrollRowPadding">Start scrolling at </label>
+							<input
+								id="graph.scrollRowPadding"
+								name="graph.scrollRowPadding"
+								type="number"
+								placeholder="0"
+								data-setting
+								data-default-value="0"
+							/>
+							<label for="graph.searchItemLimit"> rows from the edge</label>
```
d13, line 78: Is this label meant to point to `graph.scrollRowPadding`?

axosoft-ramint: Oops! Yep, nice catch. Will update it. Thanks!

d13 (approved):

ericf-axosoft (approved):
LGTM

PR #2386, opened 2022-12-09
Add "current branch only" filter logic for Graph

axosoft-ramint:
Adds a filter for showing the current branch only. Preliminary hard-coded UX has a button on the graph status bar at the bottom for toggling it.

Limitations:
- Won't work with the GitHub provider just yet (need to figure out how to apply filters with GitHub API)
- When it's turned on, we currently won't show any stashes (needs further investigation as the way we currently pull stashes and inject them into the Git call makes it tricky).

### DIFF
--- a/src/env/node/git/localGitProvider.ts
+++ b/src/env/node/git/localGitProvider.ts
@@ -29,6 +29,7 @@ import {
 	WorktreeDeleteErrorReason,
 } from '../../../git/errors';
 import type {
+	GitIncludeOptions,
 	GitProvider,
 	GitProviderDescriptor,
 	NextComparisonUrisResult,
@@ -1629,9 +1630,8 @@ export class LocalGitProvider implements GitProvider, Disposable {
 		asWebviewUri: (uri: Uri) => Uri,
 		options?: {
 			branch?: string;
-			currentBranchOnly?: boolean;
+			includes?: GitIncludeOptions;
 			limit?: number;
-			mode?: 'single' | 'local' | 'all';
 			ref?: string;
 		},
 	): Promise<GitGraph> {
@@ -1652,6 +1652,12 @@ export class LocalGitProvider implements GitProvider, Disposable {
 
 		const branches = getSettledValue(branchesResult)?.values;
 		const branchMap = branches != null ? new Map(branches.map(r => [r.name, r])) : new Map<string, GitBranch>();
+
+		const targetBranch: GitBranch | undefined = options?.branch
+			? (options.branch === 'HEAD' ? branches?.find(b => b.current) : branchMap.get(options.branch))
+			: undefined;
+		const targetBranchName: string | undefined = targetBranch?.name;
+		const targetBranchHasUpstream: boolean = targetBranch?.upstream != null && targetBranch?.upstream !== undefined;
 
 		const currentUser = getSettledValue(currentUserResult);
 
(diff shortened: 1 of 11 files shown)


axosoft-ramint:
@eamodio @d13 This is a draft for now until I get the GitHub provider part working. I'm also debating whether to keep the hard-coded, preliminary UX in here or not (@d13 is looking into that separately).

eamodio:
@axosoft-ramint Currently the GitHub stuff is already "filtered" (limited) to current branch only. So we'll just need to update the UI to force you into that mode when it's a virtual repository

src/git/gitProvider.ts
```diff
@@ -34,6 +34,12 @@ export const enum GitProviderId {
 	Vsls = 'vsls',
 }
 
+export interface GitIncludeOptions {
```
eamodio, line 37: Rename to `GitGraphIncludes` or `GitGraphIncludeOptions`

src/env/node/git/localGitProvider.ts
```diff
@@ -4396,6 +4427,16 @@ export class LocalGitProvider implements GitProvider, Disposable {
 				'--use-mailmap',
 			];
 
+			if (options?.branch) {
+				const branch = await this.getBranch(repoPath, options.branch);
+				if (branch != undefined) {
```
eamodio, line 4432: Since its rare for us to compare with `null`, please use `==` or `!=` only with `null` rather than `undefined` (its also shorter)

src/env/node/git/localGitProvider.ts
```diff
@@ -4416,7 +4457,8 @@ export class LocalGitProvider implements GitProvider, Disposable {
 							cancellation: options?.cancellation,
 							configs: ['-C', repoPath, ...gitLogDefaultConfigs],
 							errors: GitErrorHandling.Throw,
-							stdin: stdin,
+							// TODO@ramint Figure out a way to get stashes to work when supplying a local branch
+							stdin: stdin && !(options?.branch) ? stdin : undefined,
```
eamodio, line 4461: I'd move this to line 4417 above

src/env/node/git/localGitProvider.ts
axosoft-ramint: To confirm, would it be ok to shorten to `if (branch) {` here?

src/env/node/git/localGitProvider.ts
```diff
@@ -4396,6 +4427,16 @@ export class LocalGitProvider implements GitProvider, Disposable {
 				'--use-mailmap',
 			];
 
+			if (options?.branch) {
+				const branch = await this.getBranch(repoPath, options.branch);
+				if (branch != undefined) {
+					args.push(branch.name);
+					if (branch.upstream !== undefined && options?.includes?.remotes) {
```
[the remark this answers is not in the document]
axosoft-ramint: Updated in commit: [Fix upstream check logic](https://github.com/gitkraken/vscode-gitlens/pull/2386/commits/f7c7470b5d744627f552f80bf6250493f1797fc8)

axosoft-ramint:
@eamodio @d13 I snuck stash and tag filter support into the PR. See latest commit: [Add support to filter stashes and tags](https://github.com/gitkraken/vscode-gitlens/pull/2386/commits/33d3233623183aa3fd3dda8f5eef3200bcd905fd)

axosoft-ramint:
Closing this PR for now - we can revisit git log modifications as startup behavior after using client-side filtering.

PR #2388, opened 2022-12-11
(wip) hack to demo "include" filtering in Graph

eamodio:
This is just a demo hack to use the `includeRefsById` feature in https://github.com/gitkraken/GitKrakenComponents/pull/156

/cc @gitkraken/gitlens-devs

### DIFF
--- a/src/storage.ts
+++ b/src/storage.ts
@@ -191,9 +191,11 @@ export interface WorkspaceStorage {
 			dismissed?: Record<string, boolean>;
 		};
 		columns?: Record<string, StoredGraphColumn>;
+		// this should be excludeRefs, but we need to keep this name for backward compatibility
 		hiddenRefs?: Record<string, StoredGraphExcludedRef>;
 		includeOnlyRefs?: Record<string, StoredGraphIncludeOnlyRef>;
-		hiddenTypes?: Record<string, boolean>;
+		excludeTypes?: Record<string, boolean>;
+		filters: Record<string, StoredGraphFilters>;
 	};
 	remote: {
 		default?: string;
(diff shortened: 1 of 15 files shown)


src/storage.ts
```diff
@@ -191,7 +191,7 @@ export interface WorkspaceStorage {
 			dismissed?: Record<string, boolean>;
 		};
 		columns?: Record<string, StoredGraphColumn>;
-		hiddenRefs?: Record<string, StoredGraphHiddenRef>;
+		excludeRefs?: Record<string, StoredGraphExcludedRef>;
```
eamodio, line 194: Because this is stored data, I think we should avoid the naming change here. And we don't have a system for good data migration (as vscode doesn't make that easy)

src/webviews/apps/plus/graph/GraphWrapper.tsx
```diff
@@ -238,7 +240,8 @@ export function GraphWrapper({
 				setSelectedRows(state.selectedRows);
 				break;
 			case DidChangeRefsVisibilityNotificationType:
-				setHiddenRefsById(state.hiddenRefs);
+				setExcludeRefsById(state.excludeRefs);
+				setIncludeOnlyRefsById(state.includeOnlyRefs);
```
eamodio, line 244: Should we include `excludeTypes` here? Or should be break them into separate notifications? I think combining them together works.

axosoft-ramint: We can probably combine them for ease of use, though Stashes technically aren't refs and we may introduce other non-ref types in the future.

src/storage.ts
axosoft-ramint: Done in commit: [Revert workspace naming change](https://github.com/gitkraken/vscode-gitlens/pull/2388/commits/11c1ea003481d8e67aa1af5c43554a25f34e1d00)

PR #2391, opened 2022-12-13
153 - Rows are grayed out fairly frequently when VSCode loses the focus (Gitlens side)

ericf-axosoft:
# Summary of changes
- Added new notification to know the VS Code window focus state. In that way, we can notify the graph component to disable the hover behavior of the refs when Gitlens loses the focus. 

# Notes
- This PR works with Graph component side changes: https://github.com/gitkraken/GitKrakenComponents/pull/157

# Task
https://github.com/gitkraken/GitKrakenComponents/issues/153

### DIFF
--- a/src/plus/webviews/graph/graphWebview.ts
+++ b/src/plus/webviews/graph/graphWebview.ts
@@ -101,7 +101,8 @@ import type {
 	UpdateColumnsParams,
 	UpdateRefsVisibilityParams,
 	UpdateSelectedRepositoryParams,
-	UpdateSelectionParams} from './protocol';
+	UpdateSelectionParams
+} from './protocol';
 import {
 	DidChangeAvatarsNotificationType,
 	DidChangeColumnsNotificationType,
--- a/src/plus/webviews/graph/graphWebview.ts
+++ b/src/plus/webviews/graph/graphWebview.ts
@@ -101,8 +101,7 @@ import type {
 	UpdateColumnsParams,
 	UpdateRefsVisibilityParams,
 	UpdateSelectedRepositoryParams,
-	UpdateSelectionParams,
-} from './protocol';
+	UpdateSelectionParams} from './protocol';
 import {
 	DidChangeAvatarsNotificationType,
 	DidChangeColumnsNotificationType,
@@ -113,6 +112,7 @@ import {
 	DidChangeRowsNotificationType,
 	DidChangeSelectionNotificationType,
 	DidChangeSubscriptionNotificationType,
+	DidChangeWindowFocusNotificationType,
 	DidChangeWorkingTreeNotificationType,
 	DidEnsureRowNotificationType,
 	DidFetchNotificationType,
@@ -201,6 +201,7 @@ export class GraphWebview extends WebviewBase<State> {
 	private _lastFetchedDisposable: Disposable | undefined;
 
 	private trialBanner?: boolean;
+	private isWindowFocused: boolean = true;
 
 	constructor(container: Container) {
 		super(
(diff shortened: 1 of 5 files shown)


src/plus/webviews/graph/graphWebview.ts
```diff
@@ -105,8 +105,7 @@ import type {
 	UpdateColumnsParams,
 	UpdateRefsVisibilityParams,
 	UpdateSelectedRepositoryParams,
-	UpdateSelectionParams,
-} from './protocol';
+	UpdateSelectionParams} from './protocol';
```
axosoft-ramint, line 108: Your client may have auto-formatted this, but now it's not consistent with the other import below it and should be reverted.

src/plus/webviews/graph/protocol.ts
```diff
@@ -246,6 +247,14 @@ export const DidChangeColumnsNotificationType = new IpcNotificationType<DidChang
 	true,
 );
 
+export interface DidChangeWindowFocusParams {
+	focused: boolean;
+}
+export const DidChangeWindowFocusNotificationType= new IpcNotificationType<DidChangeWindowFocusParams>(
```
axosoft-ramint, line 253:
```suggestion
export const DidChangeWindowFocusNotificationType = new IpcNotificationType<DidChangeWindowFocusParams>(
```

axosoft-ramint (commented):
I cannot reproduce the original issue this is intended to fix, so I'm not sure how to check that it fixes anything.

Only some tiny feedback about formatting.

src/webviews/webviewBase.ts
```diff
@@ -50,7 +51,10 @@ export abstract class WebviewBase<State> implements Disposable {
 		showCommand: Commands,
 	) {
 		this._originalTitle = this._title = title;
-		this.disposables.push(registerCommand(showCommand, this.onShowCommand, this));
+		this.disposables.push(
+			registerCommand(showCommand, this.onShowCommand, this),
+			window.onDidChangeWindowState(this.onWindowStateChanged, this),
```
eamodio, line 56: We should move this down into the `show` method when we are setting up event handlers for the webview only when its visible

eamodio:
Other than my 1 comment, and a few prettier formatting issues this looks great!

src/plus/webviews/graph/graphWebview.ts
ericf-axosoft: Done in commit: `Fixed code formatting issues`.

src/plus/webviews/graph/protocol.ts
ericf-axosoft: Done in commit: `Fixed code formatting issues`.

src/webviews/webviewBase.ts
ericf-axosoft: Done in commit: `Moved "onDidChangeWindowState" code down into the show method `

axosoft-ramint:
> > I cannot reproduce the original issue this is intended to fix, so I'm not sure how to check that it fixes anything.
> > 
> > Only some tiny feedback about formatting.
> 
> @axosoft-ramint this is the same problem that Ian reported in GKC graph.
> 
> You have the steps to reproduce the problem in the issue task: https://github.com/gitkraken/GitKrakenComponents/issues/153
> 
> Basically, to reproduce the problem, you have to:
> 1. Hover one branch by moving the mouse over.
> 2. Wait to see highlighted all the commits related to that branch.
> 3. Use the keyboard keys to switch between windows (`Option+shift+tab` in MacOS). This will cause VSCode to lose focus. 
> 
> The symptoms are:
> - Rows are grayed out when VSCode loses the focus.
> - The grouped ref stays expanded when there is more than one ref with the same head.
> 
> **Note:** It may be related with the OS platform. I can reproduce it always by using MacOS in my case.

@ericf-axosoft Yes, I know. I already tried these steps and could not reproduce it. It might be related to OS as you said.

axosoft-ramint (approved):

PR #2403, opened 2022-12-16
Add "dim merge commit rows" to filter dropdown

axosoft-ramint:
Adds the option to "dim merge commit rows", which is already in graph settings, to the filter dropdown for easier access/toggling.

Notes/questions:
- Removed "always" from the wording on the setting.
- Going through the command system to do this, would love feedback if there's a simpler way to set it up.
- Having this item checked activates the indicator on the filter icon, just like having any other item in that dropdown checked. Not sure if we want this, as this is technically not a "filter" in the same sense as the other options.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2239,7 +2239,7 @@
 					"gitlens.graph.dimMergeCommits": {
 						"type": "boolean",
 						"default": false,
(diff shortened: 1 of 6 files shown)


eamodio (approved):
Looks good!

PR #2441, opened 2023-01-10
Graph: add setting to show status of local branch relative to its upstream

axosoft-ramint:
image

Adds a graph setting (default: off) to show stats on local branch refs in the graph relative to upstream when an upstream exists. These can be fetched dynamically (they use the same system as avatars).

### DIFF
--- a/src/plus/webviews/graph/graphWebview.ts
+++ b/src/plus/webviews/graph/graphWebview.ts
@@ -98,6 +98,7 @@ import type {
 	GraphRefMetadata,
 	GraphRepository,
 	GraphSelectedRows,
+	GraphUpstreamMetadata,
 	GraphWorkingTreeStats,
 	SearchOpenInViewParams,
 	SearchParams,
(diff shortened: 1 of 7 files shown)


src/plus/webviews/graph/graphWebview.ts
```diff
@@ -706,44 +708,83 @@ export class GraphWebview extends WebviewBase<State> {
...
-					webviewItem: 'gitlens:pullrequest',
-					webviewItemValue: {
-						type: 'pullrequest',
-						id: pr.id,
-						url: pr.url,
-					},
-				}),
-			};
+			if (type === 'upstream') {
+				branch = (await this.container.git.getBranches(repoPath, { filter: b => b.id === id && Boolean(b.upstream) }))
```
eamodio, line 760: It would be good to re-use the branch from above, if it already existed

src/plus/webviews/graph/graphWebview.ts
```diff
@@ -706,44 +708,83 @@ export class GraphWebview extends WebviewBase<State> {
...
+				if (upstream == null) return;
+
+				if (upstream?.missing) {
+					metadata.upstream = null;
+					this._refsMetadata.set(id, metadata);
+					return;
+				}
+
+				// Remote owner is the first part before the /.
+				const upstreamOwner = upstream.name.split('/')[0];
```
eamodio, line 773: Use `getRemoteNameFromBranchName` from `branch.ts`

src/plus/webviews/graph/graphWebview.ts
```diff
@@ -706,44 +708,83 @@ export class GraphWebview extends WebviewBase<State> {
...
+				// Name is everything in the string after th first /.
+				const upstreamName = upstream.name.substring(upstreamOwner.length + 1);
```
eamodio, line 775: Use `getBranchNameWithoutRemote` from `branch.ts`

eamodio (changes requested):
Minor changes inline, but otherwise looks great!

src/plus/webviews/graph/graphWebview.ts
axosoft-ramint: Two issues with that, assuming you mean the `branch` from the `if` block for `pullRequests` type:
1. The `branch` we get there uses a different filter and is looking for a remote branch rather than a local.
2. We always `return` at the end of that `if` block, so we wouldn't reach here if it was defined there.

axosoft-ramint:
It seems we're spamming the graph with requests for upstreams on some branches where they aren't available. Investigating to see if it's a bug, or if we just need to cutoff requests after some time to lower the spam.

axosoft-ramint:
@eamodio When an upstream on a branch has the `missing` flag on, I'm taking that as a signal for "keep asking until the upstream is available". Should it be a "stop asking" flag instead?

If so, is there ever a case where we allow the graph to keep asking for a branch's upstream metadata?

eamodio:
@axosoft-ramint yeah we should not ask again for missing. Missing is typically an indication that the upstream branch was deleted, and won't ever be available.

Were there other cases, where we were asking too often?

src/plus/webviews/graph/graphWebview.ts
eamodio: @axosoft-ramint both cases we'd be looking for the remote branch (we are just doing it slightly differently but should be the same ).

But I missed the return. I was thinking we could get asked to fill out the PR and the branch status at the same time it's not setup like that now. I'll dig more into it later tonight.

axosoft-ramint:
@eamodio I had previously thought that `missing` meant that the upstream was still being pulled in, so I took that to be an "ask for more" signal. However, the upstream never actually came in in my tests, so the component ended up just asking for the upstream metadata over and over again endlessly.

I can't seem to identify any state in which we should signal to the component to "ask again" for upstream metadata, so I will just have it stop asking for upstream stats on a branch, for now, if the upstream metadata isn't found for it.

eamodio:
Yeah the only time that would change would be on a repo change, but that should already cause a "refresh" of data.

src/plus/webviews/graph/graphWebview.ts
axosoft-ramint: > @axosoft-ramint both cases we'd be looking for the remote branch (we are just doing it slightly differently but should be the same ).
> 

Oh, really? I thought in the `upstream` case we were looking for a local branch _with_ an upstream remote, not a remote branch? Or is it actually just grabbing the remote one?

> But I missed the return. I was thinking we could get asked to fill out the PR and the branch status at the same time it's not setup like that now. I'll dig more into it later tonight.

Much appreciated!

axosoft-ramint: @eamodio Okay, I refactored `getRefMetadata` to get the branch just once for an id, and loop through the missing types internally. Have a look at the latest commit: [Move missing type loop into getRefMetadata, get branch once for id](https://github.com/gitkraken/vscode-gitlens/pull/2441/commits/6751242cbd6a03adbed1c10405c36ff02bcf7976)

eamodio (approved):
Great job. I think we will still want to tweak the "pill" styling, but that needs to be done in the component.

src/webviews/apps/plus/graph/graph.scss
```diff
@@ -372,6 +372,11 @@ button:not([disabled]),
 	vertical-align: middle;
 	line-height: 2rem;
 	letter-spacing: normal;
+
+	&.mini-icon {
+		font-size: 10px;
```
d13, line 377: Nit
```suggestion
		font-size: 1rem;
```

axosoft-ramint: Done in commit: [Uses rem for fontsize](https://github.com/gitkraken/vscode-gitlens/pull/2441/commits/1d080b7667a8ca5165c192f879ca90cabc5c30b3)

PR #2450, opened 2023-01-16
Add setting to toggle PR icons in Graph

axosoft-ramint:
We had a setting to enable/disable upstream status for local branches in the graph. This adds a setting to toggle the other metadata type: Pull Requests.

This also combines all the active metadata from settings into a single array to send to the graph. This requires a graph library update after the breaking change lands: https://github.com/gitkraken/GitKrakenComponents/pull/189

Once that merges, the dependency will be updated in this PR and it can leave draft state.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12731,7 +12731,7 @@
 		"vscode:prepublish": "yarn run bundle"
 	},
 	"dependencies": {
-		"@gitkraken/gitkraken-components": "1.1.0-rc.22",
+		"@gitkraken/gitkraken-components": "2.0.0",
 		"@microsoft/fast-element": "1.11.0",
 		"@microsoft/fast-react-wrapper": "0.3.16-0",
 		"@octokit/core": "4.1.0",
--- a/package.json
+++ b/package.json
@@ -2246,7 +2246,7 @@
 					"gitlens.graph.showUpstreamStatus": {
 						"type": "boolean",
 						"default": false,
-						"markdownDescription": "Specifies whether to show a local branch's upstream status in the _Commit Graph_. Requires a connection to a supported remote service (e.g. GitHub)",
+						"markdownDescription": "Specifies whether to show a local branch's upstream status in the _Commit Graph_",
 						"scope": "window",
 						"order": 26
 					},
--- a/package.json
+++ b/package.json
@@ -2250,6 +2250,13 @@
 						"scope": "window",
 						"order": 26
 					},
+					"gitlens.graph.showPullRequests": {
+						"type": "boolean",
+						"default": false,
+						"markdownDescription": "Specifies whether to show associated pull requests on remote branches in the _Commit Graph_",
+						"scope": "window",
+						"order": 27
+					},
 					"gitlens.graph.commitOrdering": {
 						"type": "string",
 						"default": "date",
--- a/package.json
+++ b/package.json
@@ -2246,14 +2246,14 @@
 					"gitlens.graph.showUpstreamStatus": {
 						"type": "boolean",
 						"default": false,
-						"markdownDescription": "Specifies whether to show a local branch's upstream status in the _Commit Graph_",
+						"markdownDescription": "Specifies whether to show a local branch's upstream status in the _Commit Graph_. Requires a connection to a supported remote service (e.g. GitHub)",
 						"scope": "window",
 						"order": 26
 					},
 					"gitlens.graph.pullRequests.enabled": {
 						"type": "boolean",
 						"default": false,
-						"markdownDescription": "Specifies whether to show associated pull requests on remote branches in the _Commit Graph_",
+						"markdownDescription": "Specifies whether to show associated pull requests on remote branches in the _Commit Graph_. Requires a connection to a supported remote service (e.g. GitHub)",
 						"scope": "window",
 						"order": 27
 					},
--- a/package.json
+++ b/package.json
@@ -2250,7 +2250,7 @@
 						"scope": "window",
 						"order": 26
 					},
-					"gitlens.graph.showPullRequests": {
+					"gitlens.graph.pullRequests.enabled": {
 						"type": "boolean",
 						"default": false,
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -2250,6 +2250,13 @@
 						"scope": "window",
 						"order": 26
 					},
+					"gitlens.graph.showPullRequests": {
```
eamodio, line 2253: Lets call it `gitlens.graph.pullRequests.enabled` to align with the other PR related options

src/plus/webviews/graph/protocol.ts
```diff
@@ -37,10 +37,16 @@ export type GraphRefMetadata = RefMetadata | null;
...
 export type GraphHostingServiceType = HostingServiceType;
+export type GraphRefMetadataType = RefMetadataType;
 export type GraphMissingRefsMetadataType = RefMetadataType;
 export type GraphMissingRefsMetadata = Record</*id*/ string, /*missingType*/ GraphMissingRefsMetadataType[]>;
 export type GraphPullRequestMetadata = PullRequestMetadata;
 
+export const GraphRefMetadataTypes = {
+	Upstream: 'upstream',
+	PullRequest: 'pullRequests',
+};
```
axosoft-ramint, line 48: A note on this piece: I originally had it pulling these strings directly from the graph library. However, this caused the GitLens extension to fail to load entirely when debugging. I'm guessing it's because I was attempting to pull constants out of a typedef file, but not sure exactly why it failed.

package.json
axosoft-ramint: Done in commit: [Rename setting](https://github.com/gitkraken/vscode-gitlens/pull/2450/commits/52d8b0ad6b5a0d33781410dae49af11c5fba1dc8)

src/plus/webviews/graph/protocol.ts
eamodio: Yeah, can't use `const`s as values from d.ts files. d.ts are just types/shapes and are not available at runtime.

That's why https://github.dev/gitkraken/vscode-gitlens/blob/52d8b0ad6b5a0d33781410dae49af11c5fba1dc8/src/%40types/vscode.git.enums.ts#L1 exists, I had to separate out certain `const`s from the vscode.git.d.ts file so that they would get compiled in

axosoft-ramint: Thanks! We'll definitely want an enums export in the Graph in general then - I'll add that in as a separate task.

eamodio: I would also suggest changing this to:
```ts
export enum GraphRefMetadataTypes {
	Upstream = 'upstream',
	PullRequest = 'pullRequests',
}

export const supportedRefMetadataTypes: GraphRefMetadataType[] = Object.values(GraphRefMetadataTypes);
```

To avoid having to call `Object.values` each time when checking the metadata, and that will also make sure that enum stays in sync with the types, since the `supportedRefMetadataTypes` will fail if an enum/type value don't change together

src/webviews/apps/settings/partials/commit-graph.html
```diff
@@ -170,6 +170,20 @@ <h2>
...
+								id="graph.pullRequests.enabled"
+								name="graph.pullRequests.enabled"
+								type="checkbox"
+								data-setting
+							/>
+							<label for="graph.pullRequests.enabled"
+								>Show associated pull requests on remote branches</label
+							>
+						</div>
+					</div>
```
eamodio, line 185: Add
```html
<p class="setting__hint hidden" data-visibility="currentLine.pullRequests.enabled">
	<i class="icon icon__info"></i>Requires a connection to a supported remote service (e.g. GitHub)
</p>
```

src/webviews/apps/settings/partials/commit-graph.html
```diff
@@ -170,6 +170,20 @@ <h2>
 						</div>
 					</div>
 
+					<div class="setting">
```
eamodio, line 173: Can do it outside this PR, but we should look at the UX/UI of the graph settings and group some of them up to make them easier for users to understand

axosoft-ramint: Wrote up a task for it, thanks!

src/webviews/apps/settings/partials/commit-graph.html
axosoft-ramint: Thanks. Should I add that for the upstream indicator setting as well?

src/plus/webviews/graph/protocol.ts
axosoft-ramint: Done in commit: [Use enum and values array](https://github.com/gitkraken/vscode-gitlens/pull/2450/commits/ca61b441157e60ab0e166f148d9dce994b424b53)

src/webviews/apps/settings/partials/commit-graph.html
axosoft-ramint: Done in commit: [Add hints for remote service connection](https://github.com/gitkraken/vscode-gitlens/pull/2450/commits/19b2f1435bdabc519efb051bdbbbfc57fe267fc9)

axosoft-ramint:
@eamodio The dependency on the graph side has been merged. Bumped the graph version in this PR to match it.

src/webviews/apps/settings/partials/commit-graph.html
eamodio: Upstream, doesn't need an integration -- just the PRs

axosoft-ramint: Got it, thanks! Fixed in commit: [Remove remote service hint from upstream setting](https://github.com/gitkraken/vscode-gitlens/pull/2450/commits/30826595701a68a8e031799236abfc60174f826f)

eamodio (approved):
Looks good!

d13 (approved):

Issue #2419, opened 2022-12-23
Add a context menu to the WIP row in the Graph


eamodio:
Currently there is no context menu for the work-in-progress row on the Commit Graph. We should a couple of menu choices:

```
Stash All Changes 
--- 
Open Details
Open Source Control
```

- Open Details -- opens the Commit Details view
- Open Source Control -- opens the Source Control sidebar

PR #2458, opened 2023-01-19
App context menu opens to WIP row on graph

axosoft-ramint:
Closes: #2419 

Adds three context menu options to work-in-progress row on graph to stash changes, open in commit details, and open SCM:

WIP context menu

### DIFF
--- a/src/plus/webviews/graph/graphWebview.ts
+++ b/src/plus/webviews/graph/graphWebview.ts
@@ -1622,7 +1622,7 @@ export class GraphWebview extends WebviewBase<State> {
 			context: serializeWebviewItemContext<GraphItemContext>({
 				webviewItem: 'gitlens:wip',
 				webviewItemValue: {
-					type: 'revision',
+					type: 'commit',
 					ref:
 						this.getRevisionReference(
 							this.repository.path,
@@ -2385,8 +2385,7 @@ export type GraphItemRefContextValue =
 	| GraphBranchContextValue
 	| GraphCommitContextValue
 	| GraphStashContextValue
-	| GraphTagContextValue
-	| GraphRevisionContextValue;
+	| GraphTagContextValue;
 
 export type GraphItemRefGroupContext<T = GraphItemRefGroupContextValue> = WebviewItemGroupContext<T>;
 export interface GraphItemRefGroupContextValue {
@@ -2431,11 +2430,6 @@ export interface GraphStashContextValue {
 export interface GraphTagContextValue {
 	type: 'tag';
 	ref: GitTagReference;
-}
-
-export interface GraphRevisionContextValue {
-	type: 'revision';
-	ref: GitRevisionReference;
 }
 
 function isGraphItemContext(item: unknown): item is GraphItemContext {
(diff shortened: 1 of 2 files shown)


src/plus/webviews/graph/graphWebview.ts
```diff
@@ -2395,6 +2433,11 @@ export interface GraphTagContextValue {
 	ref: GitTagReference;
 }
 
+export interface GraphRevisionContextValue {
```
eamodio, line 2436: I don't think we need both `GraphCommitContextValue` and `GraphRevisionContextValue`, can we reuse `GraphCommitContextValue` (since internal GL handles WIP as the "uncommitted" commit) or if we just want to go with revision rename it to `GraphRevisionContextValue`?

If we go with revision, we'd just need to update the usages of the context where we set the `type` of the context to `commit` and change it to `revision` (which does align with how we call it internally)

src/plus/webviews/graph/graphWebview.ts
```diff
@@ -1611,6 +1619,19 @@ export class GraphWebview extends WebviewBase<State> {
...
+				webviewItem: 'gitlens:wip',
+				webviewItemValue: {
+					type: 'revision',
+					ref:
+						this.getRevisionReference(
+							this.repository.path,
+							GitRevision.uncommitted,
+							GitGraphRowType.Working,
+						) ||
+						GitReference.create(GitRevision.uncommitted, this.repository.path, { refType: 'revision' }),
```
eamodio, line 1632: Why do we need this `||` fallback? `getRevisionReference` will only return `undefined` if the `this.repository.path` is `null`, which should already be handled above (and either way that GitReference wouldn't be valid without a repo path)

eamodio (changes requested):
Good overall, just some comments inline.

eamodio:
One question from a UX point of view -- should `Stash All Changes` come after the "opens"? Generally, the way you have it follows our patterns, but I wonder in this case if the 2 opens are the much more likely actions. /cc @d13 @justinrobots

src/plus/webviews/graph/graphWebview.ts
axosoft-ramint: Decided to just use `GraphCommitContextValue`. Thanks!

Done in commit: [Use commit context value](https://github.com/gitkraken/vscode-gitlens/pull/2458/commits/c081cdd5267c362eb7e49cba287a2f10bb81b648)

axosoft-ramint:
> One question from a UX point of view -- should `Stash All Changes` come after the "opens"? Generally, the way you have it follows our patterns, but I wonder in this case if the 2 opens are the much more likely actions. /cc @d13 @justinrobots

Good point and would like feedback from @d13 and @justinrobots

It's a tough one to answer, because sometimes I'm committing my WIP stuff more frequently (in which case I would prefer `Open Source Control` on top) and sometimes I'm stashing more frequently (in which case I would prefer `Stash All Changes` on top).

Overall I find that I stash more often, as usually I'm writing stuff on `main` and end up wanting it on its own branch, or I need to switch context for a bit to test a PR before going back to my changes, so my vote would be to have `Stash All Changes` on top.

axosoft-ramint:
@eamodio Haven't heard any additional feedback on this. Should we go ahead with the current ordering of the menu items and then rearrange them if needed once we get feedback after this lands?

eamodio (approved):
Looks great -- yeah, let's get this in, we can always change the order later.

PR #2484, opened 2023-02-02
Add support for copying deep links to context menus

axosoft-ramint:
From branches/commits/tags in the graph and Source Control views, you can now generate deep links from the 'Copy As' flyout in the context menu.

This utilizes the "open on remote" command flow in the same way that "copy X on remote" does.

It includes a rough first implementation of "get first commit SHA" for the local git provider (not implemented yet for the GitHub git provider) to generate repo ID for the repo.

If a first commit SHA is not found, the command will instead use the repo's built-in ID property as a backup (since we can match the repo using remote URL).


src/plus/webviews/graph/graphWebview.ts
```diff
@@ -2101,6 +2106,46 @@ export class GraphWebview extends WebviewBase<State> {
...
 	}
 
+	@debug()
+	private copyDeepLinkToBranch(item?: GraphItemContext) {
+		if (isGraphItemRefContext(item, 'branch')) {
+			const { ref } = item.webviewItemValue;
+			return executeCommand(Commands.CopyDeepLinkToBranch, ref);
+		}
+
+		return Promise.resolve();
```
d13, line 2116: should we provide a notification for any of these copy actions if they do nothing?

d13:
Reviewing

d13 (changes requested):

src/plus/webviews/graph/graphWebview.ts
eamodio: Maybe? I think we very often just log and fail today. It's definitely a balance.

axosoft-ramint: In this case I was following the setup we had set for other context menu options in the graph webview, but that's a good point. Maybe if we get user feedback about it we can change the standard and start giving notifications for failures with other commands too? Or should we consider this one an exception case where we definitely think the notification on failure is warranted?

package.json
```diff
@@ -4840,6 +4840,30 @@
 				"category": "GitLens",
 				"icon": "$(globe)"
 			},
+			{
+				"command": "gitlens.copyDeepLinkToBranch",
+				"title": "Copy Link to Branch on Remote",
```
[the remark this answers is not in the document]
axosoft-ramint: I wouldn't mind including it. I thought "deep link" is more technical terminology than we needed for user-facing messaging, but on the other hand it does help to distinguish from copying a link to the branch on GitHub, for example.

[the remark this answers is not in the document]
axosoft-ramint: Updated in commit: [Copy link -> copy deep link](https://github.com/gitkraken/vscode-gitlens/pull/2484/commits/0e028bb23321952bb829b682f630e8b122a23756)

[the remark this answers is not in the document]
d13: "Deep Link" is a techie term for sure and probably a better way to word that. For me, "Copy Link to Branch on Remote" makes me think I'm copying the URL to the remote provider.

src/plus/webviews/graph/graphWebview.ts
d13: Makes sense to stay consistent. Sounds like there's a more holistic discussion around that if we want to change it.

src/quickpicks/remoteProviderPicker.ts
```diff
@@ -149,11 +194,28 @@ export namespace RemoteProviderPicker {
 		placeHolder: string,
 		resource: RemoteResource,
 		remotes: GitRemote<RemoteProvider>[],
-		options?: { autoPick?: 'default' | boolean; clipboard?: boolean; setDefault?: boolean },
+		options?: {
+			autoPick?: 'default' | boolean;
+			clipboard?: boolean;
+			deepLink?: boolean;
+			setDefault?: boolean;
+			preSelected?: string;
```
eamodio, line 202: Nit: Rename to `picked` to match our other quickpicks

package.json
```diff
@@ -4861,6 +4861,30 @@
 				"category": "GitLens",
 				"icon": "$(globe)"
 			},
+			{
+				"command": "gitlens.copyDeepLinkToBranch",
+				"title": "Copy Deep Link to Branch on Remote",
```
eamodio, line 4866: Let's nuke the "Deep" and "on Remote" parts of these.
So this one would be `Copy Link to Branch`

I'm a little tempted to call it `Copy GitLens Link to Branch` because we will soon need to add a `Copy GitKraken Link to Branch`. Though it might be better to keep it as 1 or maybe two actions: `Copy Link to Branch` and `Copy Link to Branch [in|via|something]...` where the `...` version will prompt you for the target -- `GitLens`, `GitKraken Client`, `GitKraken.dev`, etc. And the non-`...` version would default to `GitLens`, but could be changed with a setting

package.json
```diff
@@ -7536,6 +7584,22 @@
...
+					"command": "gitlens.copyDeepLinkToBranch",
+					"when": "false"
+				},
+				{
+					"command": "gitlens.copyDeepLinkToCommit",
+					"when": "false"
+				},
+				{
+					"command": "gitlens.copyDeepLinkToRepo",
+					"when": "false"
```
eamodio, line 7597: We might want to allow this one in the command palette

eamodio:
@axosoft-ramint Reworked this quite a bit and there are still a few TODOs but some can probably be moved out into another PR. Let me know if you want to sync on these changes.

PR #2494, opened 2023-02-08
Graph: add support for activity column

axosoft-ramint:
Adds support for the new activity column in the commit graph. Allows the column to be toggled, and sends over colors for the bar visuals in the column.

**Note:** The colors being sent over are hard-coded, until we decide what colors we want to send over and handle theming.

**Note 2:** This cannot be merged until its [dependency PR on the component side](https://github.com/gitkraken/GitKrakenComponents/pull/196) is merged, and the graph library version updated, at which point the PR can leave draft.

### DIFF
--- a/src/plus/webviews/graph/graphWebview.ts
+++ b/src/plus/webviews/graph/graphWebview.ts
@@ -174,7 +174,7 @@ const defaultGraphColumnsSettings: GraphColumnsSettings = {
 	author: { width: 130, isHidden: false },
 	datetime: { width: 130, isHidden: false },
 	sha: { width: 130, isHidden: false },
-	changes: { width: 130, isHidden: false },
+	changes: { width: 130, isHidden: true },
 };
 
 export class GraphWebview extends WebviewBase<State> {
--- a/src/plus/webviews/graph/graphWebview.ts
+++ b/src/plus/webviews/graph/graphWebview.ts
@@ -174,7 +174,7 @@ const defaultGraphColumnsSettings: GraphColumnsSettings = {
 	author: { width: 130, isHidden: false },
 	datetime: { width: 130, isHidden: false },
 	sha: { width: 130, isHidden: false },
-	activity: { width: 130, isHidden: false },
+	changes: { width: 130, isHidden: false },
 };
 
 export class GraphWebview extends WebviewBase<State> {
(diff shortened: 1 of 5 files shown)


src/webviews/apps/plus/graph/graph.tsx
```diff
@@ -457,6 +457,10 @@ export class GraphApp extends App<State> {
 		bodyStyle.setProperty('--color-graph-minimap-remoteForeground', pillLabel);
 		bodyStyle.setProperty('--color-graph-minimap-remoteMarker', opacity(branchMarker, 30));
 
+		bodyStyle.setProperty('--color-graph-stats-deleted', '#c45478');
```
eamodio, line 460: Do we need 3 colors still? Also can we consolidate to a single set of CSS variables with the Graph component?

src/plus/webviews/graph/graphWebview.ts
```diff
@@ -173,6 +173,7 @@ const defaultGraphColumnsSettings: GraphColumnsSettings = {
 	author: { width: 130, isHidden: false },
 	datetime: { width: 130, isHidden: false },
 	sha: { width: 130, isHidden: false },
+	changes: { width: 130, isHidden: false },
```
eamodio, line 176: Because of the perf impact, lets hide this by default (at least for now)

eamodio (changes requested):
Looks good -- few minor changes

src/webviews/apps/plus/graph/graph.tsx
d13: there 3 properties already:

- add `--vscode-gitDecoration-addedResourceForeground`
- modified `--vscode-gitDecoration-modifiedResourceForeground`
- deleted `--vscode-gitDecoration-deletedResourceForeground`

eamodio: We'd need to look at those -- as those are for text foreground colors -- so they might not be the right versions to use for this bar chart. Though they might be a good starting point.

axosoft-ramint: This will be easier to resolve once I know what colors we want. If they are existing colors already available in the graph DOM I can just use them directly. If we need to set up new ones I can add them to the body here as is being done, and access them from within the graph.

There are other colors we are setting up in JS and passing through the cssVariables prop for other graph bits that I will review in another task/PR, but we can at least start on the path here by not adding more to that prop.

src/plus/webviews/graph/graphWebview.ts
axosoft-ramint: Done in commit: [Hides the column by default](https://github.com/gitkraken/vscode-gitlens/pull/2494/commits/53c3f6009a2f81dc6642a149e985de9751d4128c)

src/webviews/apps/plus/graph/graph.tsx
axosoft-ramint: Removed the redundant CSS properties here: [Remove redundant css properties](https://github.com/gitkraken/vscode-gitlens/pull/2494/commits/96e76987cdf53fe072310f4c4cc4416dd770d628)

Applied the suggest colors above here: [Use vscode colors](https://github.com/gitkraken/vscode-gitlens/pull/2494/commits/d7a269b46d654a04b50d6c92727be3359bde5be2)

It looks a tad dark in the light theme. May need to mess with the luminance a bit. But I agree it's a good start.

axosoft-ramint: @eamodio For the second part of your comment...

> Do we need 3 colors still? Also can we consolidate to a single set of CSS variables with the Graph component?

...is the suggestion to get rid of the `cssVariables` prop entirely and pass all styles to the graph through something like `graph.scss`?


I can definitely do that, but as it touches most of the graph's styling, likely better to handle it as a separate issue. What do you think?

axosoft-ramint: Finished in commit: [Put colors and other params directly into graph style](https://github.com/gitkraken/vscode-gitlens/pull/2494/commits/352921ce638b65f29c938fa949c4b45d6a6ef4b2)

PR #2503, opened 2023-02-11
Deep Linking: Prompt to add repo for deep link when no matching repos found

axosoft-ramint:
Expands the deep link service to include a prompt to add the repo using three options (open in current window, open in new window, add to workspace) when a matching repo is not found while resolving a deep link. When a repo is chosen, a notification will indicate that the repo is being opened to resolve the deep link, and can be cancelled.

Includes pending state storage and recall for when adding the repo requires a new instance of the deep link service to continue resolving the link (such as when opening in the current window).

### DIFF
--- a/src/storage.ts
+++ b/src/storage.ts
@@ -189,13 +189,8 @@ export interface StoredBranchComparisons {
 }
 
 export interface StoredDeepLinkContext {
-	state: number;
-	action?: number;
-	uri?: string | undefined;
-	repoId?: string | undefined;
-	remoteUrl?: string | undefined;
-	targetId?: string | undefined;
-	targetType?: string | undefined;
+	url?: string | undefined;
+	repoPath?: string | undefined;
 }
 
 export interface StoredGraphColumn {
--- a/src/storage.ts
+++ b/src/storage.ts
@@ -119,6 +119,7 @@ export type DeprecatedGlobalStorage = {
 
 export type GlobalStorage = {
 	avatars: [string, StoredAvatar][];
+	'deepLinks:pending': StoredDeepLinkContext;
 	'home:actions:completed': CompletedActions[];
 	'home:steps:completed': string[];
 	'home:sections:dismissed': string[];
@@ -185,6 +186,16 @@ export interface StoredBranchComparison {
 
 export interface StoredBranchComparisons {
 	[id: string]: string | StoredBranchComparison;
+}
+
+export interface StoredDeepLinkContext {
+	state: number;
+	action?: number;
+	uri?: string | undefined;
+	repoId?: string | undefined;
+	remoteUrl?: string | undefined;
+	targetId?: string | undefined;
+	targetType?: string | undefined;
 }
 
 export interface StoredGraphColumn {
--- a/src/uris/deepLinks/deepLink.ts
+++ b/src/uris/deepLinks/deepLink.ts
@@ -2,7 +2,6 @@ import type { Uri } from 'vscode';
 import type { GitReference } from '../../git/models/reference';
 import type { GitRemote } from '../../git/models/remote';
 import type { Repository } from '../../git/models/repository';
-import { OpenWorkspaceLocation } from '../../system/utils';
 
 export const enum UriTypes {
 	DeepLink = 'link',
@@ -81,7 +80,6 @@ export const enum DeepLinkServiceState {
 	Idle,
 	RepoMatch,
 	CloneOrAddRepo,
-	OpenedRepo,
 	OpeningRepo,
 	AddedRepoMatch,
 	RemoteMatch,
@@ -96,12 +94,14 @@ export const enum DeepLinkServiceAction {
 	DeepLinkEventFired,
 	DeepLinkCancelled,
 	DeepLinkResolved,
+	DeepLinkStored,
 	DeepLinkErrored,
 	OpenRepo,
 	RepoMatchedWithId,
 	RepoMatchedWithRemoteUrl,
 	RepoMatchFailed,
 	RepoAdded,
+	RepoOpened,
 	RemoteMatched,
 	RemoteMatchFailed,
 	RemoteAdded,
@@ -111,22 +111,9 @@ export const enum DeepLinkServiceAction {
 	TargetFetched,
 }
 
-export enum DeepLinkRepoOpenAction {
-	OpenInCurrentWindow = 'Open in Current Window',
-	OpenInNewWindow = 'Open in New Window',
-	AddToWorkspace = 'Add to Workspace',
-	Cancel = 'Cancel',
-}
-
-export const deepLinkRepoOpenActionToOpenWorkspaceLocation = {
-	[DeepLinkRepoOpenAction.OpenInCurrentWindow]: OpenWorkspaceLocation.CurrentWindow,
-	[DeepLinkRepoOpenAction.OpenInNewWindow]: OpenWorkspaceLocation.NewWindow,
-	[DeepLinkRepoOpenAction.AddToWorkspace]: OpenWorkspaceLocation.AddToWorkspace,
-};
-
 export interface DeepLinkServiceContext {
 	state: DeepLinkServiceState;
-	uri?: string | undefined;
+	url?: string | undefined;
 	repoId?: string | undefined;
 	repo?: Repository | undefined;
 	remoteUrl?: string | undefined;
@@ -147,8 +134,10 @@ export const deepLinkStateTransitionTable: { [state: string]: { [action: string]
 	},
 	[DeepLinkServiceState.CloneOrAddRepo]: {
 		[DeepLinkServiceAction.OpenRepo]: DeepLinkServiceState.OpeningRepo,
+		[DeepLinkServiceAction.RepoOpened]: DeepLinkServiceState.RemoteMatch,
 		[DeepLinkServiceAction.DeepLinkErrored]: DeepLinkServiceState.Idle,
 		[DeepLinkServiceAction.DeepLinkCancelled]: DeepLinkServiceState.Idle,
+		[DeepLinkServiceAction.DeepLinkStored]: DeepLinkServiceState.Idle,
 	},
 	[DeepLinkServiceState.OpeningRepo]: {
 		[DeepLinkServiceAction.RepoAdded]: DeepLinkServiceState.AddedRepoMatch,
--- a/src/uris/deepLinks/deepLink.ts
+++ b/src/uris/deepLinks/deepLink.ts
@@ -111,6 +111,11 @@ export const enum DeepLinkServiceAction {
 	TargetFetched,
 }
 
+export const enum DeepLinkRepoOpenType {
+	Folder = 'folder',
+	Workspace = 'workspace',
+}
+
 export interface DeepLinkServiceContext {
 	state: DeepLinkServiceState;
 	url?: string | undefined;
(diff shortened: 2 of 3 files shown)


src/uris/deepLinks/deepLinkService.ts
```diff
@@ -19,15 +24,16 @@ import {
 } from './deepLink';
 
 export class DeepLinkService implements Disposable {
-	private _disposable: Disposable;
+	private _onDeepLinkEvent: Disposable | undefined;
```
eamodio, line 27: Since we don't actually care about this individually, I'd suggest just a `private readonly _disposables: Disposable[];`

src/uris/deepLinks/deepLink.ts
```diff
@@ -107,9 +111,22 @@ export const enum DeepLinkServiceAction {
...
+export const deepLinkRepoOpenActionToOpenWorkspaceLocation = {
+	[DeepLinkRepoOpenAction.OpenInCurrentWindow]: OpenWorkspaceLocation.CurrentWindow,
+	[DeepLinkRepoOpenAction.OpenInNewWindow]: OpenWorkspaceLocation.NewWindow,
+	[DeepLinkRepoOpenAction.AddToWorkspace]: OpenWorkspaceLocation.AddToWorkspace,
+};
+
 export interface DeepLinkServiceContext {
 	state: DeepLinkServiceState;
-	uri?: Uri | undefined;
+	uri?: string | undefined;
```
eamodio, line 129: Rename to `url` since it is a string (try to keep `uri` for `Uri` instances only)

src/uris/deepLinks/deepLink.ts
```diff
@@ -107,9 +111,22 @@ export const enum DeepLinkServiceAction {
 	TargetFetched,
 }
 
+export enum DeepLinkRepoOpenAction {
```
eamodio, line 114: We shouldn't use `enum`s for UI -- as localization would break this

src/storage.ts
```diff
@@ -187,6 +188,16 @@ export interface StoredBranchComparisons {
 	[id: string]: string | StoredBranchComparison;
 }
 
+export interface StoredDeepLinkContext {
```
eamodio, line 191: What is the absolute minimum we can store here? The less we store the less chance we need to do data migrations. I feel like we could probably get away with the `url` only, but we probably need `state` to pick up where we left off (otherwise we have to start processing over from the url -- which probably would work in our current use-cases).

src/uris/deepLinks/deepLinkService.ts
```diff
@@ -119,35 +143,61 @@ export class DeepLinkService implements Disposable {
 		return undefined;
 	}
 
-	private async processDeepLink(): Promise<void> {
+	private async showOpenRepoPrompt(): Promise<DeepLinkRepoOpenAction> {
+		const result = await window.showInformationMessage(
+			`No matching repository found. Please choose an option to open the repository.`,
+			{ modal: true },
+			{ title: DeepLinkRepoOpenAction.OpenInCurrentWindow },
```
eamodio, line 150: We can extend/add extra props here, so we could add an `action` prop or something that points directly to `OpenWorkspaceLocation`, and it could be `undefined` or missing on `Cancel`

src/uris/deepLinks/deepLinkService.ts
```diff
@@ -119,35 +143,61 @@ export class DeepLinkService implements Disposable {
...
 						break;
 					}
 
+					[, remoteDomain, remotePath] = parseGitRemoteUrl(remoteUrl);
 					// Try to match a repo using the remote URL first, since that saves us some steps.
 					// As a fallback, try to match using the repo id.
 					for (const repo of this.container.git.repositories) {
-						matchingRemotes = await repo.getRemotes({ filter: r => r.url === remoteUrl });
+						// eslint-disable-next-line no-loop-func
+						matchingRemotes = await repo.getRemotes({ filter: r => r.matches(remoteDomain, remotePath) });
```
eamodio, line 200: Not a big deal, but could avoid the eslint error by extracting the filter function to be outside the loop: `const remoteFilter = (r: GitRemote) => r.matches(remoteDomain, remotePath);` and then just use it here. That way we avoid creating new closures for each iteration of the loop.

src/uris/deepLinks/deepLinkService.ts
```diff
@@ -182,13 +232,83 @@ export class DeepLinkService implements Disposable {
...
+						await this.container.storage.store('deepLinks:pending', {
+							state: this._context.state,
+							action: DeepLinkServiceAction.OpenRepo,
+							uri: this._context.uri,
+							repoId: this._context.repoId,
+							remoteUrl: this._context.remoteUrl,
+							targetType: this._context.targetType,
+							targetId: this._context.targetId,
+						});
+						action = DeepLinkServiceAction.DeepLinkCancelled;
```
eamodio, line 272: Probably doesn't matter for today, but we might want to separate `cancelled` from `completed` -- in case we want to log or otherwise do something specific in the one vs the other

src/uris/deepLinks/deepLinkService.ts
```diff
@@ -182,13 +232,83 @@ export class DeepLinkService implements Disposable {
...
 					break;
 
+				case DeepLinkServiceState.OpeningRepo:
+					queueMicrotask(
+						() =>
+							void window.withProgress(
+								{
+									cancellable: true,
+									location: ProgressLocation.Notification,
+									title: `Opening repo for link: ${uri}`,
```
eamodio, line 287: Let's say `repository`

src/storage.ts
axosoft-ramint: I set this up to be flexible enough to allow us to pick up where we left off at any point, though it may be that we really only need to "store" in the "open repo" case.

1.  If we don't mind starting over completely, we probably only need the `url`, and then we can separate out the "on did receive uri" event logic and just call it on the uri again.
2.  If we want to pick up exactly where we left off, but wanted to make it specific to this case only for now, we would want at least the url, repo id and remote url, and we would transition it into the `Added Repo Match` state when we call `processDeepLink` by setting state and action accordingly.

axosoft-ramint: Actually, it looks like we might be able to just use the `url`, set the state and action to take us to "added repo match" and process the other context using our "on did receive uri" logic, separated into its own function. That might be sufficient for now.

src/uris/deepLinks/deepLinkService.ts
axosoft-ramint: I still get `no-loop-func` in that case because I'm in a bigger `while` loop. Also, it gets even worse because I'm declaring a variable in a case statement and get `no-case-declarations`. Despite the new closures, I think this might be the cleanest way to handle it for now.

src/uris/deepLinks/deepLinkService.ts
axosoft-ramint: We have a `DeepLinkResolved` - I could just use that here, but I'm thinking maybe adding a new action `DeepLinkStored` to be more precise.

src/uris/deepLinks/deepLinkService.ts
axosoft-ramint: Done in commit: [Improvements from feedback](https://github.com/gitkraken/vscode-gitlens/pull/2503/commits/9e01ae826ec49b794057aa9570ec3df54ef0042d)

src/uris/deepLinks/deepLinkService.ts
axosoft-ramint: Done in commit: [Improvements from feedback](https://github.com/gitkraken/vscode-gitlens/pull/2503/commits/9e01ae826ec49b794057aa9570ec3df54ef0042d)

src/uris/deepLinks/deepLinkService.ts
axosoft-ramint: Done in commit: [Improvements from feedback](https://github.com/gitkraken/vscode-gitlens/pull/2503/commits/9e01ae826ec49b794057aa9570ec3df54ef0042d)

src/storage.ts
axosoft-ramint: Updated in commit: [Improvements from feedback](https://github.com/gitkraken/vscode-gitlens/pull/2503/commits/9e01ae826ec49b794057aa9570ec3df54ef0042d)

- We store the deep link url and the repo path.
- When the storage value is read, we try to get the repo from the repo path. If we do get it, we skip straight to remote matching. If not, we process the link as normal (i.e. wait for repo discovery to add the repo)

src/uris/deepLinks/deepLink.ts
axosoft-ramint: Done in commit: [Improvements from feedback](https://github.com/gitkraken/vscode-gitlens/pull/2503/commits/9e01ae826ec49b794057aa9570ec3df54ef0042d)

src/uris/deepLinks/deepLink.ts
axosoft-ramint: Done in commit: [Improvements from feedback](https://github.com/gitkraken/vscode-gitlens/pull/2503/commits/9e01ae826ec49b794057aa9570ec3df54ef0042d)

src/uris/deepLinks/deepLinkService.ts
axosoft-ramint: Done in commit: [Improvements from feedback](https://github.com/gitkraken/vscode-gitlens/pull/2503/commits/9e01ae826ec49b794057aa9570ec3df54ef0042d)

src/uris/deepLinks/deepLinkService.ts
```diff
@@ -119,35 +161,61 @@ export class DeepLinkService implements Disposable {
 		return undefined;
 	}
 
-	private async processDeepLink(): Promise<void> {
+	private async showOpenRepoPrompt(): Promise<OpenWorkspaceLocation | undefined> {
```
eamodio, line 164: I think we need another prompt here. This prompt is a bit confusing, since we are saying we can't find the repo, but then asking you to open it.

Instead, I think we should tell the user we couldn't find a matching repo, and ask if they want find it themselves, with buttons to "Open Folder...", "Open Workspace...", "Cancel" (if "Open Workspace..." is tricky we can punt).

Then we open the folder/file picker, and then ask where to open this folder/workspace like this prompt.

Then open the

src/uris/deepLinks/deepLinkService.ts
```diff
@@ -119,35 +161,61 @@ export class DeepLinkService implements Disposable {
...
+		const result = await window.showInformationMessage(
+			`No matching repository found. Please choose an option to open the repository.`,
+			{ modal: true },
+			{ title: 'Open in Current Window', action: OpenWorkspaceLocation.CurrentWindow },
```
eamodio, line 168: `'Open'` rather than `'Open in Current Window'`, since it matches how we and VS Code have similar prompts.

axosoft-ramint: Done in commit: [Update verbiage, open repo flow](https://github.com/gitkraken/vscode-gitlens/pull/2503/commits/2154cca1a37764d8f4416e8b37a0ff75a990a2d0)

src/uris/deepLinks/deepLinkService.ts
axosoft-ramint: Done in commit: [Update verbiage, open repo flow](https://github.com/gitkraken/vscode-gitlens/pull/2503/commits/2154cca1a37764d8f4416e8b37a0ff75a990a2d0)

PR #2514, opened 2023-02-21
Add settings for additional graph minimap marker types

axosoft-ramint:
This PR accomplishes the following:
- Updates the "always-on" markers for rich scrollbar and minimap to HEAD/Upstream, Current Selection, Search Results (Highlights)
- Updates the "default" markers on rich scrollbar and minimap to Local Branches/Stashes.
- Adds a new setting `gitlens.graph.experimental.minimap.additionalTypes` to set up additional markers for the minimap (default are Local Branches and Stashes, and Remote Branches and Tags are available as well).
- The setting is available in User Settings and updating it should immediately update the markers in the minimap to reflect the new settings.

### DIFF
--- a/src/plus/webviews/graph/protocol.ts
+++ b/src/plus/webviews/graph/protocol.ts
@@ -61,6 +61,17 @@ export const enum GraphScrollMarkerTypes {
 	Upstream = 'upstream',
 }
 
+export const enum GraphMinimapMarkerTypes {
+	Selection = 'selection',
+	Head = 'head',
+	Highlights = 'highlights',
+	LocalBranches = 'localBranches',
+	RemoteBranches = 'remoteBranches',
+	Stashes = 'stashes',
+	Tags = 'tags',
+	Upstream = 'upstream',
+}
+
 export const supportedRefMetadataTypes: GraphRefMetadataType[] = Object.values(GraphRefMetadataTypes);
 
 export type GraphCommitDateTimeSource = CommitDateTimeSource;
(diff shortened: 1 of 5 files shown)


axosoft-ramint:
@d13 Asking for your help reviewing this one since @eamodio is out this week. Should be fairly straightforward - let me know if you spot any issues or have questions.

src/plus/webviews/graph/protocol.ts
```diff
@@ -61,6 +61,17 @@ export const enum GraphScrollMarkerTypes {
...
+	Selection = 'selection',
+	Head = 'head',
+	Highlights = 'highlights',
+	LocalBranches = 'localBranches',
+	RemoteBranches = 'remoteBranches',
+	Stashes = 'stashes',
+	Tags = 'tags',
+	Upstream = 'upstream',
+}
+
```
d13, line 74: I noticed this enum doesn't match the `GraphMinimapTypes` in `config.ts`. Should these match?

d13 (commented):
I have one question, otherwise LGTM!

src/plus/webviews/graph/protocol.ts
axosoft-ramint: HEAD and upstream are controlled by a single `head` setting in the configuration, but they are two different markers in the minimap. Similar setup with scroll markers.

axosoft-ramint: But I guess since some of these are always-on, we don't need to include them in the config. Let me take a look at it. We might be able to remove `Head`, `Selection`, and `Highlights` from `GraphMinimapTypes` and `GraphScrollMarkerTypes`.

d13 (approved):

src/plus/webviews/graph/protocol.ts
axosoft-ramint: Let's keep the full list for now and we can trim down later if needed. 👍🏼

PR #2544, opened 2023-03-06
Skips caching when getting main repo for worktree creation

axosoft-ramint:
Issue repro steps:
1. Turn the `gitlens.detectNestedRepositories` setting on
1. Using git command line, create a worktree in the `.git` folder of a repo
2. Open that worktree
4. Attempt to create a new worktree while inside that worktree

Expected: You are able to create a new wotkree
Actual: You get a missing repository error in the debug output.

### DIFF
--- a/src/git/models/repository.ts
+++ b/src/git/models/repository.ts
@@ -647,12 +647,16 @@ export class Repository implements Disposable {
 
 	@gate()
 	@log<Repository['getMainRepository']>({ exit: r => `returned ${r?.path}` })
-	async getMainRepository(options?: { detectNested?: boolean; force?: boolean }): Promise<Repository | undefined> {
+	async getMainRepository(): Promise<Repository | undefined> {
 		const gitDir = await this.getGitDir();
 		if (gitDir?.commonUri == null) return this;
 
 		// If the repository isn't already opened, then open it as a "closed" repo (won't show up in the UI)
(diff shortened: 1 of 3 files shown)


src/git/models/repository.ts
```diff
@@ -647,12 +647,12 @@ export class Repository implements Disposable {
 
 	@gate()
 	@log<Repository['getMainRepository']>({ exit: r => `returned ${r?.path}` })
-	async getMainRepository(): Promise<Repository | undefined> {
+	async getMainRepository(options?: { detectNested?: boolean; force?: boolean }): Promise<Repository | undefined> {
```
eamodio, line 650: Can we avoid taking the options in here, and just pass `{ detectNested: false, force: true }` to the `getOrOpenRepository` call -- since we always want `getMainRepository` to apply those

eamodio:
Other than the 1 comment above this looks good

src/git/models/repository.ts
axosoft-ramint: Done in commit: [Always sets force and no-nested from getMainRepository](https://github.com/gitkraken/vscode-gitlens/pull/2544/commits/1de358c86fae0a5c23f57ea4b0ecdd55ffd48805)

eamodio (approved):

PR #2552, opened 2023-03-07
Pipe downstream info through to graph

axosoft-ramint:
Sends downstream information to graph necessary to achieve the following: always show upstream branches (in the graph, minimap, and rich scrollbar) of visible local branches even when remote branches are filtered off in the graph.


src/env/node/git/localGitProvider.ts
```diff
@@ -2001,6 +2005,16 @@ export class LocalGitProvider implements GitProvider, Disposable {
...
 						refHeads.push(refHead);
+						if (branch?.upstream?.name != null) {
+							// Add the branch name (tip) to the upstream name entry in the downstreams map
+							let downstreams = downstreamMap.get(branch.upstream.name);
+							if (downstreams == null) {
+								downstreams = [];
+							}
+
+							downstreams.push(tip);
+							downstreamMap.set(branch.upstream.name, downstreams);
```
eamodio, line 2016: Can move this up, as you only need to re-set the array in the map when its `undefined`

src/plus/github/githubGitProvider.ts
```diff
@@ -1246,6 +1249,17 @@ export class GitHubGitProvider implements GitProvider, Disposable {
...
+
+				if (branch.upstream?.name != null) {
+					// Add the branch name (tip) to the upstream name entry in the downstreams map
+					let downstreams = downstreamMap.get(branch.upstream.name);
+					if (downstreams == null) {
+						downstreams = [];
+					}
+
+					downstreams.push(branch.name);
+					downstreamMap.set(branch.upstream.name, downstreams);
```
eamodio, line 1261: Same here.

eamodio (approved):
Looks great with just those 2 minor optimization changes.

src/env/node/git/localGitProvider.ts
axosoft-ramint: Done in commit: [Optimizations](https://github.com/gitkraken/vscode-gitlens/pull/2552/commits/822bbc116c4740fb919c2a1af03d019b4a56e701)

src/plus/github/githubGitProvider.ts
axosoft-ramint: Done in commit: [Optimizations](https://github.com/gitkraken/vscode-gitlens/pull/2552/commits/822bbc116c4740fb919c2a1af03d019b4a56e701)

PR #2560, opened 2023-03-09
Sends upstream id on a local branch ref to graph

axosoft-ramint:
Small change which sends the ID of upstreams along with their name (previously we only sent the upstream name) to local branch refs in the graph, for this feature to work: https://github.com/gitkraken/GitKrakenComponents/pull/205

Once the dependency PR is merged, the graph library published, and the dependency updated here, this PR will leave draft and be ready for review.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13407,7 +13407,7 @@
 		"vscode:prepublish": "yarn run bundle"
 	},
 	"dependencies": {
-		"@gitkraken/gitkraken-components": "7.3.0",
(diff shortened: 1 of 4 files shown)


eamodio (approved):
:shipit: I committed a tweak to avoid a needless object allocation.

axosoft-ramint:
> :shipit: I committed a tweak to avoid a needless object allocation.

Thanks!

PR #2619, opened 2023-04-03
Graph: Add "publish branch" context item for local branches

axosoft-ramint:
Adds the ability to "Publish Branch" in the context menu for local branches in the graph without an upstream.

This feature already exists in the Branches view - simply adding similar functionality to the graph.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6975,6 +6975,13 @@
 				"enablement": "!operationInProgress"
 			},
 			{
+				"command": "gitlens.graph.publishBranch",
+				"title": "Publish Branch",
+				"category": "GitLens",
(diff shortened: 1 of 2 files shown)


axosoft-ramint:
@d13 Had to make a PR for this one as I am unable to test it myself. The "publish branch" action in general is not working for me in VSCode (not in the SCM view nor the Branches view).

Mind giving it a quick test/pass?

To test, right click a local branch in the commit graph that doesn't have an upstream, on a remote-connected repo. Should see a "Publish Branch" action. That option should not be there if the local branch has an upstream.

PR #2650, opened 2023-04-25
Workspaces sidebar view

axosoft-ramint:
Implements a workspaces sidebar view (v1, to be expanded upon).

Implements functionality via buttons on view items.

image

**Notes:**

 - Repositories in a workspace that you currently have open in VS Code show up in green.
 - Local workspaces are read-only. You currently cannot interact with them (you cannot add, delete, add a repository to, or remove a repository from a local workspace).
 - Shared cloud workspaces require GitLens Pro and will be filtered (with a message in the view) if the api fetches them but you do not have a pro account.
 - There is no polling. Cloud workspaces are fetched once on view load and once whenever it is reset (i.e. as a result of changing account state or issuing a Refresh command on the view).


axosoft-ramint:
You can now also open a workspace as a code-workspace file:

image

PR #2651, opened 2023-04-26
Unlock all graph columns (for moving and hiding)

axosoft-ramint:
This PR unlocks graph columns that were previously locked in so that you can rearrange them and hide them.

The unlocked columns include the Branch/Tag column. the Graph column, and the Commit Message column.

On the graph component side, these columns were unlocked so that they can be  rearranged. On GL side, added commands to the graph header to be able to hide/unhide these columns.

A few caveats:
- Hiding all columns puts the graph into a weird state, so I added a restriction on hide commands that hides the "hide column" option when it's the only visible column.
- To calculate this "only visible column" and add it to the header context, I had to narrow the settings loop to only consider the most recent names of the columns, as there were old column settings that were never cleaned up in my storage. So I added a `validColumns` array.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7236,12 +7236,12 @@
 			},
 			{
 				"command": "gitlens.graph.columnRefOn",
-				"title": "Show Branch/Tag Column",
+				"title": "Show Branch / Tag Column",
 				"category": "GitLens"
 			},
 			{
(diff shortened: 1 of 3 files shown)


eamodio (approved):
Looks good!

PR #2686, opened 2023-05-19
Adds support for Clone, Fetch, and Remote Add to Deep Link service

axosoft-ramint:
Adds a few extra options when processing deep links:

1. Clone - you can choose to clone a repository if one is not found in your open repositories.
2. Fetch - if your target repo and remote are found, but the commit/branch/tag targeted by the deep link are not found, you are prompted to fetch first.
3. Add remote - If a matching repository is found but not a remote (i.e. the remote url is not matched to any of the remotes in that repository) you are prompted to add the remote.

**Notes:**
1. Hooking this system up to the "repoMapping.json" shared file to "remember" repo disk locations will be done in a different task. It still needs discussion (i.e. how the UX would look if you have multiple disk paths pointing to a repo)
2. Implemented the ability to "open as remote repository" when opening the repo for the deep link (as an alternative to cloning), but our Github git provider has no implementation for `validateReference` and this may need some further thought, so I will track it as a separate issue/follow-up.

### DIFF
--- a/src/uris/deepLinks/deepLinkService.ts
+++ b/src/uris/deepLinks/deepLinkService.ts
@@ -180,6 +180,7 @@ export class DeepLinkService implements Disposable {
 			{ modal: true },
 			{ title: 'Open Folder', action: DeepLinkRepoOpenType.Folder },
 			{ title: 'Open Workspace', action: DeepLinkRepoOpenType.Workspace },
+			{ title: 'Clone Repository', action: DeepLinkRepoOpenType.Clone },
 			{ title: 'Cancel', isCloseAffordance: true },
 		);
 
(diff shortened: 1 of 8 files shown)


src/uris/deepLinks/deepLinkService.ts
```diff
@@ -180,6 +180,7 @@ export class DeepLinkService implements Disposable {
 			{ modal: true },
 			{ title: 'Open Folder', action: DeepLinkRepoOpenType.Folder },
 			{ title: 'Open Workspace', action: DeepLinkRepoOpenType.Workspace },
```
eamodio, line 182: Change to `Add to Workspace` and `Clone`?

PR #2711, opened 2023-06-02
Support deep links that provide repo path

axosoft-ramint:
Provides support for the `path` query parameter in deep links which provides a path to the repo on disk.

Note: while implementing this I noticed that deep links currently aren't working when the graph is in a view, since  the "open in commit graph" command seems to be no-oping. This is a regression (the command used to work even when the graph was in a view) and should be fixed in a separate task.


eamodio (approved):
🚢

Issue #2728, opened 2023-06-08
Submodule commit graph will not open in the panel layout


jkelroy:
### Description

1. clone this repository https://github.com/jkelroy/main-repo.git and open it in VS Code
2. In the Source Control view, attempt to open the submodule's commit graph by selecting `show commit graph`

Expected behavior: The submodules (vscode-gitlens) commit graph is opened

Actual Behavior: The parent repository (main-repo) is opened in the commit graph

This works as expected in the release version (13.6) but does not work in the pre-release (v2023.6.810).

![image](https://github.com/gitkraken/vscode-gitlens/assets/27739784/c9d7ee7c-7e78-467d-a8c0-7f8c55ad1112)

### GitLens Version

v2023.6.810

### VS Code Version

Version: 1.78.2
Commit: b3e4e68a0bc097f0ae7907b217c1119af9e03435
Date: 2023-05-10T15:00:40.428Z (4 wks ago)
Electron: 22.5.2
Chromium: 108.0.5359.215
Node.js: 16.17.1
V8: 10.8.168.25-electron.0
OS: Darwin arm64 22.3.0
Sandboxed: Yes

### Git Version

2.37.1 (Apple Git-137.1)

### Logs, Screenshots, Screen Captures, etc

[gitlens.txt](https://github.com/gitkraken/vscode-gitlens/files/11693978/gitlens.txt)

https://github.com/gitkraken/vscode-gitlens/assets/27739784/87c24730-d39b-45be-98f9-3b47c2eafaa2

axosoft-ramint:
The issue seems to be related to the graph being in "Panel" layout. 

![Image](https://github.com/gitkraken/vscode-gitlens/assets/67011668/e5855993-422b-4d00-a9a4-ac4dc9816d32)

If you switch the layout to "Editor" layout and then try the repro steps above again, it should open the graph to the correct repo.

jkelroy:
@axosoft-ramint yes, it looks like I cannot view the submodule when its in the panel layout. I updated the title of this issue. 

https://github.com/gitkraken/vscode-gitlens/assets/27739784/f5891455-557e-4a25-aff8-ba4d6c9c81a8

lee1409:
Gitlens version
v13.6.0

VS Code
Version: 1.78.2 (Universal)
Commit: b3e4e68a0bc097f0ae7907b217c1119af9e03435
Date: 2023-05-10T14:44:45.204Z
Electron: 22.5.2
[...]

PR #2733, opened 2023-06-12
2728 Submodule commit graph will not open in the panel layout

axosoft-ramint:
Completes #2728

Command args were being lost in the intermediary logic when parsing the "Show Graph" command and the graph is in a view (aka Panel Layout, rather than Editor Layout).

Updates the `showView` pre-execution and execution steps to send the args through (and in case it's an SCM node, pack the node into args since the graph logic will parse it again anyway).

### DIFF
--- a/src/commands/showView.ts
+++ b/src/commands/showView.ts
@@ -27,11 +27,13 @@ export class ShowViewCommand extends Command {
 		]);
 	}
 
-	protected override preExecute(context: CommandContext) {
-		return this.execute(context.command as Commands);
+	protected override preExecute(context: CommandContext, ...args: any[]) {
+		return this.execute(context, ...args);
 	}
 
-	async execute(command: Commands) {
+	async execute(context: CommandContext, ...args: any[]) {
+		const command = context.command as Commands;
+		let commandArgs = args;
 		switch (command) {
 			case Commands.ShowBranchesView:
 				return this.container.branchesView.show();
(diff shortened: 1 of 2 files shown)


axosoft-ramint:
@eamodio Used a PR in this case because I modified the `onShowing` logic in the graph webview as the original didn't quite make sense to me. In one case it had an `if` check where `this.repository == null`, but inside that block it has another `if` check where `this.repository != null` so there was seemingly unreachable code. Let me know if you can see any side effects with it.

src/commands/showView.ts
```diff
@@ -48,7 +50,10 @@ export class ShowViewCommand extends Command {
 			case Commands.ShowAccountView:
 				return this.container.accountView.show();
 			case Commands.ShowGraphView:
-				return this.container.graphView.show();
+				if (context.type === 'scm' && context.scm !== undefined) {
```
eamodio, line 53: Let's convert the SCM object into a Repository here, and then no changes are required on the Graph

```suggestion
			case Commands.ShowGraphView: {
				let commandArgs = args;
				if (context.type === 'scm' && context.scm.rootUri != null) {
					const repo = this.container.git.getRepository(context.scm.rootUri);
					commandArgs = repo != null ? [repo, ...args] : args;
				}
				return this.container.graphView.show(undefined, ...commandArgs);
			}
```

axosoft-ramint: Our linter doesn't like declaractions in case blocks, so I'll have to move that line up, but the rest sounds good!

eamodio: Should be fine as long as it has scope, e.g. wrapped in `{}`

PR #2736, opened 2023-06-13
Remaps links in settings to Help Center sections

axosoft-ramint:
Updates the info links from the old readme sections to the new help center docs.

### DIFF
--- a/src/webviews/apps/settings/partials/modes.html
+++ b/src/webviews/apps/settings/partials/modes.html
@@ -2,7 +2,11 @@
 	<div class="section__header">
 		<h2>
 			Modes
-			<a class="link__learn-more" title="Learn more" href="https://help.gitkraken.com/gitlens/settings/#modes-settings">
+			<a
+				class="link__learn-more"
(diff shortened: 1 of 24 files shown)


src/webviews/apps/settings/partials/modes.html
```diff
@@ -2,7 +2,7 @@
 	<div class="section__header">
 		<h2>
 			Modes
-			<a class="link__learn-more" title="Learn more" href="https://github.com/gitkraken/vscode-gitlens/#modes-">
+			<a class="link__learn-more" title="Learn more" href="https://help.gitkraken.com/gitlens/settings/#modes-settings">
```
axosoft-ramint, line 5: Important note: We don't seem to have a section in the Help Center for Modes (other than the one in Settings which specifies settings variables but doesn't overview the feature). This needs to be updated once we have that section.

axosoft-ramint:
Marking this as draft until we figure out what to do with the "Modes" section - see my comment above @eamodio

PR #2761, opened 2023-06-22
Improves linking between GK and Local Workspaces

axosoft-ramint:
Improves linkage between vscode and cloud workspaces:

- Command next to a GitKraken Workspace in the Workspaces View, previously labeled "open as vscode workspace", split out to several commands: "create vscode workspace". "open vscode workspace (new window)", and "open vscode workspace (current window)".
- When you create a vscode workspace, it saves the GK workspace id in that code workspace file's settings. It also saves the file path of that code-workspace file for that cloud workspace in `cloudWorkspaces.json` in the shared folder (i.e. C`:/Users/(user)/.gk`).
- This changes the command for this cloud workspace from "Create" to "open". From now on, using the command opens the existing workspace at that location.
- When you create the vscode workspace, you now also have the option to enable syncing via three methods: Always, Ask every time or Never.
 - `Always` means whenever a new repo is added to the cloud workspace, it will always automatically add that repo to the linked vscode workspace if it is open.
- `Ask` is similar to above, but it will ask each time if you want to add the new repos to the local workspace.
- `Never` will never do the above syncing.

* Note that the syncing is **one-way** intentionally: we want to be minimally invasive with this syncing for now, so it will only add from cloud to local and only for adding repos, not removing if they were removed from cloud.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4238,6 +4238,15 @@
 				}
 			},
 			{
+				"id": "gitlens.decorations.workspaceCurrentForegroundColor",
+				"description": "Specifies the decoration foreground color of workspaces which are currently open as a Code Workspace file",
+				"defaults": {
+					"dark": "#35b15e",
+					"light": "#35b15e",
+					"highContrast": "#4dff88"
+				}
+			},
+			{
 				"id": "gitlens.decorations.workspaceRepoOpenForegroundColor",
 				"description": "Specifies the decoration foreground color of workspace repos which are open in the current workspace",
 				"defaults": {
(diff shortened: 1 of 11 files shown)


axosoft-ramint:
Having this as a draft for now for two reasons:
1. There is a bigger bug in GL right now affecting this view and others: when repositories get added to the workspace, the Repositories view and Workspaces view aren't refreshing properly. The new repo should show up in both immediately but it doesn't. This means when you choose to sync, it doesn't update the "open" status on the repos listed under the cloud workspace until you do a hard refresh (or reload).
2. The UX/messaging need some work.

axosoft-ramint:
@d13 I'm gonna check in the above, and then if we need to polish the UX/messaging before release, we can. The refresh issue seems to be affecting other views as well so I'll look at it separately.

PR #2825, opened 2023-07-20
Fixes #708 - Prompt for file path when missing from revision

### LINKED ISSUE #708: open any file from another branch
Recently https://github.com/eamodio/vscode-gitlens/issues/646 was implemented, but it doesn't completely work for me. The reason being that sometimes a file is renamed, and then the command (understandably) fails. In emacs/magit this is handled like that: you run the command, then by default the same filename as your current file is displayed and you just press enter if you want the same filename, but you can edit the filename if needed.

Maybe that gitlens feature, `gitlens.openFileRevisionFromBranch`, could be improved so that if it doesn't find the file in another branch, it doesn't fail but rather gives the user the opportunity to manually specify a filename.

In my case I renamed a `.ts` file to `.tsx`, but there could be many examples of renaming files.

mogelbrod:
# Description

Fixes #708 - Prompt for file path when missing from revision

Improves the `openFileAtRevision()` function by:
- Asking user to enter another path when the requested file doesn't exist in the
  selected revision.
- Showing an error message on subsequent failure (or if an error is thrown).

The above function is used by a number of commands:
- `gitlens.openFileRevision`
- `gitlens.openFileRevisionFrom`
- `gitlens.openRevisionFile`

# Checklist

### DIFF
--- a/src/git/actions/commit.ts
+++ b/src/git/actions/commit.ts
@@ -506,26 +506,33 @@ export async function openFileAtRevision(
 
 	let editor: TextEditor | undefined;
 	try {
-		editor = await findOrOpenEditor(uri, { throwOnError: true, ...opts }).catch(error => {
-			if (error?.message?.includes('Unable to resolve nonexistent file')) {
-				return showRevisionPicker(gitUri, {
-					title: 'File not found in revision - pick another file to open instead',
-				}).then(pickedUri => {
-					return pickedUri ? findOrOpenEditor(pickedUri, opts) : undefined;
-				});
-			}
-			throw error;
-		});
-
-		if (annotationType != null && editor != null) {
-			void (await Container.instance.fileAnnotations.show(editor, annotationType, {
-				selection: { line: line },
-			}));
+		editor = await findOrOpenEditor(uri, { throwOnError: true, ...opts });
+	} catch (ex) {
+		if (!ex?.message?.includes('Unable to resolve nonexistent file')) {
+			void window.showErrorMessage(`Unable to open '${gitUri.relativePath}' in revision '${gitUri.sha}'`);
+			return;
 		}
-	} catch (error) {
-		await window.showErrorMessage(
-			`Unable to open '${gitUri.relativePath}' - file doesn't exist in selected revision`,
-		);
+
+		const pickedUri = await showRevisionPicker(Container.instance, gitUri, {
+			ignoreFocusOut: true,
+			title: `Open File at Revision \u2022 Unable to open '${gitUri.relativePath}'`,
+			placeholder: 'Choose a file revision to open',
+			keyboard: {
+				keys: ['right', 'alt+right', 'ctrl+right'],
+				onDidPressKey: async (key, uri) => {
+					await findOrOpenEditor(uri, { ...opts, preserveFocus: true, preview: true });
+				},
+			},
+		});
+		if (pickedUri == null) return;
+
+		editor = await findOrOpenEditor(pickedUri, opts);
+	}
+
+	if (annotationType != null && editor != null) {
+		void (await Container.instance.fileAnnotations.show(editor, annotationType, {
+			selection: { line: line },
+		}));
 	}
 }
 
--- a/src/git/actions/commit.ts
+++ b/src/git/actions/commit.ts
@@ -1,4 +1,4 @@
-import type { TextDocumentShowOptions } from 'vscode';
+import type { TextDocumentShowOptions, TextEditor } from 'vscode';
 import { env, Range, Uri, window, workspace } from 'vscode';
 import type { DiffWithCommandArgs } from '../../commands/diffWith';
 import type { DiffWithPreviousCommandArgs } from '../../commands/diffWithPrevious';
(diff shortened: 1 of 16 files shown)


mogelbrod:
Would love to get this reviewed (and hopefully merged) @eamodio 😃

mogelbrod:
Rebased the PR onto the latest main branch again, eagerly awaiting a review 😄

eamodio:
@mogelbrod Thanks for this -- I've started taking a pass through this and having issues with the `OpenFileByRevisionCommand` -- it needs an active editor (though shouldn't really) and the keyboarding doesn't feel right. In GitLens (and VS Code in general), the using the right arrow key in a quick pick, when appropriate, leaves the quick pick open and acts upon the selected item (opening it, etc). This functionality should be moved into the `pickFileAtRevision` quick pick rather than in the refs picker.

But before I dig more into this, I'd love to have the 2 parts split apart. So split out the `OpenFileByRevisionCommand` into a separate PR, as I think there are more UX implications (naming, etc).

src/git/actions/commit.ts
```diff
@@ -440,11 +441,74 @@ export async function openFileAtRevision(
...
+			);
+		});
+
+		return pick
+			? Container.instance.git.getRevisionUri(uri.sha!, `${uri.repoPath}/${pick}`, uri.repoPath!)
+			: undefined;
+	} finally {
+		disposables.forEach(d => {
+			d.dispose();
+		});
```
eamodio, line 511: Let's move this into `src/quickpicks/revisionPicker.ts` similar to the other pickers

src/git/actions/commit.ts
```diff
@@ -440,11 +441,74 @@ export async function openFileAtRevision(
...
+	const disposables: Disposable[] = [];
+	try {
+		const picker = window.createQuickPick();
+		Object.assign(picker, {
+			title: options.title,
+			value: options.initialPath ?? uri.relativePath,
+			placeholder: 'Enter path to file...',
+			busy: true,
+			ignoreFocusOut: getQuickPickIgnoreFocusOut(),
+		});
```
eamodio, line 487: Please set the props directly -- avoid `Object.assign`

src/git/actions/commit.ts
```diff
@@ -440,11 +441,74 @@ export async function openFileAtRevision(
...
+		picker.show();
+
+		const tree = await Container.instance.git.getTreeForRevision(uri.repoPath, uri.sha!);
+		picker.items = tree.filter(file => file.type === 'blob').map((file): QuickPickItem => ({ label: file.path }));
```
eamodio, line 491: Use `filterMap` from `iterable.ts` to avoid filtering and mapping separately

eamodio, line 491: I'd like to try only putting the base filename as the label and putting the rest of the folder path in the description -- closer to the vscode picker. Will need to enable `picker.matchOnDescription` as well.

src/git/actions/commit.ts
```diff
@@ -440,11 +441,74 @@ export async function openFileAtRevision(
...
+		const pick = await new Promise<string | undefined>(resolve => {
+			disposables.push(
+				picker,
+				picker.onDidHide(() => resolve(undefined)),
+				picker.onDidAccept(() => {
+					if (picker.activeItems.length === 0) return;
+					resolve(picker.activeItems[0].label);
+				}),
+			);
+		});
```
eamodio, line 503: It would be nice to add the extra keyboard (right arrow) support here to open the selected file in a preview editor.

eamodio (changes requested):

src/git/actions/commit.ts
mogelbrod: I didn't know about this functionality 😄 Not sure how though, do you have any pointers to existing implementations?

Would it potentially break text selections using left/right arrow keys in the text field?

src/git/actions/commit.ts
mogelbrod: Sure! Should it be called `fileRevisionPicker` (or a variation thereof) or `revisionPicker`? I'm leaning towards the former since it's technically picking a file at a specific revision, rather than picking a revision (unless the "file" part is implicit for revisions?).

src/quickpicks/revisionPicker.ts
```diff
@@ -0,0 +1,58 @@
...
+		picker.show();
+
+		const tree = await Container.instance.git.getTreeForRevision(uri.repoPath, uri.sha!);
+		picker.items = Array.from(filterMap(tree, file => {
+			// Exclude directories
+			if (file.type !== 'blob') { return null }
+			return { label: file.path }
+			// FIXME: Remove this unless we opt to show the directory in the description
+			// const parsed = path.parse(file.path)
+			// return { label: parsed.base, description: parsed.dir }
```
mogelbrod, line 35: Left this code in if you wanted to try it out yourself @eamodio. It does depend on `path`, which appears to be a forbidden import.

mogelbrod:
Thanks for taking the time to dig into this @eamodio!

I've extracted the `openFileByRevision` command addition into [a separate PR](https://github.com/gitkraken/vscode-gitlens/pull/3036). It's currently marked as draft - I'll update it to reflect all changes once this PR lands in `main`.

PR #2832, opened 2023-07-25
Adds native Pull and Push implementations to experimental Native Git flag

axosoft-ramint:
Adds our own implementations of `pull` and `push` and puts them behind the `gitlens.experimental.nativeGit` flag.

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -600,6 +600,7 @@ export type CoreGitConfiguration =
 	| 'git.enabled'
 	| 'git.fetchOnPull'
 	| 'git.path'
+	| 'git.pullTags'
 	| 'git.repositoryScanIgnoredFolders'
 	| 'git.repositoryScanMaxDepth'
 	| 'git.useForcePushWithLease';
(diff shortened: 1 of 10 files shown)


axosoft-ramint:
@eamodio Updated the git logic to throw the error to the caller after making it more readable, rather than outputting it to a message.

Let me know if it looks good to merge.

eamodio:
Can we create a specific error type for these similar to https://github.com/gitkraken/vscode-gitlens/blob/ee2a0c42a92d33059a39fd15fbbd5dd3d5ab6440/src/git/errors.ts#L13?

axosoft-ramint:
> Can we create a specific error type for these similar to
> ?

Yep, I can add a `PushError`, `PullError` and `FetchError`.

eamodio:
Pushed some minor changes to avoid reaching out the git provider.

Are the `[Fetch|Push|Pull]Error`s getting caught now? Also it doesn't look like pull is checking for a possible rebase and prompting/handling that appropriately

axosoft-ramint:
> Pushed some minor changes to avoid reaching out the git provider.
> 
> Are the `[Fetch|Push|Pull]Error`s getting caught now? Also it doesn't look like pull is checking for a possible rebase and prompting/handling that appropriately

The errors will should now be caught by the local git provider and output there.

Will do prompting/handling for rebase, conflicts, etc. in a follow-up.

axosoft-ramint:
@eamodio Going to check this in and will write up an issue for prompting/intercepting pull with rebase and conflicts

eamodio:
👍

PR #2835, opened 2023-07-27
Adds a "show only changed files" command to command palette and graph

axosoft-ramint:
The new command can be accessed via the GitLens Command Palette: `GitLens: Show Only Changed Files` or on the context menu of the graph WIP node.

It simply combines the "open changed files" and "close unchanged files" commands into a single command for convenience.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5280,8 +5280,8 @@
 				"category": "GitLens"
 			},
 			{
-				"command": "gitlens.showOnlyChangedFiles",
-				"title": "Show Only Changed Files",
+				"command": "gitlens.openOnlyChangedFiles",
+				"title": "Open Changed & Close Unchanged Files",
 				"category": "GitLens"
 			},
 			{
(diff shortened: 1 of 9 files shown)


eamodio:
I think we should be consistent with `Open` over `Show` -- How about `Open Only Changed Files` or `Open Changed Files Only`?

axosoft-ramint:
> I think we should be consistent with `Open` over `Show` -- How about `Open Only Changed Files` or `Open Changed Files Only`?

@eamodio I originally considered calling it `open` but moved away from that because `open` doesn't quite fully capture the fact that we are closing files as well. `Open Only Changed Files` and `Open Changed Files Only` both sound the same as `Open Changed Files` to me in terms of the behavior they are suggesting.

eamodio:
Hrm, I see where you are coming from, but definitely want to avoid "show" as that is pretty exclusively used to show views and whatnot in VS Code.

How about _Close & Open Only Changed Files_ or  _Close All & Open Only Changed Files_?

@d13 @justinrobots  thoughts?

axosoft-ramint:
How about _Open Changed & Close Unchanged Files_ @eamodio ?

eamodio (approved):
:shipit:

PR #2840, opened 2023-08-01
Allows graph "reset to commit" context menu option on branch item

axosoft-ramint:
"Reset Current Branch to Commit..." context menu option in the graph would previously only appear when you were over a commit, rather than a branch.

This tiny change allows the option to appear on the context menu of a branch label in the graph, and should reset to the commit that branch is pointing to.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6092,6 +6092,12 @@
 				"enablement": "!operationInProgress"
 			},
 			{
+				"command": "gitlens.views.resetToTip",
+				"title": "Reset Current Branch to Tip...",
+				"category": "GitLens",
+				"enablement": "!operationInProgress"
+			},
+			{
 				"command": "gitlens.views.revert",
 				"title": "Revert Commit...",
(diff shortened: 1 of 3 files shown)


eamodio:
Code looks good, but I think we should create 2 new commands for this, so we can have the label be _Reset Current Branch to Tip..._? instead of _Reset Current Branch to Commit..._

Thoughts? @d13 @justinrobots

axosoft-ramint:
> Code looks good, but I think we should create 2 new commands for this, so we can have the label be _Reset Current Branch to Tip..._? instead of _Reset Current Branch to Commit..._
> 
> Thoughts? @d13 @justinrobots

That works with me. I'll update it and if we decide against it I can revert.

axosoft-ramint:
@eamodio Updated to "Reset Current Branch to Tip..." wording

eamodio (approved):
:shipit:

PR #2845, opened 2023-08-05
Fixes #1651 - search & compare opens file instead of diff

### LINKED ISSUE #1651: Allow opening file instead of diff
When clicking on the filename in "Search & Compare", it opens a diff by default. To open the file itself one needs to right-click and select "Open File". Would be nice to have a setting to open files by default. This is similar to https://github.com/microsoft/vscode/issues/36885, but for GitLens panels.

Svaught598:
# Description

Closes #1651. Instead of opening a diff from the _Search & Compare_ view, this opens the current revision's file.

The use case seems to be setting a up the workspace, so line numbers seem to be inconsequential here. Can definitely put that back in if it's needed.

# Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 
 ## [Unreleased]
 
+### Changed
+
+- Changes _Search & Compare_ to default to open file instead of diff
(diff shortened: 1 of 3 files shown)


eamodio (changes requested):
@Svaught598 Thanks for the contribution, but this would need to be on a setting new setting, e.g. `gitlens.views.<view-name>.files.openDiffOnClick` which would default to `true`. Also we should definitely keep jumping to the line number of the first diff hunk.

Thanks!

PR #2883, opened 2023-08-24
Adds GK Provider API Library Integration

axosoft-ramint:
-Moves integration authentication session logic into its own domain, migrating that logic out from GitHub and GitLab remotes.

-Adds Azure DevOps and BitBucket authentication prompts (using PATs for now, switching to oAuth in a follow-up PR).

-Adds providers API integration through `providersApi` and `providersService`

-Adds the ability to get the authenticated user info from a provider.

-Adds the ability to get PRs and issues for a list of repos from all providers except Jira and Trello.

-Adds the ability to get repos for an Azure Project, which will be useful for project-synced Azure workspaces.


axosoft-ramint:
@eamodio @d13 This is up-to-date again and ready for review.

eamodio:
@axosoft-ramint I've made a bunch of changes to fully move over to the Integration model and removed the RichRemoteProvider model. I did some very basic sanity checking (and only for GitHub, not GitLab) and it seems to work, but needs a lot more testing.

We also still need to figure out how `src/git/remotes/remoteProviderService.ts` gets integrated or goes away

PR #2886, opened 2023-08-25
2880 Deep Link Bug: When opening vscode as a part of a deep link, sometimes it asks to manually open a repo even if it's already open

### LINKED ISSUE #2880: Deep Link Bug: When opening vscode as a part of a deep link, sometimes it asks to manually open a repo even if it's already open
Note: This may be a race condition between the deep link's repo check and GitLens' repo discovery. We may need to delay the deep link processing until GitLens has had a chance to finish repo discovery.

axosoft-ramint:
Fixes #2880

When opening a deep link to GitLens requires VS Code to open, the deep link service enters a race condition with the git provider service on repo discovery. If the deep link service starts checking for a repo match before the git provider service has had a chance to finish at least some form of repo discovery, you can get situations where the correct repo is already open in the workspace but the service doesn't know it exists, so it asks for the user to manually open it.

This workaround exposes the `isDiscoveringRepositories` promise that the git provider service was previously using internally, and simply has the deep link service wait for it when processing a link.

Ideally the git provider service would expose its repo discovery and workspace folder knowledge in a more robust way (and I'm open to suggestions for building a broader system like that, hence this PR), but this should at least put a dent in the issue if deep links on open are the only place where it's a problem.

### DIFF
--- a/src/git/gitProviderService.ts
+++ b/src/git/gitProviderService.ts
@@ -582,6 +582,9 @@ export class GitProviderService implements Disposable {
 	private _discoveredWorkspaceFolders = new Map<WorkspaceFolder, Promise<Repository[]>>();
 
 	private _isDiscoveringRepositories: Promise<void> | undefined;
+	get isDiscoveringRepositories(): Promise<void> | undefined {
+		return this._isDiscoveringRepositories;
+	}
 
 	@log<GitProviderService['discoverRepositories']>({ args: { 0: folders => folders.length } })
 	async discoverRepositories(folders: readonly WorkspaceFolder[], options?: { force?: boolean }): Promise<void> {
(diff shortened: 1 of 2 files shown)


axosoft-ramint:
This one is a bit difficult to test. I simulated the situation by dropping a `wait(20000)` at the beginning of `discoverRepositories` of `gitProviderService` to ensure it's still working on discovery by the time I click my deep link.

eamodio:
@axosoft-ramint Like you I'm not thrilled about exposing that, but I don't have a better alternative, unless we wanted to add a "awaited" version of `container.git.repositories` that would check the promise and wait if so.

But for that to work, the `container.git.repositories` has to be the only call that needs to be protected for the deeplink case -- which I think it is as `getOrOpenRepository` already checks it.

axosoft-ramint:
> @axosoft-ramint Like you I'm not thrilled about exposing that, but I don't have a better alternative, unless we wanted to add a "awaited" version of `container.git.repositories` that would check the promise and wait if so.
> 
> But for that to work, the `container.git.repositories` has to be the only call that needs to be protected for the deeplink case -- which I think it is as `getOrOpenRepository` already checks it.

That alternative doesn't seem worth it for this case. I'll check this in and if we find a better way for the git provider to inform other services of its repo discovery state, we can switch to using that.

PR #2888, opened 2023-08-26
Allows the Commit Graph to be open in a view & a tab simultaneously

eamodio:
I think this is good to go, but not sure on the wording of:
 - "Switch Commit Graph to Editor Layout" -> "Prefer Editor Layout" 
 - "Switch Commit Graph to Panel Layout" -> "Prefer Panel Layout"

### DIFF
--- a/package.json
+++ b/package.json
@@ -2402,10 +2402,10 @@
 							"panel"
 						],
 						"enumDescriptions": [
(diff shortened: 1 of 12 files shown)


axosoft-ramint (approved):
This was merged already I think, so feel free to close it.

PR #2892, opened 2023-08-30
Hooks deep link service into local repo path mapping

### LINKED ISSUE #2891: Deep Link Bug: If the repo is not open, you always have to locate it even if you already did before
We should hook the service up to the repo path mapping file so that we remember locations you chose for the repo before and let you choose from them.

axosoft-ramint:
Fixes #2891

Previously, the deep link service was not leveraging the repo path mapping file, so when you choose a location of a repo it is not saved to the mapping file, and when matching a repo for a deep link it does not try looking up the path in the mapping file.

In this PR:

- After choosing a location for a repo, it is saved to the mapping file for that remote url.
- When opening the deep link, we check the path mapping file if other forms of repo matching fail, and if matches are found, we show them in a quickpick and let the user either choose a path from the list or 'choose a different location' which reverts to manual choice.

### DIFF
--- a/src/uris/deepLinks/deepLink.ts
+++ b/src/uris/deepLinks/deepLink.ts
@@ -129,6 +129,7 @@ export const enum DeepLinkServiceAction {
 	DeepLinkStored,
 	DeepLinkErrored,
 	OpenRepo,
+	RepoMatchedInLocalMapping,
 	RepoMatchedWithId,
 	RepoMatchedWithPath,
 	RepoMatchedWithRemoteUrl,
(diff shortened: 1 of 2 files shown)


axosoft-ramint:
One thing that is left out here (intentionally) is validating the repo paths coming from the mapping file. It is left out because if feels a bit heavy to do `getOrOpenRepository` with `closeOnOpen: true` for every matching repo path in the file to filter out the ones that are no longer valid.

If we feel it would be useful, I can include that validation/filtering before showing the options to the user (and also clean up any paths that are invalid from the file).

@eamodio let me know if we want these extra bells and whistles. Can even add them post-merge.

eamodio:
@axosoft-ramint I think this looks good to merge. I agree with not bothering to validate the repo, but maybe a lightweight existence check on the folder might be worth it. Though I wouldn't hold up getting this in for it.

eamodio (approved):

PR #2894, opened 2023-08-30
Improve comparison deep link processing from PRs

### LINKED ISSUE #2893: Deep Link to Comparison: Add option to add remote if missing
When resolving a deep link to a comparison, if you don't have a matching remote it just shows a blank comparison rather than offering you the option to add the remote.

axosoft-ramint:
Fixes #2893

This PR achieves three main improvements to deep link processing for comparison links from PRs (using the web extension):

1. It now also considers the PR remote url in addition to the base url (the base url is the URL of the remote of the branch being merged, and the PR remote url is the url of the remote of the branch being merged into) for remote matching. If you are missing either remote, it offers the ability to add that remote.
2. It now supports branches that use the owner as the remote name. For example, if my branch name was `axosoft-ramint/main`, the matching branch may be `origin/main`. Previously we could not determine that this is the ref we really want; now it tries remote owner matching to resolve to the correct ref/sha.
3. We were doing a hacky workaround in the service by storing ref names as shas - the code was refactored to properly separate sha from id for a ref/target.

### DIFF
--- a/src/uris/deepLinks/deepLink.ts
+++ b/src/uris/deepLinks/deepLink.ts
@@ -53,6 +53,7 @@ export interface DeepLink {
 	repoPath?: string;
 	targetId?: string;
 	secondaryTargetId?: string;
+	secondaryRemoteUrl?: string;
 }
 
 export function parseDeepLinkUri(uri: Uri): DeepLink | undefined {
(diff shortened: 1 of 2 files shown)


axosoft-ramint:
Note @eamodio: even after this is merged, deep links to comparisons for PRs using the web extension won't work until a bug is addressed on the web extension's side: it needs to prepend the remote owner name to the branch name for each branch name in the deep link to handle PRs from forks.

For example, if this PR was coming from my own fork of gitlens, the branch names in the deep link should look like: `gitkraken/main...axosoft-ramint/bugs/2893-deep-link-to-comparison-add-option`. Currently it instead just uses the base branch names in fork scenarios i.e. `main...bugs/2893-deep-link-to-comparison-add-option`.

PR #3009, opened 2023-11-09
Adds basic welcome section for cloud patches

axosoft-ramint:


axosoft-ramint:
Currently it looks like this:

image

Before merging:

1. The text/wording needs a review
2. We need at least one image of the feature to put in the section
3. The [help center link](https://help.gitkraken.com/gitlens/gitlens-features/#cloud-patches) currently doesn't have a Cloud Patches section, so that should be added in.

axosoft-ramint:
Since we probably don't have time to get images and other content for this section, I'll just merge this as-is for now and we can improve upon it later.

It should meet the requirements of the [original ticket](https://gitkraken.atlassian.net/browse/GLVSC-361).

PR #3067, opened 2023-12-20
Fix indefinite line tracker suspension after closing large dirty editor

### LINKED ISSUE #3066: Dirtying Large File Indefinitely Suspends Git Line Blame
## Description

I love GitLens and get a lot of value out of it. The "_GitLens Current Line Blame_" is particularly valuable and something I use all the time. Multiple times a day, I've noticed VS Code get into a state where the line blame disappears.

Here's a screenshot of the feature I'm referring to:

Screenshot 2023-12-19 at 9 27 35 PM

Here's a screenshot of what it looks like when the feature mysteriously disappears. When this happens, it completely disappears and is not present as a status bar item when I right click on the bottom status bar.

Screenshot 2023-12-19 at 9 28 25 PM

I usually have to restart the VS Code extension server to get out of this state.

## Investigating

I've had this problem for multiple years. It's been very difficult to replicate since it usually happens a few hours into my work day and hadn't appeared in the past when I cloned the GitLens repo and started an [_Extension Development Host_ window](https://code.visualstudio.com/api/get-started/your-first-extension#debugging-the-extension).

Today, I was feeling particularly motivated to debug and figured out how to attach an interactive debugger with source maps to an already running Extension Host process after it got into this bad state. 😎

I noticed that `this.updateState` is never called when I get into the bad state due to `this.suspended` always evaluating to `true` here.

Screenshot 2023-12-19 at 4 54 34 PM

## Theory

I'm not exactly sure how `this.suspended` became `true` indefinitely in my editor today, but reading GitLens's source code, I see one reproducible way this could happen:

1. A user edits a file causing the editor state to become  "dirty" and `this.suspend()` to be called.

https://github.com/gitkraken/vscode-gitlens/blob/a70e2b930a9e10697b919b158dbf3e244a8bdd5c/src/trackers/gitLineTracker.ts#L107-L113

2. After the user stops editing the file for 5 seconds (the `advanced.blame.delayAfterEdit` time), the `onDirtyIdleTriggered` callback is fired and `this.resume()` is _supposed_ to unsuspend `GitLineTracker`.

https://github.com/gitkraken/vscode-gitlens/blob/a70e2b930a9e10697b919b158dbf3e244a8bdd5c/src/trackers/gitLineTracker.ts#L95-L100

As the code above shows, this intentionally doesn't resume the `GitLineTracker` if the dirty file is `>5000` lines long (or `advanced.blame.sizeThresholdAfterEdit`). That part feels correct.

However, what's surprising (and this code doesn't seem to handle), is the case when users switch tabs or open a different file after the idle timeout. The `GitLineTracker` shared instance is still suspended for the previous large file and never resumes for other smaller files in other tabs.

## Steps to Reproduce

I was excited to see the theory above is indeed reproducible. Using the current latest version at the time of writing (`v14.6.1`) and the GitLens code base itself:

1. Open `src/trackers/gitLineTracker.ts` and observe the current line blame updating properly as the cursor moves.
2. Open `yarn.lock` (currently at `7582` lines) and make an edit.
3. Wait 5 seconds.
4. Close the file and do not save changes.
5. Observe the current line blame no longer appear in `gitLineTracker.ts` or any other file.

https://github.com/gitkraken/vscode-gitlens/assets/906558/29e44897-ce46-47d9-a6f3-235404f03a37

## Pull Request

Here's one possible fix. https://github.com/gitkraken/vscode-gitlens/pull/3067

----

### GitLens Version

v14.6.1

### VS Code Version

Version: 1.85.1
Commit: 0ee08df0cf4527e40edc9aa28f4b5bd38bbff2b2
Date: 2023-12-13T09:48:16.874Z
Electron: 25.9.7
ElectronBuildId: 25551756
Chromium: 114.0.5735.289
Node.js: 18.15.0
V8: 11.4.183.29-electron.0
OS: Darwin arm64 23.2.0

### Git Version

git version 2.43.0

gluxon:
# Description

Fixes #3066. See issue link for an in-depth explanation of the bug, steps to reproduce, and a video.

# Checklist

### DIFF
--- a/src/trackers/lineTracker.ts
+++ b/src/trackers/lineTracker.ts
@@ -44,6 +44,7 @@ export class LineTracker<T> implements Disposable {
 		this._selections = toLineSelections(editor?.selections);
 
 		this.notifyLinesChanged('editor');
+		this.onActiveEditorChanged?.();
 	}
 
 	private onTextEditorSelectionChanged(e: TextEditorSelectionChangeEvent) {
@@ -196,6 +197,12 @@ export class LineTracker<T> implements Disposable {
 		this.onSuspend?.();
 		this.notifyLinesChanged('editor');
 	}
+
+	/**
+	 * Called after the active editor is changed and line tracker state has been
+	 * updated for it.
+	 */
+	protected onActiveEditorChanged?(): void;
 
 	protected fireLinesChanged(e: LinesChangeEvent) {
 		this._onDidChangeActiveLines.fire(e);
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 ### Fixed
 
 - Fixes [#2926](https://github.com/gitkraken/vscode-gitlens/issues/2926) in more cases - "Open File at Revision" has incorrect editor label if revision contains path separator &mdash; thanks to [PR #3060](https://github.com/gitkraken/vscode-gitlens/issues/3060) by Ian Chamberlain ([@ian-h-chamberlain](https://github.com/ian-h-chamberlain)
+- Fixes [#3066](https://github.com/gitkraken/vscode-gitlens/issues/3066) - Editing a large file and switching away to another file without saving causes current line blame to disappear; thanks to [PR #3067](https://github.com/gitkraken/vscode-gitlens/pulls/3067) by Brandon Cheng ([@gluxon](https://github.com/gluxon)).
 
 ## [14.6.1] - 2023-12-14
 
--- a/README.md
+++ b/README.md
@@ -427,6 +427,7 @@ A big thanks to the people that have contributed to this project 🙏❤️:
 - WofWca ([@WofWca](https://github.com/WofWca)) &mdash; [contributions](https://github.com/gitkraken/vscode-gitlens/commits?author=WofWca)
 - 不见月 ([@nooooooom](https://github.com/nooooooom)) &mdash; [contributions](https://github.com/gitkraken/vscode-gitlens/commits?author=nooooooom)
 - Ian Chamberlain ([@ian-h-chamberlain](https://github.com/ian-h-chamberlain)) &mdash; [contributions](https://github.com/gitkraken/vscode-gitlens/commits?author=ian-h-chamberlain)
+- Brandon Cheng ([@gluxon](https://github.com/gluxon)) &mdash; [contributions](https://github.com/gitkraken/vscode-gitlens/commits?author=gluxon)`
 
 Also special thanks to the people that have provided support, testing, brainstorming, etc:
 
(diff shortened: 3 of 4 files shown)


src/trackers/lineTracker.ts
```diff
@@ -44,6 +44,7 @@ export class LineTracker<T> implements Disposable {
 		this._selections = toLineSelections(editor?.selections);
 
 		this.notifyLinesChanged('editor');
+		this.onActiveEditorChanged?.();
```
gluxon, line 47: I was brainstorming a few alternatives to this approach.

1. Instead of the indirection through `GitLineTracker`, we could also call `this.resume()` right here. That felt strange since nothing else in `lineTracker.ts` suspends or resumes the tracker. This feels like a `gitLineTracker.ts` responsibility based on what's written today.
2. We could call `this.fireDocumentDirtyStateChanged` in the `window.onDidChangeActiveTextEditor` handler. This feels like a reasonable alternative approach to me. The main downside is that it's a bit of a slippery slope with regard to what document-specific change events are re-fired when the active editor changes. Ex: Would we also want to call `onBlameStateChanged` when the active editor changes?

I settled on the current approach, but happy to consider others in more depth if others have thoughts.

gluxon:
## Testing

Here's a test double checking that the original issue is fixed. A video of the original bug is present in https://github.com/gitkraken/vscode-gitlens/issues/3066.

https://github.com/gitkraken/vscode-gitlens/assets/906558/6f6ee65a-4918-4b05-8946-f202bd727ed8

gluxon:
I found it surprising that the `GitLineTracker` disappears completely while a file is dirty for 5 seconds and completely disappears for large dirty files. It makes sense that GitLens shouldn't compute diff frequently while editing, but I wonder if we should show its status in the user interface?

One idea would be to show an hourglass similar to what GitLens currently does as you move between lines. On hover, the indicator would show why it's delaying blame computation with a suggestion to tweak `advanced.blame.sizeThresholdAfterEdit` or `advanced.blame.delayAfterEdit`.

eamodio:
Thank you so much for your contribution here! Great investigation work and resolution! I've merged this directly, but I will be making some further changes to unify the LineTracker and GitLineTracker (no real need for the separation at this point), and will look into some kind of "notification" statusbar or something to let the user know why things are suspended and be able to point them to changing values to tweak it if you want more responsiveness and can spare the extra compute overhead

Thank you!

gluxon:
Thanks so much for reviewing this! Further changes post-merge are all welcome. I appreciate you taking the time to consider my thoughts.

eamodio:
Refactored things a bit here:
https://github.com/gitkraken/vscode-gitlens/commit/e522866348b3fa6fc8ede895eb6cf5517ba9152f

And added the following when the delay is active:

image

And this when the file is over the threshold:

image

gluxon:
That looks fantastic! Love it.

PR #3089, opened 2024-01-05
Updates to gk.dev for GK login flows

axosoft-ramint:
Also splits up 'login' and 'register' depending on the link clicked instead of just defaulting to login.


src/plus/gk/account/authenticationProvider.ts
```diff
@@ -75,9 +75,10 @@ export class AccountAuthenticationProvider implements AuthenticationProvider, Di
 		// Ensure that the scopes are sorted consistently (since we use them for matching and order doesn't matter)
 		scopes = scopes.sort();
```
eamodio, line 76: I think we need to remove the `signUp` scope from scopes before continuing, because we use the scopes as a key to look up existing sessions, and we don't want that difference to end up forcing a different session.

src/plus/gk/account/authenticationConnection.ts
```diff
@@ -69,25 +69,24 @@ export class AuthenticationConnection implements Disposable {
...
-				callbackUri.toString(),
-			)}`,
+		const uri = this.connection.getGkDevAccountsUri(
+			signUp ? 'register' : 'login',
+			`${scopes.includes('gitlens') ? 'source=gitlens&' : ''}state=${encodeURIComponent(
+				gkstate,
+			)}&redirect_uri=${encodeURIComponent(callbackUri.toString())}`,
 		);
-		void (await env.openExternal(uri));
+		// @ts-expect-error openExternal does allow a string, though the typing doesn't reflect it. Must pass as a string to avoid a double encoding
```
eamodio, line 81: Rather than using // @ts-expect-error, we could cast, as we do here:

https://github.com/gitkraken/vscode-gitlens/blob/690e17403f098e00bb3d3dc61e645ce8128598e3/src/git/remotes/remoteProvider.ts#L172-L174

eamodio: We could also add that to a helper function in system/utils.ts or something too

src/plus/gk/account/authenticationProvider.ts
axosoft-ramint: Done in commit: [Pulls signUp flag out of scopes once used](https://github.com/gitkraken/vscode-gitlens/pull/3089/commits/757e1ba6f0c1955855876985fb7f0c397c9979e2)

src/plus/gk/account/authenticationConnection.ts
axosoft-ramint: Done in commit: [Switches to existing fn for opening external URL](https://github.com/gitkraken/vscode-gitlens/pull/3089/commits/43f057c4993c87c5c842d8d8a7a4b476e417bd4f)

eamodio (approved):
Looks great!

PR #3094, opened 2024-01-08
Adds file/lines deep link support

axosoft-ramint:
Adds new commands:

 - `Copy Link to File`
 - `Copy Link to File at Reference...`
 - `Copy Link to Lines`
 - `Copy Link to Lines at Reference...`

The "at reference" commands allow you to choose a ref as an additional step, which will be attached to the link. This will open the file at that reference when used.

Currently, the commands appear in only a few places:
 - `Copy As` submenu of the editor context menu on a file (`Copy Link to Lines` commands will only appear if you have a line or lines selected)
 - Files within GitLens views

In addition, adds support for opening a file deep link.

Format is:

`vscode://eamodio.gitlens/link/r/{repoId}/f/{filePath}?url={remoteUrl}&lines={lines}&ref={refId}`

where:

`filePath` represents the path to the file relative to the repo root i.e. `src/commands/copyDeepLink.ts`

and in the query (both of these params are optional)...

`lines={lines}` represents the lines, can be one line like `2` or multiple like `103-108`

`ref={refId}` tells you what target to use.  Can be a fully qualified ref i.e. `refs/heads/main` or just a short name like `main` or a commit SHA, tag name, etc. If not provided we just go to the working copy of the file


src/uris/deepLinks/deepLinkService.ts
```diff
@@ -1108,4 +1210,71 @@ export class DeepLinkService implements Disposable {
...
+		}
+
+		const deepLinkRedirectUrl = new URL(
+			`https://${modePrefixString}gitkraken.dev/link/${encodeURIComponent(
+				Buffer.from(deepLink.href).toString('base64'),
+			)}`,
+		);
+
+		deepLinkRedirectUrl.searchParams.set('origin', 'gitlens');
+		return deepLinkRedirectUrl;
```
axosoft-ramint, line 1278: Currently sitting in draft state because gitkraken.dev has a bug with file paths preventing these links from working. PR to address that here: https://github.com/gitkraken/gitkraken.dev/pull/312

In the meantime you can make the below modification in `generateFileDeepLinkUrl` to test it:

```suggestion
		// return deepLinkRedirectUrl;
		return deepLink;
```

axosoft-ramint:
Other places where the commands can be added:

- In the editor line number context menu
- In the editor title menu
- In the explorer context menu
- In the SCM file context menu

All of the above, though, were omitted because it's not clear in these views that we are creating a GitLens-specific deep link, and GitLens has no submenu within them. We need to discuss how these commands would look in these areas.

axosoft-ramint:
Another note: I decided to separate out file/line deep link commands from the rest because the overloads for those classes/functions were getting messy and unwieldy, and trying to fit in file/line support into the existing functions as overloads would make it even messier.

axosoft-ramint:
@eamodio @d13 PR is ready for review. The blocking issue in gkdev has been fixed in all environments.

eamodio (approved):

PR #3114, opened 2024-01-19
Ensures split between hosted/self-hosted integrations

eamodio:
Fixes possible auth issues with self-hosted integrations

### DIFF
--- a/src/plus/integrations/providerIntegration.ts
+++ b/src/plus/integrations/providerIntegration.ts
@@ -26,22 +26,30 @@ import type {
 	PagedProjectInput,
 	PagedRepoInput,
 	ProviderAccount,
+	ProviderId,
 	ProviderIssue,
 	ProviderPullRequest,
 	ProviderRepoInput,
 	ProviderReposInput,
+	SelfHostedProviderId,
 } from './providers/models';
-import { IssueFilter, PagingMode, ProviderId, PullRequestFilter } from './providers/models';
+import { HostedProviderId, IssueFilter, PagingMode, PullRequestFilter } from './providers/models';
 import type { ProvidersApi } from './providers/providersApi';
 
-// TODO@eamodio revisit how once authenticated, all remotes are always connected, even after a restart
-
 export type SupportedProviderIds = ProviderId;
-export type ProviderKey = `${SupportedProviderIds}|${string}`;
+export type SupportedHostedProviderIds = HostedProviderId;
+export type SupportedSelfHostedProviderIds = SelfHostedProviderId;
+export type ProviderKey = `${SupportedHostedProviderIds}` | `${SupportedSelfHostedProviderIds}:${string}`;
+export type ProviderKeyById<T extends SupportedProviderIds> = T extends SupportedHostedProviderIds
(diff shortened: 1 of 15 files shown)


src/plus/integrations/providers/gitlab.ts
```diff
@@ -157,19 +151,31 @@ export class GitLabIntegration extends ProviderIntegration<GitLabRepositoryDescr
...
+	}
+
+	protected get apiBaseUrl(): string {
+		return 'https://gitlab.com/api';
+	}
+}
+
+export class GitLabSelfHostedIntegration extends GitLabIntegrationBase<SelfHostedProviderId.GitHubEnterprise> {
+	readonly authProvider = enterpriseAuthProvider;
+	readonly id = SelfHostedProviderId.GitHubEnterprise;
```
axosoft-ramint, line 170: This should be `GitLabEnterprise`

src/plus/integrations/providerIntegration.ts
```diff
@@ -26,22 +26,30 @@ import type {
...
+export type ProviderKeyById<T extends SupportedProviderIds> = T extends SupportedHostedProviderIds
+	? `${SupportedHostedProviderIds}`
+	: `${SupportedSelfHostedProviderIds}:${string}`;
 
 export type RepositoryDescriptor = { key: string } & Record<string, unknown>;
 
-export abstract class ProviderIntegration<T extends RepositoryDescriptor = RepositoryDescriptor> {
+export abstract class ProviderIntegration<
+	ID extends SupportedProviderIds = SupportedProviderIds,
+	T extends RepositoryDescriptor = RepositoryDescriptor,
```
axosoft-ramint, line 51: While we are refactoring all this, can we do away with requiring a repository descriptor for an integration? Certain integrations (like Jira, which I'm looking at now) don't support repositories - they're just issue providers that work with boards or projects.

axosoft-ramint (commented):
Some preliminary feedback on a quick scan before I start testing

src/plus/integrations/providers/gitlab.ts
eamodio: Please feel free to commit and take over this.

src/plus/integrations/providerIntegration.ts
eamodio: I'm not sure the implications of removing that yet (haven't had the time to look), but I would suggest doing that outside of this.

axosoft-ramint: Ok, no worries, I can include that in the Jira stuff.

src/plus/integrations/providers/gitlab.ts
axosoft-ramint: Roger that 👍🏼

axosoft-ramint (approved):
I don't have a GitLab Self-Hosted instance to test with, but given that it's currently broken, and given that this is correcting functions like `getByRemote` where the self-hosted context was previously being lost, I'd say it's safe to merge.

I did manage to test the Focus View since some updated functions touch that - it works fine 👍🏼

PR #3148, opened 2024-02-14
Refactor: renames to integrations and splits out issue integrations

axosoft-ramint:


src/plus/integrations/integration.ts
```diff
@@ -74,15 +87,20 @@ export abstract class ProviderIntegration<
 		return this.id;
 	}
 
-	private get connectedKey(): `connected:${ProviderIntegration['key']}` {
+	private get connectedKey(): `connected:${HostingIntegration['key']}` {
```
d13, line 90: Should this be `Integration['key']`?

axosoft-ramint: Well spotted! Updated it 👍🏼

eamodio (approved):
:shipit:

KidiIT (commented):
All reviewed and approved🫶❤️

PR #3200, opened 2024-03-25
Updates quick focus PR categorization

axosoft-ramint:
Uses shared provider utility and its corresponding categories rather than our own.

Note: This requires querying github for additional properties we were not previously, in order to supply the library with all the information it needs on a PR to categorize properly.

### DIFF
--- a/src/plus/focus/focus.ts
+++ b/src/plus/focus/focus.ts
@@ -35,12 +35,15 @@ import { groupAndSortFocusItems } from './focusProvider';
 
 const actionGroupMap = new Map<FocusActionCategory, string[]>([
 	['mergeable', ['Ready to Merge', 'Ready to merge']],
-	['mergeable-conflicts', ['Resolve Conflicts', 'You need to resolve merge conflicts, before this can be merged']],
+	['unassigned-reviewers', ['Unassigned Reviewers', 'You need to assign reviewers']],
 	['failed-checks', ['Failed Checks', 'You need to resolve the failing checks']],
 	['conflicts', ['Resolve Conflicts', 'You need to resolve merge conflicts']],
-	['needs-review', ['Needs Your Review', `\${author} requested your review`]],
+	['needs-my-review', ['Needs Your Review', `\${author} requested your review`]],
 	['changes-requested', ['Changes Requested', 'Reviewers requested changes before this can be merged']],
+	['reviewer-commented', ['Reviewers Commented', 'Reviewers have commented on this pull request']],
 	['waiting-for-review', ['Waiting for Review', 'Waiting for reviewers to approve this pull request']],
+	['draft', ['Draft', 'Continue working on your draft']],
+	['other', ['Other', 'Other pull requests']],
 ]);
 
 const groupMap = new Map<FocusGroup, [string, ThemeIcon | undefined]>([
@@ -51,6 +54,8 @@ const groupMap = new Map<FocusGroup, [string, ThemeIcon | undefined]>([
 	['needs-attention', ['Needs Your Attention', new ThemeIcon('bell-dot')]], //comment-unresolved
 	['needs-review', ['Needs Your Review', new ThemeIcon('comment-draft')]], // feedback
 	['waiting-for-review', ['Waiting for Review', new ThemeIcon('gitlens-clock')]],
+	['draft', ['Draft', new ThemeIcon('comment-discussion')]],
+	['other', ['Other', new ThemeIcon('question')]],
 	['snoozed', ['Snoozed', new ThemeIcon('bell-slash')]],
 ]);
 
--- a/src/plus/focus/focus.ts
+++ b/src/plus/focus/focus.ts
@@ -312,7 +312,7 @@ export class FocusCommand extends QuickCommand<State> {
 					state.item.id
 				} \u2022 ${fromNow(state.item.updatedDate)}`,
 				detail: interpolate(actionGroupMap.get(state.item.actionableCategory)![1], {
-					author: state.item.author,
+					author: state.item.author!.username,
 				}),
 				iconPath: state.item.author?.avatarUrl != null ? Uri.parse(state.item.author.avatarUrl) : undefined,
 			}),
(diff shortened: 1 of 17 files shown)


axosoft-ramint:
@eamodio @d13 This is up to date with the newest shared provider library util function and is ready to review. Hoping to get it in soon as there is a lot sitting on top of it. Please have a look when you can. Thanks!

axosoft-ramint:
@eamodio @d13 reminder to review this when you can. I have a lot sitting on top of it now - hoping it can go in soon so I can open up the remaining PRs. Thanks!

d13 (commented):
Some nits with enums

src/plus/integrations/providers/models.ts
```diff
@@ -479,3 +495,183 @@ export function toAccount(account: ProviderAccount, provider: ProviderReference)
...
+	[PullRequestReviewState.ReviewRequested]: 'REVIEW_REQUESTED' as GitPullRequestReviewState,
+	[PullRequestReviewState.Dismissed]: null,
+	[PullRequestReviewState.Pending]: null,
+};
+
+export const toProviderPullRequestMergeableState = {
+	[PullRequestMergeableState.Mergeable]: 'MERGEABLE' as GitPullRequestMergeableState,
+	[PullRequestMergeableState.Conflicting]: 'CONFLICTS' as GitPullRequestMergeableState,
+	[PullRequestMergeableState.Unknown]: 'UNKNOWN' as GitPullRequestMergeableState,
+};
```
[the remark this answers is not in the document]
axosoft-ramint: Ahh yes thanks for reminding me! When I originally wrote this, these enums were being exported as types from the shared library so I couldn't use their values. That has since been fixed, and I forgot to update these.

Done in commit: [Uses enum values](https://github.com/gitkraken/vscode-gitlens/pull/3200/commits/f89dd9bf9d93986293ffb27bcdf199efe003357a)

Nice catch!

src/plus/focus/focus.ts
```diff
@@ -51,6 +54,8 @@ const groupMap = new Map<FocusGroup, [string, ThemeIcon | undefined]>([
 	['needs-attention', ['Needs Your Attention', new ThemeIcon('bell-dot')]], //comment-unresolved
 	['needs-review', ['Needs Your Review', new ThemeIcon('comment-draft')]], // feedback
 	['waiting-for-review', ['Waiting for Review', new ThemeIcon('gitlens-clock')]],
+	['draft', ['Draft', new ThemeIcon('comment-discussion')]],
```
eamodio, line 57: There is a draft PR icon in codicons

src/plus/focus/focusProvider.ts
```diff
@@ -313,230 +328,86 @@ export class FocusProvider implements Disposable {
...
-		// 		items.sort((a, b) => (a.pinned ? -1 : 1) - (b.pinned ? -1 : 1) || b.sortTime - a.sortTime),
-		// 	);
-		// }
-
-		this.updateGroupedIds(categorized);
-		this._onDidRefresh.fire({ items: categorized });
-		return categorized;
-	}
+		if (prs != null) {
+			const github = await this.container.integrations.get(HostingIntegrationId.GitHub);
```
eamodio, line 349: Can we move this up to be in parallel with the PR request (its too bad we need to make another call, since we have the viewer information when we make the PR request)

src/plus/focus/focusProvider.ts
```diff
@@ -313,230 +328,86 @@ export class FocusProvider implements Disposable {
...
-										repoName: item.issue.repository?.repo,
-									},
-							  }
-							: undefined,
-			  };
+						!this._groupedIds.has(`${item.uuid}:${focusCategoryToGroupMap.get(actionableCategory)}`),
+					actionableCategory: actionableCategory,
+					suggestedActions: suggestedActions,
+				};
+			}) as FocusItem[];
```
eamodio, line 396: Can we avoid the `as` here -- maybe `satisfies` or at the type onto the `.map` call?

src/plus/integrations/integration.ts
```diff
@@ -362,6 +362,31 @@ export abstract class IntegrationBase<
...
+					return account;
+				} catch (ex) {
+					return this.handleProviderException<Account | undefined>(ex, scope, undefined);
+				}
+			})(),
+		}));
+		return currentAccount;
+	}
+
+	protected abstract getProviderCurrentAccount(
```
eamodio, line 386: We can make this optional rather than `abstract` so that the providers that don't support it, don't need a dummy implementation.

I had an example of that in the Cloud Patch I sent with the PR search changes.

eamodio (approved):
Minor comments -- ready to merge imo

axosoft-ramint:
Will address the above comments post-merge.

src/plus/focus/focusProvider.ts
axosoft-ramint: We have to use an `as` somewhere in here because even though the shared provider library function is expecting a certain shape for PRs, we are adding additional fields to the inputs which we expect to see on the output. To prevent TS from thinking we lost those fields, we need to cast until the util function typing can be updated.

For now I will use `satisfies` here and move the `as` up to the output of the `getActionablePullRequests` call to be more accurate with where the issue is.

src/plus/focus/focusProvider.ts
axosoft-ramint: We have to use an `as` somewhere in here because even though the shared provider library function is expecting a certain shape for PRs, we are adding additional fields to the inputs which we expect to see on the output. To prevent TS from thinking we lost those fields, we need to cast until the util function typing can be updated.

For now I will use `satisfies` here and move the `as` up to the output of the `getActionablePullRequests` call to be more accurate with where the issue is.

src/plus/focus/focusProvider.ts
axosoft-ramint: I have it here to avoid making the extra api call unless we actually have PRs that we want to categorize. If I move it up with the PR request we always have an extra api call even when the output isn't used.

PR #3212, opened 2024-04-02
Optimizes theme color derivatives

d13:
This moves colors that can be generated using the CSS `color-mix()` function out of JavaScript code and back into CSS files. The remaining computed colors will require cross-browser support of CSS [Relative Color functions](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_colors/Relative_colors) and will be updated in the future.


eamodio (approved):
:shipit:

PR #3214, opened 2024-04-03
Adds cloud integrations syncing and ux

axosoft-ramint:


axosoft-ramint:
The [blocking dependency](https://gitkraken.atlassian.net/browse/GKDOTDEV-513) has been released, so this PR is clear for review @eamodio @d13.

Note that there is not yet any change to our Autolinks settings page because it doesn't seem capable of using state in its current form.

Perhaps I can just add a simple text with a link, like "Connecting the Jira cloud integration with your GitKraken account will give you access to Jira autolinks"?

src/plus/integrations/authentication/jira.ts
```diff
@@ -1,111 +1,70 @@
...
-	error: string | null;
-}; */
-
-type providerInput = 'jira' | 'trello' | 'gitlab' | 'github' | 'bitbucket' | 'azure';
-
 export class JiraAuthenticationProvider implements IntegrationAuthenticationProvider {
-	constructor(private readonly connection: ServerConnection) {}
+	constructor(
+		private readonly container: Container,
+		private readonly getCloudIntegrationsApi: () => Promise<CloudIntegrationsApi>,
```
eamodio, line 18: If we move the access to the `Container` we can avoid having to bind & pass this here

src/plus/integrations/authentication/jira.ts
```diff
@@ -1,111 +1,70 @@
...
-		return (await authorizeRsp.json())?.data as Promise<ProviderAuthorizationData | undefined>;
-	}
-
 	async createSession(
 		descriptor?: IntegrationAuthenticationSessionDescriptor,
+		options?: { authorizeIfNeeded?: boolean },
 	): Promise<ProviderAuthenticationSession | undefined> {
-		let tokenData = await this.getTokenData(this.authProviderId);
+		const api = await this.getCloudIntegrationsApi();
+		let tokenData = await api.getTokenData(this.authProviderId);
```
eamodio, line 32: Should we move the "refreshing" into this call (on a flag) or should we keep the control at the provider level?

src/plus/integrations/integrationService.ts
```diff
@@ -44,21 +50,71 @@ export class IntegrationService implements Disposable {
...
 	}
 
 	dispose() {
 		this._disposable?.dispose();
 	}
 
+	private registerCommands(): Disposable[] {
+		return [registerCommand(Commands.PlusManageCloudIntegrations, () => this.manageCloudIntegrations())];
+	}
+
```
eamodio, line 76: Since this service is delay loaded (not until access), we'll need to move the command elsewhere -- probably just create a new command similar to https://github.com/gitkraken/vscode-gitlens/blob/4fa1227c8c99039de09d4b911bd3f84beae0ddeb/src/commands/resets.ts#L9-L18

src/webviews/apps/home/home.html
```diff
@@ -396,6 +396,14 @@ <h2 class="nav-list__title t-eyebrow">Configuration</h2>
 					><code-icon class="nav-list__icon" icon="link"></code-icon
 					><span class="nav-list__label">Autolinks Settings</span></a
 				>
+				<a
```
eamodio, line 399: Let's do the following:
- Move the GitLens settings to the top, above Popular Views -- @d13 what do you think?
- Create a new Integrations section and put integrations first and autolinks 2nd, and place if after the activity bar section

eamodio (approved):
I'm good with merging and addressing these changes in a follow up. Since only the command issue is a functional one.

axosoft-ramint:
Will address the above comments post-merge.

src/plus/integrations/authentication/jira.ts
axosoft-ramint: The "refresh" is currently automatically performed in this call when your retrieved token is expired. I don't see a need to ever have that not be the case, so I'm not sure if we even need it manually controlled.

src/webviews/apps/home/home.html
d13: I like it

PR #767, opened 2019-06-12
Fixes #692 - make remote work with recent versions of Bitbucket Server

grozan:
# Description

recent versions of Bitbucket server (formerly Stash) introduce an extra /scm in the URL, compared to older versions.
If it's detected in the path it's removed, therefore should work with all versions of Bitbucket server.
Tested with BB v6.1.2

# Checklist

- [x ] I have followed the guidelines in the Contributing document
- [x ] My changes are based off of the `develop` branch
- [x ] My changes follow the coding style of this project
- [x ] My changes build without any errors or warnings
- [x ] My changes have been formatted and linted
- [x ] My changes include any required corresponding changes to the documentation
- [x ] My changes have been rebased and squashed to the minimal number (typically 1) of relevant commits
- [x ] My changes have a descriptive commit message with a short title, including a `Fixes $XXX -` or `Closes #XXX -` prefix to auto-close the issue that your PR addresses

PR #3219, opened 2024-04-06
Fixes #3218 - fix bitbucket-server splitPath "scm/" regression

### LINKED ISSUE #3218: Bitbucket Server remote - "scm/" path prefix not removed, possible regression from #767?
### Description

Hi, I've been having issues getting my Bitbucket Server remote setup with the most recent version of GitLens. After searching through related issues in the issue tracker, what I could best describe is that I've been experiencing a combination of #267 and #692.

My remote URL look something like this: `https://domain/scm/web/someRepo.git`, assuming project name is "web" and repo name is "someRepo". From what I understand, the original fix for #267 would work if my URL looked like: `https://domain/web/someRepo.git`. However, because I have "scm/" prefix as well, this ends up turning `repoBase` into `scm` instead of `web`, and `repoPath` becomes `web/someRepo` instead of `someRepo`.

From what I gather, PR #767 (722a7d213b52a14595f734c65a68a90e37648304) originally fixed this issue for Bitbucket Server URLs, by providing a special case for this `scm/` prefix. However, I noticed in the following change from 5 months ago (https://github.com/gitkraken/vscode-gitlens/commit/ba0a6c87bfb29421b22fcab1aac5455605be94c0#diff-3df44eca214dcc6e95c893331e04e784c958b3167e664a01dd724acb68cdefefR54), the "scm/" handling was moved into a new `splitPath` override method, but the part that does the actual splitting splits the original path instead of the modified path (without the "scm/"), which is not what the previous version did. This I believe is causing the "scm/" to still appear in transformed Bitbucket Server URLs, and also any project name over 3 characters long is also mangled because of this. I have more details in my Reproduction Steps below.

I already have a fix for this here: https://github.com/Coteh/vscode-gitlens/commit/0ba49ab459ab0e4eb17cc5f6c30b9ec48d18c499
I will put up a PR for this shortly if that's ok?

## Reproduction Steps

If you need them, here are some reproduction steps that you can do to replicate the issue on your end. (I guess you'll need a Bitbucket Server as well to test, or something that mimics it)

1. Use the following config in user settings (modify domain and name to match your remote)
```json
{
    "gitlens.integrations.enabled": true,
    "gitlens.remotes": [
        {
            "domain": "domain",
            "type": "BitbucketServer",
            "name": "My Server",
            "protocol": "https"
        }
    ]
}
```
2. Attempt to open any remote link (such as when clicking on the globe icon when the commit overlay appears on a line)
3. Remote link should properly contain a structure like `projects/web/repos/someRepo`. Instead it ends up being `projects/scm/repos/web/someRepo`.

This is further exacerbated when a project name is more than 3 characters long. The index that `splitPath` method inside bitbucket-server uses when splitting is going off of the new transformed `path` without `scm/` prefix, but it's being applied to the original path with `scm/` prefix in it. This will cause project names to have a slash and other extra characters and the repo name to be cut off. For example, if my project name was "tests", the transformed path would contain something like `projects/scm/t/repos/sts/someRepo`, the "e" would be cut off because that's where the index was in the transformed path "tests/someRepo".

### GitLens Version

v14.9.0

### VS Code Version

Version: 1.87.2 (Universal)
Commit: 863d2581ecda6849923a2118d93a088b0745d9d6
Date: 2024-03-08T15:21:31.043Z
Electron: 27.3.2
ElectronBuildId: 26836302
Chromium: 118.0.5993.159
Node.js: 18.17.1
V8: 11.8.172.18-electron.0
OS: Darwin arm64 23.2.0

### Git Version

git version 2.44.0

_No response_

Coteh:
# Description

As discussed in #3218, this is the fix for the bitbucket-server remote integration, to correctly remove any "scm/" prefixes in the remote links if they're present.

This was originally fixed in #767, but then the following change (https://github.com/gitkraken/vscode-gitlens/commit/ba0a6c87bfb29421b22fcab1aac5455605be94c0#diff-3df44eca214dcc6e95c893331e04e784c958b3167e664a01dd724acb68cdefefR54) introduced what I believe is a typo in the substringing portion that regressed the behaviour. It should be splitting substrings on transformed path, not the original path.

# Checklist

- [X] My changes include any required corresponding changes to the documentation (including CHANGELOG.md and README.md)
- [X] My changes have a descriptive commit message with a short title, including a `Fixes $XXX -` or `Closes #XXX -` prefix to auto-close the issue that your PR addresses


troopyq:
Thanks a lot! I wanted to correct it myself the other day) you’re cool

bengeisler-syskron:
It would be great if this PR could be merged. ❤️ 
I'm also heavily relying on this feature when communicating with my team.

Coteh:
I just tested @sergeibbb's recent fix (f18c6b6645fd773431a28f11b8cca6d04e18edbf) locally and it works. This is no longer needed. Closing.

PR #3234, opened 2024-04-16
Makes Quick Focus and Indicators responsive to GitHub connection/disconnection

axosoft-ramint:
Previously, the quick focus and indicators would get into weird states if you're not connected to GitHub.

This PR addresses that as follows:
- New step in quick focus to confirm you have a supported integration connected, and offer ability to connect if not connected (GitHub only for now, but set to expand to new providers as added)
- New function in Focus Provider to check if a supported integration is connected (GitHub only for now)
- Updated state in Focus Indicator when not connected to GitHub, and a listener which updates the state when integration state changes. Includes a CTA link to connect to GitHub when disconnected.
- Special case logic in GitHub integration refresh to disconnect if we were disconnected in vscode (since our GH integration is controlled by vscode).


d13 (approved):
LGTM

PR #3236, opened 2024-04-17
Add features to DraftService

d13:
## Moves git-repositories lookup onto drafts

The `/v1/git-repositories/${repoId}` route in `RepositoryIdentityService.getRepositoryIdentity` is deprecated in favor of a route that directly associates it to a draft `/v1/drafts/${draftId}/git-repositories/${repoId}`. The `getRepositoryIdentity` method was moved to `DraftService` and now implements the new route.

The main difference in the `RepositoryIdentityService` functionality, besides removing that method, is that now the methods only interact with `RepositoryIdentityDescriptor` and no longer need to handle `GkRepositoryId`.

## Utility function to create an EntityIdentifier

This converts issues and pull requests into a resource format that our APIs use to uniquely look them up.

## Creating Code Suggestions

Adds new params into DraftService.createDraft to support the `suggested_pr_change` type

### DIFF
--- a/package.json
+++ b/package.json
@@ -16927,7 +16927,7 @@
 	},
 	"dependencies": {
 		"@gitkraken/gitkraken-components": "10.2.13",
-		"@gitkraken/provider-apis": "0.21.0",
+		"@gitkraken/provider-apis": "0.22.1",
 		"@gitkraken/shared-web-components": "0.1.1-rc.15",
 		"@lit/react": "1.0.3",
 		"@microsoft/fast-element": "1.12.0",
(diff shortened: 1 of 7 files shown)


axosoft-ramint:
> Moves repository identity calls into the DraftService

Curious why these shouldn't live in the repo identity service? Seems like any repo identity stuff would fall under it's domain, no?

d13:
> > Moves repository identity calls into the DraftService
> 
> Curious why these shouldn't live in the repo identity service? Seems like any repo identity stuff would fall under it's domain, no?

Sorry about that, I should've added more context. The `/v1/git-repositories/${repoId}` route in `RepositoryIdentityService.getRepositoryIdentity` is deprecated in favor of a route that directly associates it to a draft  `/v1/drafts/${draftId}/git-repositories/${repoId}`.

axosoft-ramint (approved):
LGTM

PR #3242, opened 2024-04-22
Updates Focus UI strings

axosoft-ramint:
Updates user-facing strings to Launchpad.
Updates icon to rocket.

Note: Does not update any internal naming - this should be done in separate work, and telemetry events/command naming should be carefully reviewed.


eamodio (approved):
🚢

PR #3251, opened 2024-04-25
Overhauls launchpad settings and commands

axosoft-ramint:
-Renames existing settings and commands, removing experimental labeling. Uses the pattern `gitlens.launchpad` now.

-Adds new settings for launchpad and indicator:

`ignoredRepositories`: Array of repos of the form `{owner}/{repoName}` to exclude from the launchpad
`staleThreshold`: Value, in days ago, beyond which PRs are moved to the "other" action category in the launchpad
`indicator.useColors`: Boolean to specify if the status bar text should use color
`indicator.showTopItem`: Boolean to specify if the top item should be shown in the status bar text
`indicator.groups` Array to specify which of the supported indicator/status bar groups to show in the hover.


package.json
```diff
@@ -5317,19 +5380,19 @@
...
-				"command": "gitlens.focus.split",
+				"command": "gitlens.launchpad.split",
 				"title": "Split Launchpad in Editor",
 				"category": "GitLens",
 				"icon": "$(split-horizontal)"
 			},
 			{
-				"command": "gitlens.focus.experimental.updateIndicators",
-				"title": "Update Launchpad Indicators",
+				"command": "gitlens.launchpad.indicator.update",
```
eamodio, line 5389: Is this a refresh? Maybe call it `Refresh Launchpad Indicator`?

package.json
```diff
@@ -2790,40 +2790,103 @@
...
-					"gitlens.focus.experimental.indicators.data.refreshRate": {
+					"gitlens.launchpad.indicator.polling.interval": {
 						"type": "number",
 						"default": 30,
-						"markdownDescription": "(Experimental) Specifies the rate (in minutes) at which to refresh the status bar indicators for _Launchpad_. Use 0 to disable automatic refreshing",
+						"markdownDescription": "Specifies the rate (in minutes) at which the status bar indicator will fetch pull request data for _Launchpad_. Use 0 to disable automatic polling",
 						"scope": "window",
 						"order": 24
+					},
+					"gitlens.launchpad.indicator.showTopItem": {
```
eamodio, line 2848: What do you think of something like `gitlens.launchpad.indicator.style`

`icon` -- rocket icon
`priorityItem` (don't love the name at all, but can't think of something better off-hand) -- rocket icon + priority item name
`priorityGroupIcon` -- rocket icon if no priority group, otherwise the group icon?

Thoughts?

package.json
```diff
@@ -2790,40 +2790,103 @@
...
+						"type": "boolean",
+						"default": true,
+						"markdownDescription": "Specifies whether to show the top _Launchpad_ item in the status bar indicator text",
+						"scope": "window",
+						"order": 25
+					},
+					"gitlens.launchpad.indicator.useColors": {
+						"type": "boolean",
+						"default": true,
```
eamodio, line 2857: Lets default this off for now

eamodio (approved):
Looks good, we can either address my comments here or in a follow up

package.json
axosoft-ramint: It's not really a refresh - it's essentially a command to update one of the indicator settings. Maybe I can change it to `updateSettings`

package.json
axosoft-ramint: So `priorityItem` would be the equivalent of `showTopItem: true`, `icon` would be the equivalent of `showTopItem: false`, and `priorityGroupIcon` is similar to `showTopItem: false` except the icon itself changes with the top group?

package.json
axosoft-ramint: Ehh now that I think about it, it also does things like 'connect to GitHub' etc. so maybe `update` is fine for now until we can think of a better name for what this is doing:

```
			registerCommand('gitlens.launchpad.indicator.update', async (action: string) => {
				switch (action) {
					case 'mute':
						void configuration.updateEffective('launchpad.indicator.polling.enabled', false);
						break;
					case 'unmute':
						void configuration.updateEffective('launchpad.indicator.polling.enabled', true);
						break;
					case 'hide': {
						const action = await window.showInformationMessage(
							'Would you like to hide the Launchpad status bar icon? You can re-enable it at any time using the "GitLens: Toggle Launchpad Status Bar Icon" command.',
							{ modal: true },
							'Hide',
						);
						if (action === 'Hide') {
							void configuration.updateEffective('launchpad.indicator.enabled', false);
						}
						break;
					}
					case 'connectGitHub': {
						const github = await this.container.integrations?.get(HostingIntegrationId.GitHub);
						if (github == null) break;
						if (!(github.maybeConnected ?? (await github.isConnected()))) {
							void github.connect();
						}
						break;
					}
					default:
						break;
				}
			}
```

PR #3289, opened 2024-05-14
Fixes `launchpad/open` event always being sent on extension load

axosoft-ramint:
We call the constructor of all registrable events once when mapping them on extension load. This unfortunately causes the `launchpad/open` event to trigger, because we didn't include a `true` value for the `hidden` param in the constructor. This makes an exception for just that event in the mapping, so that the event doesn't trigger.

### DIFF
--- a/src/system/command.ts
+++ b/src/system/command.ts
@@ -9,7 +9,7 @@ import { isWebviewContext } from './webview';
 
 export type CommandCallback = Parameters<typeof commands.registerCommand>[1];
 
-type CommandConstructor = new (container: Container) => Command;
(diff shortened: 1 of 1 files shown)


axosoft-ramint:
Honestly, maybe we can just get rid of this event since it provides no more information than `launchpad/opened` already does? @eamodio WDYT?

eamodio:
Did we want to use it to tell us how long it takes the Launchpad to load and/or get abandoned before opened completes? 

If it was only timing we could just take a timestamp, but right now we don't have a good hook for closing to capture the abandoned

eamodio (approved):

PR #3303, opened 2024-05-17
Saves collapsed groups on launchpad between uses

axosoft-ramint:
- Stores collapsed groups as a string array in global storage
- Updates this when a user toggles a group, but only when the launchpad wasn't opened to a specific group (when clicking a specific group from the launchpad indicator hover tooltip)


eamodio (approved):
:shipit:

PR #3321, opened 2024-05-27
Adds hovers to the Graph (wip)

eamodio:
Adds hovers to the Graph. This is still a work-in-progress and should probably be moved into the graph component itself.

See [GLVSC-524](https://gitkraken.atlassian.net/browse/GLVSC-524)

[GLVSC-524]: https://gitkraken.atlassian.net/browse/GLVSC-524?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/package.json
+++ b/package.json
@@ -17511,7 +17511,7 @@
 		"vscode:prepublish": "yarn run bundle"
 	},
 	"dependencies": {
-		"@gitkraken/gitkraken-components": "10.3.0",
+		"@gitkraken/gitkraken-components": "10.4.0",
 		"@gitkraken/provider-apis": "0.22.9",
 		"@gitkraken/shared-web-components": "0.1.1-rc.15",
 		"@lit/react": "1.0.5",
--- a/package.json
+++ b/package.json
@@ -17531,6 +17531,7 @@
 		"https-proxy-agent": "5.0.1",
 		"iconv-lite": "0.6.3",
 		"lit": "3.1.4",
+		"marked": "12.0.2",
 		"node-fetch": "2.7.0",
 		"os-browserify": "0.3.0",
 		"path-browserify": "1.0.1",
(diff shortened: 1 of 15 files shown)


package.json
```diff
@@ -17505,6 +17505,7 @@
 		"https-proxy-agent": "5.0.1",
 		"iconv-lite": "0.6.3",
 		"lit": "3.1.3",
+		"marked": "12.0.2",
```
d13, line 17508: `yarn.lock` is not updated

eamodio: That was on purpose to avoid conflicts. I think that is something we should do generally, because it just causes issues when merging. Though it is needed for the ultimate merge so it adds another merge step.

eamodio:
Hovers are cached (in both the webview and in the exthost -- except for WIP as it needs to be able to update with changes). We also re-use the stats if the "Changes" column is loaded, though we aren't invalidating the hover caches if that changes.

I've also move the hover to be below the row (and offset from the mouse so you can more easily move up/down and mouse wheel scroll, but if the hover inverts right now all bets are off, so maybe @d13 can look more into that.

package.json
axosoft-ramint: Isn't it a good thing that it causes conflicts (i.e. wouldn't we want to catch/fix these conflicts ahead of time, rather than on merge?) or am I missing something?

axosoft-ramint (commented):
LGTM - definitely still some funkiness with scrolling, but it's still a better experience than current.

Once [this dependency PR](https://github.com/gitkraken/GitKrakenComponents/pull/350) is merged, I will bump the graph component version here and add the new property we need to suppress the graph's internal tooltips, and then we can merge.

PR #3345, opened 2024-06-08
Adds first parent toggle to Graph

### LINKED ISSUE #3313: Allow toggling `--first-parent` in commit graph
Only following the first parent of commits is an essential feature for viewing otherwise messy commit graphs.

This is the `--first-parent` option in `git log --graph`.

In the [vscode-git-graph](https://github.com/mhutchie/vscode-git-graph/) extension this is the difference it makes:

Turned off:

image

Turned on:

image

Ideally this would be an easy to toggle setting from the commit graph UI, and would apply per workspace

eamodio:
Closes #3313 - adds a new `Simply Merge History` filter option to the Commit Graph

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -2824,6 +2824,12 @@
 						"scope": "window",
 						"order": 30
 					},
+					"gitlens.graph.onlyFollowFirstParent": {
+						"type": "boolean",
+						"default": false,
+						"markdownDescription": "Specifies whether to only follow the first parent when showing commits on the _Commit Graph_",
+						"order": 31
+					},
 					"gitlens.graph.dateStyle": {
 						"type": [
 							"string",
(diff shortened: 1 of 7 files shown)


axosoft-ramint (approved):

nicooprat:
Great! Will it be available to the `Gitlens: commits` view too? I mean here:

image

eamodio:
@nicooprat No, only the Commit Graph right now. It should be pretty straightforward to add for the Commits view (in the `...` menu like the `Hide Merge Commits` option), but given our current priorities (and vacations of the next few weeks) it isn't something that is likely to land soon.

Though we'd be happy to accept a PR for it from the community in the meantime.

nicooprat:
I understand, thanks.

PR #3349, opened 2024-06-11
Replaces "Open in Editor" link in Launchpad with a link to gk.dev [GLVSC-504]

sergeibbb:
# Description

image

# Checklist

### DIFF
--- a/src/container.ts
+++ b/src/container.ts
@@ -234,7 +234,7 @@ export class Container {
 		this._disposables.push((this._keyboard = new Keyboard()));
 		this._disposables.push((this._vsls = new VslsController(this)));
 		this._disposables.push((this._eventBus = new EventBus()));
-		this._disposables.push((this._focusProvider = new FocusProvider(this)));
+		this._disposables.push((this._focusProvider = new FocusProvider(this, this._connection)));
 
 		this._disposables.push((this._fileAnnotationController = new FileAnnotationController(this)));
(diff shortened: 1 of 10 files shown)


src/container.ts
```diff
@@ -918,6 +918,31 @@ export class Container {
...
+		let uri = path != null ? Uri.joinPath(this.baseGkDevUri, path) : this.baseGkDevUri;
+		if (query != null) {
+			uri = uri.with({ query: query });
+		}
+		return uri;
+	}
+
+	generateWebGkDevUrl(path?: string): string {
+		return this.getGkDevUri(path, '?source=gitlens').toString();
+	}
```
eamodio, line 945: Do we need this? I'm not sure `source=gitlens` is something gk.dev can support on any URL, but if so, it seems like that should just be integrated into `getGKDevUri`

@schinizel can we add `source=gitlens` to any gk.dev route?

axosoft-ramint: I would move the `source=gitlens` part directly into `getGkDevUri`, as it seems based on responses from the team, that this would never break functionality on GKDev pages. Then we can get rid of this.

schinizel: https://gitkraken-hq.slack.com/archives/C05H8JUFK8Q/p1718141676567519

... while it won't break anything, why add it to every route/request? I will post in the Slack thread.

PR #3356, opened 2024-06-18
Converts GitHub integration authentication to use gk.dev GLVSC-554

sergeibbb:
# Description

Introduces a provider for GitHub integration that tries to check for existing GitHub session before accessing an integration on GKDev (GLVSC-554)

# Checklist

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -8,7 +8,7 @@ import type { Environment } from './container';
 import type { StoredSearchQuery } from './git/search';
 import type { Subscription, SubscriptionPlanId, SubscriptionState } from './plus/gk/account/subscription';
 import type { Integration } from './plus/integrations/integration';
-import type { IntegrationId, IssueIntegrationId } from './plus/integrations/providers/models';
+import type { HostingIntegrationId, IntegrationId, IssueIntegrationId } from './plus/integrations/providers/models';
 import type { TelemetryEventData } from './telemetry/telemetry';
 import type { TrackedUsage, TrackedUsageKeys } from './telemetry/usageTracker';
 
@@ -846,6 +846,7 @@ export type Sources =
 	| 'notification'
 	| 'patchDetails'
 	| 'prompt'
+	| 'remoteProvider'
 	| 'settings'
 	| 'timeline'
 	| 'trial-indicator'
@@ -874,6 +875,15 @@ export type SupportedAIModels =
 	| `google:${AIModels<'gemini'>}`
 	| `openai:${AIModels<'openai'>}`
 	| 'vscode';
+
+// export const supportedCloudIntegrationIds = [HostingIntegrationId.GitHub, IssueIntegrationId.Jira];
+// export type SupportedCloudIntegrationIds = (typeof supportedCloudIntegrationIds)[number];
+const supportedCloudIntegrationIds = ['github', 'jira'];
+export type SupportedCloudIntegrationIds = HostingIntegrationId.GitHub | IssueIntegrationId.Jira;
+
+export function isSupportedCloudIntegrationId(id: string): id is SupportedCloudIntegrationIds {
+	return supportedCloudIntegrationIds.includes(id as SupportedCloudIntegrationIds);
+}
 
 export type SecretKeys =
 	| `gitlens.integration.auth:${IntegrationId}|${string}`
@@ -1224,7 +1234,7 @@ export type TelemetryEvents = {
 	};
 	/** Sent when a user chooses to manage the cloud integrations */
 	'cloudIntegrations/settingsOpened': {
-		'integration.id': IssueIntegrationId | undefined;
+		'integration.id': SupportedCloudIntegrationIds | undefined;
 	};
 
 	/** Sent when a code suggestion is archived */
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -887,6 +887,7 @@ export function isSupportedCloudIntegrationId(id: string): id is SupportedCloudI
 
 export type SecretKeys =
 	| `gitlens.integration.auth:${IntegrationId}|${string}`
+	| `gitlens.integration.auth.cloud:${IntegrationId}|${string}`
 	| `gitlens.${AIProviders}.key`
 	| `gitlens.plus.auth:${Environment}`;
 
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -7,8 +7,9 @@ import type { FileAnnotationType, ViewShowBranchComparison } from './config';
 import type { Environment } from './container';
 import type { StoredSearchQuery } from './git/search';
 import type { Subscription, SubscriptionPlanId, SubscriptionState } from './plus/gk/account/subscription';
+import type { SupportedCloudIntegrationIds } from './plus/integrations/authentication/models';
 import type { Integration } from './plus/integrations/integration';
-import type { HostingIntegrationId, IntegrationId, IssueIntegrationId } from './plus/integrations/providers/models';
+import type { IntegrationId } from './plus/integrations/providers/models';
 import type { TelemetryEventData } from './telemetry/telemetry';
 import type { TrackedUsage, TrackedUsageKeys } from './telemetry/usageTracker';
 
@@ -875,13 +876,6 @@ export type SupportedAIModels =
 	| `google:${AIModels<'gemini'>}`
 	| `openai:${AIModels<'openai'>}`
 	| 'vscode';
-
-const supportedCloudIntegrationIds = ['github', 'jira'];
-export type SupportedCloudIntegrationIds = HostingIntegrationId.GitHub | IssueIntegrationId.Jira;
-
-export function isSupportedCloudIntegrationId(id: string): id is SupportedCloudIntegrationIds {
-	return supportedCloudIntegrationIds.includes(id as SupportedCloudIntegrationIds);
-}
 
 export type SecretKeys =
 	| `gitlens.integration.auth:${IntegrationId}|${string}`
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -876,8 +876,6 @@ export type SupportedAIModels =
 	| `openai:${AIModels<'openai'>}`
 	| 'vscode';
 
-// export const supportedCloudIntegrationIds = [HostingIntegrationId.GitHub, IssueIntegrationId.Jira];
-// export type SupportedCloudIntegrationIds = (typeof supportedCloudIntegrationIds)[number];
 const supportedCloudIntegrationIds = ['github', 'jira'];
 export type SupportedCloudIntegrationIds = HostingIntegrationId.GitHub | IssueIntegrationId.Jira;
 
--- a/src/plus/integrations/authentication/github.ts
+++ b/src/plus/integrations/authentication/github.ts
@@ -1,9 +1,43 @@
 import type { AuthenticationSession, Disposable, QuickInputButton } from 'vscode';
-import { env, ThemeIcon, Uri, window } from 'vscode';
+import { authentication, env, ThemeIcon, Uri, window } from 'vscode';
+import { wrapForForcedInsecureSSL } from '@env/fetch';
 import type { Container } from '../../../container';
-import { SelfHostedIntegrationId } from '../providers/models';
+import { HostingIntegrationId, SelfHostedIntegrationId } from '../providers/models';
 import type { IntegrationAuthenticationSessionDescriptor } from './integrationAuthentication';
 import { IntegrationAuthenticationProvider } from './integrationAuthentication';
+
+export class GitHubAuthenticationProvider extends IntegrationAuthenticationProvider {
+	constructor(container: Container) {
+		super(container, HostingIntegrationId.GitHub);
+	}
+
+	override async createSession(
+		descriptor?: IntegrationAuthenticationSessionDescriptor,
+		options?: { authorizeIfNeeded?: boolean; createIfNeeded?: boolean; forceNewSession?: boolean },
+	): Promise<AuthenticationSession | undefined> {
+		if (descriptor != null) {
+			const { createIfNeeded, forceNewSession } = options ?? {};
+			const session = wrapForForcedInsecureSSL(
+				this.container.integrations.ignoreSSLErrors({ id: this.authProviderId, domain: descriptor?.domain }),
+				() =>
+					authentication.getSession(this.authProviderId, descriptor.scopes, {
+						createIfNone: forceNewSession ? undefined : createIfNeeded,
+						silent: !createIfNeeded && !forceNewSession ? true : undefined,
+						forceNewSession: forceNewSession ? true : undefined,
+					}),
+			);
+
+			if (session != null) {
+				return session;
+			}
+		}
+		return super.createSession(descriptor, options);
+	}
+
+	protected override getCompletionInputTitle(): string {
+		throw new Error('Method not implemented.');
+	}
+}
 
 export class GitHubEnterpriseAuthenticationProvider extends IntegrationAuthenticationProvider {
 	constructor(container: Container) {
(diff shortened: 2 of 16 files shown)


axosoft-ramint (changes requested):
A couple of suggestions on the base class and override strategies - mainly suggesting that we leverage abstract classes/fns.

src/plus/integrations/integration.ts
```diff
@@ -278,6 +287,19 @@ export abstract class IntegrationBase<
...
+					source: 'integrations',
+					detail: {
+						action: 'connect',
+						integration: this.id,
+					},
+				};
+
+				Promise.resolve(this.container.subscription.loginOrSignUp(true, source)).catch(
+					Logger.error.bind(Logger),
+				);
```
axosoft-ramint, line 301: I think, instead of this, we should route connections to cloud integrations using `manageCloudIntegrations` of the integration service, with the ID of the integration being connected. You could set up a "supported cloud integration ids" constant, then in the `execute` logic of `ConnectRemoteProviderCommand`, you could reroute to the `ManageCloudIntegrationsCommand` using the integration id, source and detail (you may need to define a new source etc.).

axosoft-ramint (changes requested):
Apologies for multiple reviews - one more important thing I forgot in the previous review regarding "connect to remote".

src/plus/integrations/authentication/integrationAuthentication.ts
```diff
@@ -36,108 +39,228 @@ export interface IntegrationAuthenticationSessionDescriptor {
...
+		) {
+			return this.createAndStoreSession(descriptor);
 		}
 
-		if (descriptor == null) return undefined;
+		return storedSession as ProviderAuthenticationSession | undefined;
+	}
+}
 
+class BuiltInAuthenticationProvider extends IntegrationAuthenticationProvider {
```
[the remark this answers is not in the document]
sergeibbb: @axosoft-ramint 
It was [here](https://github.com/gitkraken/vscode-gitlens/blob/da47858caa97552f3ec350342f74bb1bd1c1964d/src/plus/integrations/authentication/integrationAuthentication.ts#L117-L126): the case for unsupported providers

image
image

If we remove the case, then what to do instead? Let it throw?

src/plus/integrations/authentication/integrationAuthentication.ts
```diff
@@ -36,108 +39,228 @@ export interface IntegrationAuthenticationSessionDescriptor {
 	[key: string]: unknown;
 }
 
-export interface IntegrationAuthenticationProvider {
-	getSessionId(descriptor?: IntegrationAuthenticationSessionDescriptor): string;
-	createSession(
+export class IntegrationAuthenticationProvider {
```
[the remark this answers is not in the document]
sergeibbb: @axosoft-ramint 

Is it close to what you mean: https://github.com/gitkraken/vscode-gitlens/compare/draft/GLVSC-554-before...draft/GLVSC-554-after

src/plus/integrations/authentication/integrationAuthentication.ts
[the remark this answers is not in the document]
axosoft-ramint: Well, now that I look again, we have some other providers like "Gerrit" and "Google Source" etc. Perhaps we should leave this as-is then until we find out what to do with them. Let's keep in the "built-in" provider as you have it for now 👍🏼

src/plus/integrations/authentication/integrationAuthentication.ts
```diff
@@ -37,107 +40,262 @@ export interface IntegrationAuthenticationSessionDescriptor {
...
+	): Promise<ProviderAuthenticationSession | undefined> {
+		const existingSession = await this.getBuiltInExistingSession(descriptor);
+		if (existingSession != null) return existingSession;
+
+		const cloudIntegrations = await this.container.cloudIntegrations;
+		if (cloudIntegrations == null) return undefined;
+
+		let session = await cloudIntegrations.getConnectionSession(this.authProviderId);
+
+		if (session != null && session.expiresIn < 60) {
```
axosoft-ramint, line 151: We will likely need to make an exception for GitHub here since it comes back with `expiresIn` of 0 but doesn't really expire.

src/plus/integrations/authentication/integrationAuthentication.ts
```diff
@@ -37,107 +40,262 @@ export interface IntegrationAuthenticationSessionDescriptor {
...
-			return storedSession as ProviderAuthenticationSession | undefined;
+		return {
+			id: this.getSessionId(descriptor),
+			accessToken: session.accessToken,
+			scopes: descriptor?.scopes ?? [],
+			account: {
+				id: '',
+				label: '',
+			},
+			expiresAt: new Date(session.expiresIn * 1000 + Date.now()),
```
axosoft-ramint, line 198: Maybe we don't want to include `expiresAt` if the id is github since api is returning 0 (even though it doesn't expire)

src/plus/integrations/authentication/integrationAuthentication.ts
```diff
@@ -154,42 +312,43 @@ export class IntegrationAuthenticationService implements Disposable {
...
 						await import(/* webpackChunkName: "integrations" */ './github')
-					).GitHubEnterpriseAuthenticationProvider();
+					).GitHubEnterpriseAuthenticationProvider(this.container);
 					break;
 				case HostingIntegrationId.GitLab:
 				case SelfHostedIntegrationId.GitLabSelfHosted:
 					provider = new (
 						await import(/* webpackChunkName: "integrations" */ './gitlab')
-					).GitLabAuthenticationProvider();
+					).GitLabAuthenticationProvider(this.container, providerId);
```
axosoft-ramint, line 343: We should no longer need to pass `providerId` here.

src/plus/integrations/authentication/integrationAuthentication.ts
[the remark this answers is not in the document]
axosoft-ramint: Yes, thanks!

src/plus/integrations/authentication/github.ts
```diff
@@ -1,16 +1,47 @@
...
-		return descriptor?.domain ?? '';
+export class GitHubAuthenticationProvider extends CloudIntegrationAuthenticationProvider<HostingIntegrationId.GitHub> {
+	protected override get authProviderId(): HostingIntegrationId.GitHub {
+		return HostingIntegrationId.GitHub;
 	}
 
-	async createSession(
+	override async getBuiltInExistingSession(
+		descriptor?: IntegrationAuthenticationSessionDescriptor,
+		options?: { authorizeIfNeeded?: boolean; createIfNeeded?: boolean; forceNewSession?: boolean },
```
axosoft-ramint, line 18: Because this logic is serving only to get an existing session, not to create one, we should not need any of these options. On the `authentication.getSession` call, we can set `silent: true` and not send any other options.

src/plus/integrations/authentication/integrationAuthentication.ts
sergeibbb: true. Good point. I thought about it, but at the last moment have forgotten, because eventually it worked (by refreshing the session on every "createSession" call :-)

src/plus/integrations/authentication/integrationAuthentication.ts
sergeibbb: After GITCL-555 yes. But now it can be GitLabSelfHosted or GitLab

src/plus/integrations/authentication/github.ts
sergeibbb: ok

src/plus/integrations/authentication/integrationAuthentication.ts
axosoft-ramint: Ah I see, good call.

FTOSevenK (approved):
`
``
``
```

File: src/plus/integrations/integration.ts (hunk 1 above)
Comment (sergeibbb) [reply src/plus/integrations/integration.ts:300]: @axosoft-ramint 

My change about it looks like this. Is it what you mean?
image

`manageCloudIntegrations()` doesn't look as we want to launch here. Because it navigates us to "Manage integrations" web page. However, if I already have a GitHub integration connected I don't need to visit that page.
image

I'm thinking about creating a similar command, e.g. `connectCloudIntegration()` that would skip some steps if the connection already exists. I don't want to skip these steps in "manage", because the name says that we want to open the manage page intentionally. Probably, I could split the `manageCloudIntegrations()` to sub-functions and combine them differently in `manageCloudIntegrations` and `connectCloudIntegration`.

Another way that i see is to customize the "manage" call with some options such as `skipOpeningAManageScreenIfItsConnected: true`

Thoughts?

File: src/plus/integrations/authentication/integrationAuthentication.ts (hunk 4 above)
Comment (sergeibbb) [reply src/plus/integrations/authentication/integrationAuthentication.ts:151]: @axosoft-ramint 

Do we want to report it to the API team? They definitely should return different values for expired and never-expires tokens.

Comment (sergeibbb) [reply src/plus/integrations/authentication/integrationAuthentication.ts:156]: btw. done
image

File: src/plus/integrations/authentication/integrationAuthentication.ts (hunk 5 above)
Comment (sergeibbb) [reply src/plus/integrations/authentication/integrationAuthentication.ts]: What is better: to set `undefined` or to set a big value, such as 1 week or 1 year?

File: src/plus/integrations/authentication/github.ts (hunk 7 above)
Comment (sergeibbb) [reply src/plus/integrations/authentication/github.ts]: done

image

File: src/plus/integrations/authentication/integrationAuthentication.ts (hunk 5 above)
Comment (axosoft-ramint) [reply src/plus/integrations/authentication/integrationAuthentication.ts]: Actually, that brings up a good point - `expiresAt` on a session right now is the only thing distinguishing cloud token sessions from local ones - perhaps we just set a 1-year `expiresAt` on the tokens for now 👍🏼

File: src/plus/integrations/authentication/integrationAuthentication.ts [hunk 8]
```diff
@@ -37,107 +40,269 @@ export interface IntegrationAuthenticationSessionDescriptor {
...
+	) {
+		super(container);
+	}
+
+	protected override createSession(): Promise<ProviderAuthenticationSession | undefined> {
+		throw new Error('Method `createSession` should never be used in BuiltInAuthenticationProvider');
+	}
+
+	@debug()
+	override async getSession(
```
axosoft-ramint, line 266: Previously, this logic used to be in `getSession` of the authentication service, which would work for GitHub because it would fail the case of `this.supports(integrationId)`.

But now, when we call the new `get` from the service, we get the new GitHub authentication provider. When you call `getSession` with that provider, it uses the base `getSession`, which does not check for a built-in token. The only place the GitHub authentication provider does a built-in check is in `createSession`, which will not be reached when, for example, a remote is checking for a connected integration, or Launchpad is checking for existing GH connection.

Therefore, for the GitHub auth provider, we also need an override of `getSession` that does this check first (again, with only `silent: true` as in the `createSession` case, and wrapped as it is there) and if it finds it there, use that. Otherwise, call `super.getSession(descriptor, options)`.

axosoft-ramint (changes requested):
Currently, when testing with a GitHub repo on this PR branch, it shows that my remotes are not connected, and Launchpad also doesn't recognize my existing (VScode built-in) connected GitHub integration.

I did a bit of investigation and found out why - we covered existing built-in GitHub tokens in the `createSession` flow, but not in the `getSession` flow.

FTOSevenK (approved):

src/plus/integrations/authentication/integrationAuthentication.ts
sergeibbb: Ah. Clear! Thank you, it would take a lot of time for me to find this out.

src/plus/integrations/authentication/integrationAuthentication.ts
sergeibbb: done
image

src/plus/integrations/authentication/integrationAuthentication.ts
sergeibbb: @axosoft-ramint 
I think, this should be a fix: https://github.com/gitkraken/vscode-gitlens/commit/d47538e60a0a5b846cdc5eb78a1d08208afc7afc

Anyway. I need steps to reproduce, because now I cannot do it. When I have a built-in GitHub connection it never reaches my code to check it. 

Hmm. probably I can try to disable-enable the plugin, or delete-install it.

axosoft-ramint:
Hey @sergeibbb, while reviewing your PR, I'd suggest the following code changes:

👉 [**Fixes integration.connect throwing error for GitHub**](https://staging.gitkraken.dev/drafts/dcc731bd-c56c-4a42-ba1b-c8d9ede46d69?source=PRComment)
>Ensures that manageCloudIntegrations is always called before attempting integration.connect for GitHub

You can also review and apply these suggestions [locally on your machine](https://staging.gitkraken.dev/link/drafts/dcc731bd-c56c-4a42-ba1b-c8d9ede46d69?type=suggested_pr_change&prEntityId=WyJnaXRodWIiLCJwciIsIjEiLCIiLCJQUl9rd0RPQS1NUnVNNXkxQnllIl0=&source=PRComment).

<details><summary>Learn more about GitKraken Code Suggest</summary>

>[Code Suggest](https://staging.gitkraken.dev/redirect/aHR0cHM6Ly93d3cuZ2l0a3Jha2VuLmNvbS9zb2x1dGlvbnMvY29kZS1zdWdnZXN0P3NvdXJjZT1naXRodWIuY29tJm1lZGl1bT1wcl9jb21tZW50?source=PRSuggest&event=redirectToCodeSuggestPage&provider=github) liberates your code reviews from GitHub's restrictive, comment-only feedback style. As simple as suggesting changes in a Google-doc, provide real code suggestions from where you code, e.g. your IDE, and on anything in your project — not just on the lines of code changed in the PR.

_Join your team on [GitKraken](https://staging.gitkraken.dev/redirect/aHR0cHM6Ly93d3cuZ2l0a3Jha2VuLmNvbT9zb3VyY2U9Z2l0aHViLmNvbSZtZWRpdW09cHJfY29tbWVudA==?source=PRSuggest&event=redirectToIndexPage&provider=github) to speed up PR review._
</details>

axosoft-ramint:
@sergeibbb There are issues with `integration.connect` being called on a cloud integration before it is ready. `manageCloudIntegrations` should always be called before attempting the connection, or the user may not be logged in or may not see the settings page first.

I made some updates to address this - see my code suggestion above.

src/plus/integrations/authentication/integrationAuthentication.ts
sergeibbb: Well. I've debugged the steps of this case:

- ensured that it comes to `CloudIntegrationAuthenticationProvider.getSession()`
- where it calls for `GitHub...getBuiltInExistingSession()`
- so it never calls to the parent `...ProviderBase.getSession`
- therefore never tries to save the built-in GitHub credentials to the our secret storage.

As the result I see my origin as connected:
image


If I sign out the github:

image

It disconnects my origin immediately:
image

Now I'm checking that it doesn't restores the GitHub token from nowhere on re-connecting.

src/plus/integrations/integration.ts
sergeibbb: done

src/plus/integrations/authentication/integrationAuthentication.ts
sergeibbb: done

d13 (approved):
This is looking great. I've been able to successfully test:

- With local integration, without GK integration
- With GK integration, without local integration
- With local integration fallback, after signing out of GK integration

src/constants.ts
```diff
@@ -875,8 +876,16 @@ export type SupportedAIModels =
 	| `openai:${AIModels<'openai'>}`
 	| 'vscode';
 
+const supportedCloudIntegrationIds = ['github', 'jira'];
```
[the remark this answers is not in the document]
sergeibbb: Done.

I needed to move it to another file, otherwise it broke hundreds of imports of `constants.ts`

PR #3383, opened 2024-07-09
Prevents login flow from being stuck on cancel

### LINKED ISSUE #3395: Account login flow can get stuck when cancelling
### Description

There is a (fairly long-standing) bug with the login/register flow getting stuck when you cancel out of it.

To repro in 15.2, you can sign out of your GK account both in GitLens and on gkdev, then click "connect to Jira" in the Autolinks settings, and cancel the process partway through.

The "completion input fallback" promise never resolves or rejects or cancels, so it seems to just sit there (the same connect link doesn't work until you wait for the 1-minute it takes for that process to timeout, or you reload vscode).

The other annoying thing is that if you hit cancel early to stop GL from redirecting to gkdev in that flow, it still opens the completion input. It shouldn't do that, since you're not opening the URL at all.

### GitLens Version

15.2.0

### VS Code Version

Version: 1.91.1 (user setup)
Commit: f1e16e1e6214d7c44d078b1f0607b2388f29d729
Date: 2024-07-09T22:06:49.809Z
Electron: 29.4.0
ElectronBuildId: 9728852
Chromium: 122.0.6261.156
Node.js: 20.9.0
V8: 12.2.281.27-electron.0
OS: Windows_NT x64 10.0.22631

### Git Version

_No response_

_No response_

axosoft-ramint:
Fixes #3395

Two issues addressed here:
1. When initiating a login/register flow, if you hit cancel before opening gkdev, the completion input box still appears. It should not appear if you cancelled at this stage.
2. If the token input box appears, and then you hit `Escape` to cancel it, this should cancel the login flow. Right now, it does not cancel, and instead leaves it stuck, including processes that are waiting on it to complete.

To repro the second issue, log out of your account (both in GitLens and on gkdev), then click on the link under Autolink settings to "connect to Jira". Once you open the url, close out of it and then hide the code input with `Escape`. GitLens is now stuck waiting for that promise, so the next time you click the link it won't work, and if you try to reopen settings it will be blank.

### DIFF
--- a/src/plus/integrations/integrationService.ts
+++ b/src/plus/integrations/integrationService.ts
@@ -145,7 +145,9 @@ export class IntegrationService implements Disposable {
 
 		const account = (await this.container.subscription.getSubscription()).account;
 		if (account == null) {
-			if (!(await this.container.subscription.loginOrSignUp(true, source))) return;
+			if (!(await this.container.subscription.loginOrSignUp(true, source))) {
+				return;
+			}
 		}
 
 		try {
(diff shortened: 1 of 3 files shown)


src/plus/integrations/integrationService.ts
```diff
@@ -246,11 +246,9 @@ export class IntegrationService implements Disposable {
 				await openUrl(this.container.getGkDevExchangeUri(exchangeToken, `connect?${query}`).toString(true));
 			} catch (ex) {
 				Logger.error(ex, scope);
-				await env.openExternal(this.container.getGkDevUri('connect', query));
+				if (!(await env.openExternal(this.container.getGkDevUri('connect', query)))) return false;
 			}
-		} else {
-			await env.openExternal(this.container.getGkDevUri('connect', query));
-		}
+		} else if (!(await env.openExternal(this.container.getGkDevUri('connect', query)))) return false;
```
eamodio, line 251: Linting shouldn't allow this return here -- we shouldn't have single line returns outside of a single one line `if`

eamodio (approved):

eamodio (approved):
Looks good!

Issue #3367, opened 2024-06-27
Continually asked to reauthenticate


timheuer:
Type: <b>Bug</b>

At some point something changed but I'm continually asked to reauthenticate for `GitLens - Git supercharged` and I complete the authentication, but each new window reload or vscode restart the dialog pops up again as if the git auth is not correct, not honored, or something. 

![Image](https://github.com/user-attachments/assets/00db422c-48ed-4eff-a407-7ac29f59e489)

Debug log enabled:
```
[2024-06-27 16:19:21.871] [    1] GitLens (pre-release) v2024.6.2705  activating in Visual Studio Code - Insiders (1.91.0-insider) on the desktop; language='en', logLevel='debug', defaultDateLocale='null' (f8c254f156ae54d5752dcdab57c49dd1ad767f03d0ec75bd0211a98e15986ed3|03346b9b-bf64-42d1-9dbb-d13ad8f3ed001719505151207)
[2024-06-27 16:19:21.872] [    7] SubscriptionService.changeSubscription(e={"plan":{"actual":{"id":"free","name":"GitKraken","bundle":false,"cancelled":false,"trialReactivationCount":0,"startedOn":"2023-12-10T18:11:51.547Z"},"effective":{"id":"free","name":"GitKraken","bundle":false,"cancelled":false,"trialReactivationCount":0,"startedOn":"2023-12-10T18:11:51.547Z"}},"state":2,"previewTrial":{"startedOn":"2023-12-10T18:12:14.872Z","expiresOn":"2023-12-14T07:59:59.999Z"}}, t={"silent":true})
[2024-06-27 16:19:21.872] [    7] SubscriptionService.changeSubscription(e={"plan":{"actual":{"id":"free","name":"GitKraken","bundle":false,"cancelled":false,"trialReactivationCount":0,"startedOn":"2023-12-10T18:11:51.547Z"},"effective":{"id":"free","name":"GitKraken","bundle":false,"cancelled":false,"trialReactivationCount":0,"startedOn":"2023-12-10T18:11:51.547Z"}},"state":2,"previewTrial":{"startedOn":"2023-12-10T18:12:14.872Z","expiresOn":"2023-12-14T07:59:59.999Z"}}, t={"silent":true}) completed [0ms]
[2024-06-27 16:19:21.872] [    9] GitProviderService.setEnabledContext(e=false)
[...]

PR #3397, opened 2024-07-15
Supress auth error notifications for github triggered by Launchpad

axosoft-ramint:
Silences error notifications from popping up from the Launchpad indicator's automatic "get my pull requests" query.

This is a workaround until we find a better way to present this issue to users when Launchpad runs into it.

Partially addresses #3367


eamodio (approved):
Looks good, but we will ultimately want to improve how we would handle auth errors in launchpad.

axosoft-ramint:
> Looks good, but we will ultimately want to improve how we would handle auth errors in launchpad.

Yep, this is definitely a temp solution because ideally we surface this issue somewhere in the Launchpad indicator or Launchpad itself so users can correct if it they want. That needs some UX/design discussion, but this should stop the annoying notifications in the meantime.

PR #3408, opened 2024-07-18
Adds a jump to ref button to the graph

eamodio:
image

Click jumps to HEAD, Alt+Click opens a ref picker to jump to the chosen ref.

### DIFF
--- a/src/git/models/reference.ts
+++ b/src/git/models/reference.ts
@@ -104,6 +104,7 @@ export interface GitBranchReference {
 	id?: string;
 	name: string;
 	ref: string;
+	sha?: string;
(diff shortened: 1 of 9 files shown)


axosoft-ramint (approved):
Works well, though I wish we had a better icon for it.

d13 (approved):

eamodio:
@axosoft-ramint any thoughts to what icon might work better?

axosoft-ramint:
> @axosoft-ramint any thoughts to what icon might work better?

@eamodio I can't find a better one in the codicons library so it would have to be a custom one. Probably not worth the effort at this point.

PR #3412, opened 2024-07-19
Stores and continues deep link flow when switch opens new window

axosoft-ramint:
Completes:  GLVSC-595

When we use "switch to" or "open worktree in new window" on a Launchpad item, and that action opens a new window (as a result of the "switch" operation), currently the last remaining part (open the inspect view, in review mode if appropriate) doesn't happen because the new window loses the deep link context.

This is tricky to fix, because the deep link service is not aware of the user's decision while it is awaiting the completion of the 'switch' command.

This solution addresses the issue as follows:
1. Expands the "stored deep link" to include a state representing where the link left off. Optional and if not provided, default is "repo opening".
2. Whenever the deep link needs to use the switch command, it first stores a pending deep link *just in case* a new window will be opened as a result of the switch, and includes the current state ("switch to ref") in the stored link.
3. It then performs the switch. It detects if a new window was opened by checking the "current branch" name from before the switch command run to after. If the current branch wasn't changed, either the switch was canceled or the switch continued in a new window, so it resolves itself.
4. The "process pending deep link" logic was expanded to account for this state. Its repo matching was also expanded because we can't rely on repo path to find the right repo if we opened a worktree, which is now in a different folder.
5. At the new "open inspect" step we always clear the stored deep link because if the link got here, either we already processed the stored link or we did not use it (the user successfully completed the switch in the current context).

I also snuck in some extra cases where cancelling the deep link wasn't being handled in the state transition table.

Unfortunately, this one can't be tested in a debug window. To test "new window" flows, I generate a VSIX using `yarn package', install it and test with it.

### DIFF
--- a/src/uris/deepLinks/deepLinkService.ts
+++ b/src/uris/deepLinks/deepLinkService.ts
@@ -10,6 +10,7 @@ import { getBranchNameWithoutRemote } from '../../git/models/branch';
 import type { GitCommit } from '../../git/models/commit';
 import type { GitReference } from '../../git/models/reference';
 import { createReference, isSha } from '../../git/models/reference';
+import { Repository } from '../../git/models/repository';
 import type { GitTag } from '../../git/models/tag';
 import { parseGitRemoteUrl } from '../../git/parsers/remoteParser';
 import type { RepositoryIdentity } from '../../gk/models/repositoryIdentities';
@@ -94,6 +95,7 @@ export class DeepLinkService implements Disposable {
 			repoOpenLocation: undefined,
 			repoOpenUri: undefined,
 			params: undefined,
+			currentBranch: undefined,
 		};
 	}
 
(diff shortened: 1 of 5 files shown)


src/uris/deepLinks/deepLinkService.ts
```diff
@@ -1268,6 +1312,16 @@ export class DeepLinkService implements Disposable {
 							break;
 						}
 
+						// Storing link info in case the switch causes a new window to open
+						await this.container.storage.store('deepLinks:pending', {
```
axosoft-ramint, line 1316: Note: this part is definitely iffy. One issue with storing it like this before knowing if the user will open a new window or not, is that the user can 1. open a new window to complete the switch process, or 2. cancel the switch process altogether. In the case of 2, we need to clear out the stored deep link, but the service currently doesn't distinguish between the two, so the pending link doesn't get cleared out properly. In that case, the next time you open a window, it tries to continue the link by opening GitLens inspect, and gives an error if the repo wasn't found.

If we have a way of distinguishing the result of the "switch" here, we can instead use that result to determine if we need to store the pending link at all, and guarantee that we don't store it if the user did not choose to complete the switch in a new window.

axosoft-ramint: Looks like this is the best way for now, as eventing from the quickpick step is way trickier and hackier

eamodio:
@axosoft-ramint Since we know the deeplink is for a specific repo, can't we key them by the repo, so we will at least only process them when that particular repo is opened?

src/uris/deepLinks/deepLinkService.ts
eamodio: @axosoft-ramint take a look at:

https://github.com/gitkraken/vscode-gitlens/blob/3e248d338e1025e0054e7d6bcdebf5302b3b3fb0/src/git/actions/worktree.ts#L11-L39

I added this crude "result" completion that could be passed into the `worktree` command -- we could do something similar for `switch`. It could also be the inverse -- pass in a callback to get run or something like that.

axosoft-ramint:
> @axosoft-ramint Since we know the deeplink is for a specific repo, can't we key them by the repo, so we will at least only process them when that particular repo is opened?

I plan to include this part in separate cleanup work where I will improve deep link repo location/detection in general to be more in line with what we do for the launchpad view. Going to centralize that into the repo identity service and have our cloud workspaces, deep link, etc. services all route through it.

axosoft-ramint:
@eamodio Passed a callback through instead of always storing the deep link. Now it only stores if the user chose "current window" or "new window".

[Sends storage as callback to worktree command](https://github.com/gitkraken/vscode-gitlens/pull/3412/commits/96cedc2234eace5d87d20757505df4fac0ef5ee0)

eamodio (approved):
Looks good to me!

PR #3420, opened 2024-07-24
Check subscription even if users cannot check in (GLVSC-593)

sergiolms:
# Description

With this little change, we will validate the current subscription even if check-in fails or user is offline.
I tested this forcing a free account that expired yesterday into an offline client. In this case, GitLens checks if the stored license is valid even if it cannot check-in online to retrieve the latest information about the user's license.

# Checklist

### DIFF
--- a/src/plus/gk/account/subscriptionService.ts
+++ b/src/plus/gk/account/subscriptionService.ts
@@ -874,6 +874,9 @@ export class SubscriptionService implements Disposable {
 
 			Logger.error(ex, scope);
 			debugger;
+
+			// If we cannot check in, validate stored subscription
+			this.changeSubscription(this._subscription);
 			if (ex instanceof AccountValidationError) throw ex;
 
 			throw new AccountValidationError('Unable to validate account', ex);


src/plus/gk/account/subscriptionService.ts
```diff
@@ -874,6 +874,9 @@ export class SubscriptionService implements Disposable {
 
 			Logger.error(ex, scope);
 			debugger;
+
+			// If we cannot check in, validate stored subscription
+			this.changeSubscription(this._subscription);
```
axosoft-ramint, line 879: Correct me if I'm wrong but, shouldn't you spread `this._subscription` in as a copy since otherwise, it will be mutated, and the `matches` boolean in the function would always be `true`, preventing the subscription from properly being updated?

sergiolms: I thought [this commit](https://github.com/gitkraken/vscode-gitlens/commit/faea28a390df2592f822cd7a0a211b299af486ee#diff-e821a438849eca64d4d0dc885d4fd3c752fcb059b657c75bbf3fe53b6ae939f3) would prevent just that.

Besides, I have only seen usages with the spread operator when overwriting values:
Captura de pantalla 2024-07-26 a las 11 02 39
Captura de pantalla 2024-07-26 a las 11 02 54
Captura de pantalla 2024-07-26 a las 11 03 08
Captura de pantalla 2024-07-26 a las 11 04 11

axosoft-ramint (approved):
Not sure what cases to test, but at least for the following:

1. Sign out of GK account, and plus features should immediately go back to gated state

It works as expected. Thanks!

PR #3432, opened 2024-07-29
bugfix: Column resizing gets stuck when your mouse leaves the graph (GLVSC-610)

sergiolms:
# Description

The drag event would get stuck if you kept dragging the resize handle outside the graph.
Creating a `mouseup` event when leaving the graph zone will release the resize event.
This behavior simulates what GKD does.

Before the fix:

https://github.com/user-attachments/assets/8716f315-5670-412d-9eed-f20280c7b049

After the fix:

https://github.com/user-attachments/assets/2aa228b5-8de3-4d64-ad03-0b7dc8f4c983

GKD's behavior:

https://github.com/user-attachments/assets/5a69d258-dfd5-4ff0-9505-a40096d48ee1

# Checklist

### DIFF
--- a/src/webviews/apps/plus/graph/GraphWrapper.tsx
+++ b/src/webviews/apps/plus/graph/GraphWrapper.tsx
@@ -477,6 +477,14 @@ export function GraphWrapper({
 
 	const handleOnGraphMouseLeave = (_event: React.MouseEvent<any>) => {
 		minimap.current?.unselect(undefined, true);
+
+		document.dispatchEvent(
+			new MouseEvent('mouseup', {
+				view: window,
+				bubbles: true,
+				cancelable: false,
+			}),
+		);
 	};
 
 	const handleOnGraphRowHovered = (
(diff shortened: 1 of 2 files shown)


axosoft-ramint (approved):
Definitely an improvement. One question: is it possible to make the drag marker (the blue vertical line representing the drag anchor) go away as well once the drag leaves the view? Currently it stays on until your mouse re-enters.

drag anchor

If not, this is fine to merge as-is. Would just be a further improvement.

sergiolms:
> Definitely an improvement. One question: is it possible to make the drag marker (the blue vertical line representing the drag anchor) go away as well once the drag leaves the view? Currently it stays on until your mouse re-enters.

Yeah good catch! So I tried to fix this, but it's a bit tricky because it's related to the `:active` pseudo-class that the handle has while the user has the mouse down. I haven't figured out a way to completely remove it, but at least if the user isn't on the graph, we won't display it. The only corner case that's left is if the user holds the mouse down, leaves the graph, then comes back without releasing it. In that case we would have cancelled the action, but the handle remains highlighted until the mouse is released.


https://github.com/user-attachments/assets/ee717329-0e3a-42d2-b105-f0509edc8e77

axosoft-ramint (approved):

PR #3451, opened 2024-08-06
Uses custom fetch for shared provider api requests

axosoft-ramint:
Fixes GitLab items failing to load into Launchpad, and possibly fixes issues with Jira integration as well.


d13 (approved):
LGTM

PR #3453, opened 2024-08-08
Adds new cloud integration connection flow and applies to Launchpad

axosoft-ramint:
Adds support for using the new `/connect` page to connect cloud integrations.

Applies to the Launchpad "connect integrations" step when cloud integrations setting is enabled.


sergeibbb (commented):
👍

PR #3458, opened 2024-08-08
Adds "smart" branches visibility

eamodio:
Replaces the branches filter toggle with a top-line dropdown and adds a new "smart" filter mode.

The "smart" mode which includes the current branch, the current branch's upstream (if any), and either the base branch or if current branch is associated with a PR then the PR's target branch.

Also adds a `gitlens.graph.branchesVisibility` setting to control the default behavior, though once you can it via the dropdown the change is remembered per-repo.

image

While I think this is in a reasonable state to merge, it can still be improved in future follow-ups:
 - [x] ~Add better caching of the computed refs~
 - [x] ~Improve "base" branch detection when there is no PR involved (dig into the reflog)~

### DIFF
--- a/package.json
+++ b/package.json
@@ -932,6 +932,23 @@
 				"title": "Commit Graph",
 				"order": 50,
 				"properties": {
+					"gitlens.graph.branchesVisibility": {
+						"type": "string",
+						"enum": [
+							"all",
+							"smart",
+							"current"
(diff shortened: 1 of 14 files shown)


eamodio:
I've pushed some more changes that address the "future" follow-ups

PR #3463, opened 2024-08-09
Improves smart base/target branch detection

eamodio:
This is an improvement to the smart branch base/target detection. We now store target vs base separately, and favor target over base when available


axosoft-ramint (approved):
Looks like it's doing the job well

PR #3465, opened 2024-08-12
Don't hit GKDev API for a cloud session if there is no reason for it (GLVSC-624)

sergeibbb:
# Description

Stores a special value when it's know that GKDev does not have cloud integration. When we see this value, we don't try to hit the API until it's requested by an event.

# Checklist

### DIFF
--- a/src/plus/integrations/authentication/integrationAuthentication.ts
+++ b/src/plus/integrations/authentication/integrationAuthentication.ts
@@ -18,6 +18,7 @@ import {
 import type { ProviderAuthenticationSession } from './models';
 import { isSupportedCloudIntegrationId } from './models';
 
+type StoredNoSession = 'no-session-special-value';
 interface StoredSession {
 	id: string;
 	accessToken: string;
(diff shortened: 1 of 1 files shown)


axosoft-ramint:
Since this is introducing more complexity to an already complex session handling system, I'd like to review/understand the reason why we need to reach out to the API in the first place when fetching a session.

This was the previous system when Jira was the only cloud integration:
-`syncCloudIntegrations` was the only function that would handle interacting with the backend api for cloud token information, and it would automatically connect/disconnect the integrations based on their values. It would only be called either automatically or when certain actions occur like account login/change, "manage cloud integrations", or explicitly connecting a cloud integration like Jira.
- After that, the auth provider would store the token locally as a session and use only the local session rather than contacting the api (except for session expiration, in which it would call the api to refresh and get the new token).

Since then, at some point we allowed the integration auth provider to reach out to the api via this new `fetchSession` route.

So I can understand first: why did we make this change? Why is it not enough to let `syncCloudIntegrations` continue to handle the contact to the backend for cloud token data?

sergeibbb:
@axosoft-ramint 

> syncCloudIntegrations was the only function that would handle interacting with the backend api for cloud token information, 

It's not true anymore. Now GitHubGitPvoider is another source of interacting with authentication providers.

> why did we make this change?

 We did this change to cover the following case:
 
 - GL uses VS Code GitHub session
 - VS Code GitHub session disappears
 - GitHub Auth Provider is called for session from the GitHubGitProvider (with createIfNeeded=false)
 - We get nothing from VS Code so we go deep to GK Cloud integrations.
 - But since `createIfNeeded=false` we didn't reach the API.
 
We discussed it in the voice call dedicated to GITCL-589 with you and Eric on Aug,5. We decided that reaching the API should not be treated as "creating" and we can try to get the session from API whenever we don't have the local session.

The alternative was to have special solution for GitHub. When I didn't change the getSession, but added flags to cover GitHub's case. It was declined.

PR #3494, opened 2024-08-27
Add fallback/cutoff to our backend calls similar to how we handle GitHub queries #3519

sergeibbb:
# Description
`handleException` and `requestExceptionCount` etc. We should stop calling our APIs in a session if there’s a problem and X attempts in a row fail.

# Checklist


src/plus/gk/serverConnection.ts
```diff
@@ -21,8 +16,30 @@ interface GKFetchOptions extends FetchOptions {
...
+		private subscriptionService: SubscriptionService,
+	) {}
+
+	trackRequestException(): void {
+		this.subscriptionService.trackRequestException();
+	}
+
+	resetRequestExceptionCount(): void {
+		this.subscriptionService.resetRequestExceptionCount();
+	}
```
axosoft-ramint, line 35: Could these functions, and properties like `requestExceptionCount` and `requestsAreBlocked` be defined here rather than in the subscription service? If we want callers to be able to reset the exception count through this service, or to access the exception count through this service, we can define and expose those properties/functions here. Then for example, subscription service can check `this.connection.requestsAreBlocked` etc.

src/plus/gk/account/subscriptionService.ts
```diff
@@ -488,6 +491,37 @@ export class SubscriptionService implements Disposable {
...
+
+	resetRequestExceptionCount(): void {
+		this.requestExceptionCount = 0;
+		this.requestsAreBlocked = false;
+	}
+
+	trackRequestException(): void {
+		this.requestExceptionCount++;
+
+		if (this.requestExceptionCount >= 5 && this.getAuthenticationSession() !== null) {
```
axosoft-ramint, line 505: Why do we need to check that there is an active session here?

Would the following simplified version work?
```
if (this.requestExceptionCount >= 5 && !this.requestsAreBlocked) {
   void showIntegrationDisconnectedTooManyFailedRequestsWarningMessage('GitKraken');
   this.requestsAreBlocked = true;
   this.requestExceptionCount = 0;
}
```

src/plus/gk/account/subscriptionService.ts
```diff
@@ -488,6 +491,37 @@ export class SubscriptionService implements Disposable {
...
+		}
+	}
+
+	private fetchApi(
+		path: string,
+		init?: RequestInit,
+		options?: { token?: string; organizationId?: string },
+	): Promise<Response> {
+		if (this.requestsAreBlocked) {
+			throw new Error('Requests are blocked');
```
axosoft-ramint, line 520: We're throwing an error now in places where we weren't before. Should we be catching/handling this from the callers? And if not, have we verified that it doesn't break anything?

eamodio:
@sergeibbb At the high-level I'm confused by the need to have both `ServerConnection` (existing) and `FetchingService` (new), why did we need to introduce `FetchingService`? Why couldn't `ServerConnection` (or a holistic replacement of it) server the needs here?

For example, the `DraftsService` now has a reference to both, and uses them for different usages that feels like it make things more complex.

sergeibbb:
@eamodio 

> At the high-level I'm confused by the need to have both ServerConnection (existing) and FetchingService (new), why did we need to introduce FetchingService? Why couldn't ServerConnection (or a holistic replacement of it) serve our needs here?
>
> For example, the DraftsService now has a reference to both, and uses them for different usages that feels like it make things more complex.

Because `DraftsService` sends requests both to GK API and to S3. It seemed strange to block calls to S3 when GKDev is out.
So, Fetching service is responsible for making any types of requests to different services and the ServerConnection is responsible for making connection to the GK Dev server.

It was discussed here: https://gitkraken.atlassian.net/browse/GLVSC-625?focusedCommentId=123664

eamodio:
But those were already handled differently in ServerConnection, the "raw" fetch method allowed for more general fetching, and the other fetch versions covered more specific cases for the GK API.

Sure, having a "generic" fetch on ServerConnection was always a bit odd, since that class' intent was connecting to our backend, but today that was conceptually still true because the S3 calls are all in service of taking to our backend (just an implementation detail). At the same time, we could move that more generic fetch into a pure utility function. 

I'm also unsure of why we wouldn't encapsulate the connection/exception tracking within ServerConnection rather than have it in SubscriptionService. It feels like implementation details are leaking there. And if we change that, do we still need the Retriever?

sergeibbb:
@eamodio 

> that more generic fetch into a pure utility function.

Could be. But it needs a link to the container to build the `userAgent` value.

Anyway, I can join those two classes back and keep that little oddness.

>  why we wouldn't encapsulate the connection/exception tracking within ServerConnection rather than have it in SubscriptionService. 

because it's going to be reset with a new session, but the session is controlled by SubscriptionService. Anyway, it can be moved back to the ServerConnection.

sergeibbb:
@eamodio 

> And if we change that, do we still need the Retriever?

We only don't need the retriever, if we stop handling the request error inside of the generic fetch and move the handling to the `fetchApi`, that is GK specific.

src/plus/gk/account/subscriptionService.ts
sergeibbb: `fetchApi` is private inside of SubscriptionService and in all 3 places where it's called all exceptions are caught now.

But, we probably can handle it differently from other types of exceptions. For example, when `checkInAndValidateCore` catches and exception it changes subscription status. Maybe we shouldn't do it for this type of error?

src/plus/gk/account/subscriptionService.ts
sergeibbb: fixed

src/plus/gk/serverConnection.ts
sergeibbb: done

src/plus/gk/serverConnection.ts
```diff
@@ -169,28 +192,121 @@ export class ServerConnection implements Disposable {
...
 					...init,
 					headers: headers as HeadersInit,
 				},
 				options,
 			);
+			this.resetRequestExceptionCount();
+			return result;
 		} catch (ex) {
-			Logger.error(ex, scope);
+			this.handleRequestError('gitkraken', ex, scope);
```
axosoft-ramint, line 207: When we use node-fetch, in some cases, including with our backend, it is possible for the fetch call not to throw, but still get an error status. In other words, it is possible for the response to be a 429, and we never reach this `catch` block. In this case, we check `response.ok` and `response.status` to determine if an error occurred, and what code, since the `fetch` call itself will not throw. [Here is an example of this in the ai provider](https://github.com/gitkraken/vscode-gitlens/blob/main/src/ai/openaiProvider.ts#L200-L230).

We still want to track these, and we still want to stop the requests if they fail (i.e. `handleRequestError`) but we may need to be careful about throwing to the caller in these cases, as the caller may already have handling at the higher level. For example `draftsService` has `handleBadDraftResponse` and `cloudIntegrationService` still needs to send telemetry, etc.

sergeibbb: @axosoft-ramint
Should it also be fixed in GitHub and GitLab providers?

sergeibbb: @axosoft-ramint 

Something like this? https://github.com/gitkraken/vscode-gitlens/commit/49fc67e895a7f78291b3fc92b37dee5a90394caf
Here I'm wrapping an unsuccessful response to an error in order to handle it along with exceptions.

(and I've locked the error name to fight the type checks in the handler).

axosoft-ramint: @sergeibbb We can keep scope reduced to just our fetch calls from now, and update integration queries later if we need to.

The commit you linked seems to be handling it, although we should note now that these errors are being thrown to the caller where they were not before, so all the services that have handling for `!response.ok` from fetch results that route through `serverConnection` should now be updated to `try/catch` and we should ensure they are still handling the same errors appropriately in the `catch` blocks, and using `finally` as needed.

sergeibbb: @axosoft-ramint 

Well, I've completed the investigation. This change ends up in too many places:
image

Therefore I've changed my handling function. I let it detect timeout and rate limit responses and track them. Then it just exist and lets consumers to handle the result as they have been doing it before, so nothing is changed for them. Here is how I'm doing it: https://github.com/gitkraken/vscode-gitlens/commit/85bf10c7ea4333cc67b14a8e7a3d528a3d9124d2

src/plus/gk/serverConnection.ts
```diff
@@ -135,6 +155,9 @@ export class ServerConnection implements Disposable {
 	}
 
 	private async gkFetch(url: RequestInfo, init?: RequestInit, options?: GKFetchOptions): Promise<Response> {
+		if (this.requestsAreBlocked) {
+			throw new RequestsAreBlockedTemporarilyError();
```
axosoft-ramint, line 159: We will want to keep an eye on any pathways to `gkFetch` where the caller is not wrapping the call (i.e. `fetchApi` or `fetchGkDevApi`) in a try/catch block and handling a thrown error. For example, `cloudIntegrationService` has two such calls not wrapped in a try/catch block, and I'm not certain what will happen if these fail to catch errors (and it throws up the caller chain).

Not suggesting that there is an issue here - there may be existing cases where these calls throw. Just giving a heads-up that we're introducing one new possibility where a call returns with `rsp.ok` as `false` five times, triggers our temporary block, and then on that same call we start throwing errors (as we should). Hopefully these errors throwing up the chain does not cause issues.

src/plus/gk/serverConnection.ts
```diff
@@ -169,28 +192,180 @@ export class ServerConnection implements Disposable {
...
+		if (rsp.status >= 400 && rsp.status < 500) {
+			return;
+		}
+
+		if (Logger.isDebugging) {
+			void window.showErrorMessage(`GitKraken request failed: ${content} (${rsp.statusText})`);
+		}
+	}
+
+	private handleRequestError(
```
axosoft-ramint, line 284: Since this is currently for GK calls only (as the messaging and tracking indicate), we should name this something more specific, like `handleGkRequestError`, so that it is not accidentally applied in the future to fetch calls outside of the GK infrastructure.

Note: Same advice would apply for `handleUnsuccessfulResponse`.

src/plus/gk/serverConnection.ts
```diff
@@ -169,28 +192,180 @@ export class ServerConnection implements Disposable {
...
+		token: string | undefined,
+		ex: RequestError | (Error & { name: 'AbortError' }),
+		scope: LogScope | undefined,
+	): void {
+		if (ex instanceof CancellationError) throw ex;
+		if (ex.name === 'AbortError') throw new CancellationError(ex);
+
+		switch (ex.status) {
+			case 404: // Not found
+				throw new ProviderRequestNotFoundError(ex);
```
axosoft-ramint, line 294: The pattern "ProviderRequest....Error" that these are named after, refers to hosting providers or issue providers, like GitHub, GitLab, etc. Either we can name these "GitKrakenRequest...Error" or make the existing ones more generic like "Request...Error" and use that model for all of these. "ProviderRequest..." sounds integration-specific, which is not the case here.

src/plus/gk/serverConnection.ts
```diff
@@ -169,28 +192,180 @@ export class ServerConnection implements Disposable {
...
+				if (ex.message.includes('rate limit')) {
+					this.trackRequestException();
+					throw this.buildProviderRequestRateLimitError(token, ex);
+				}
+				throw new AuthenticationError('gitkraken', AuthenticationErrorReason.Forbidden, ex);
+			case 500: // Internal Server Error
+				Logger.error(ex, scope);
+				if (ex.response != null) {
+					this.trackRequestException();
+					void showIntegrationRequestFailed500WarningMessage(
```
axosoft-ramint, line 314: This function, and similar ones used in other cases, is specifically designed for integration messages. Its key tracking and other handling are specific to integration/provider requests. We need a more generic version of these "show message" functions that's scoped to GK requests.

src/plus/gk/account/subscriptionService.ts
```diff
@@ -1038,6 +1056,7 @@ export class SubscriptionService implements Disposable {
 		}
 
 		const session = await this._sessionPromise;
+		this.connection.resetRequestExceptionCount();
```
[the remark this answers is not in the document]
sergeibbb: @axosoft-ramint 
You're right. It's a wrong place. Actually, I wanted it be called when the session is changed successfully. Probably inside of `thig.getOrCreateSession()` is a better place for it.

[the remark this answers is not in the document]
sergeibbb: done

src/plus/gk/serverConnection.ts
sergeibbb: done

src/plus/gk/serverConnection.ts
sergeibbb: @axosoft-ramint 
Maybe wrap there? There not-ok-handling logs and sends telemetry events, then returns `undefined`. We can catch, log, and return `undefined`. (I suppose we don't need to telemetry here)

src/plus/gk/serverConnection.ts
sergeibbb: Done.
I've made it more generalized, because for now they don't contain anything provider specific and can be used for both services. We'll be able to introduce specific errors later, when/if they are truly needed.

https://github.com/gitkraken/vscode-gitlens/commit/1006ba3fc9059a8c2d295cf1ec4f029f4831a60f

src/plus/gk/serverConnection.ts
sergeibbb: @axosoft-ramint 

Question. So, these functions use the general method `showMessage(...)` and the difference is the `message` text itself and the `suppression-key` that lets us a specific message type.

So, do I understand right that I should keep using this `showMessage` function by providing different suppression keys. such as
```
showGkRequestFailed500WarningMessage(message) {
   showMessage('error', message, 'showGkRequestFailed500WarningMessage', ...)
}
```

So, I shouldn't create a different version of the general `showMessage()` function, should I?

sergeibbb: done. Is it what you mean?

https://github.com/gitkraken/vscode-gitlens/commit/49083abe2976b6f3e97d3d3150941c3ee83f5c6c

src/plus/gk/serverConnection.ts
axosoft-ramint: Yes, that sounds right. Can be done outside of this PR - perhaps we can write an issue for it.

src/plus/gk/serverConnection.ts
axosoft-ramint: @sergeibbb Yes that looks right, thanks!

sergeibbb: @axosoft-ramint 

Ok. So, now I'm testing it in its final state and then it can be considered for being merged.

src/plus/gk/account/subscriptionService.ts
axosoft-ramint: That's probably fine? Before these changes, it would hit this anyway if the fetch threw an error, so really the only new error we're introducing is from too many failed requests, in which case, we would probably want to hit that debugger and know about it anyway.

axosoft-ramint (approved):
Latest version looks good. I didn't really have time to exhaust all the possibilities but what I did simulate, it handles well. Thanks for working on it!

Let's wait to merge this until after our patch release, because it's targeting `main` and I would like this to sit in the prerelease for longer before we release it.

sergeibbb:
Hi @axosoft-ramint 
A release has happened. Can I merge?

axosoft-ramint:
> A release has happened. Can I merge?

@sergeibbb Thanks for the reminder! Merged it.

PR #3495, opened 2024-08-27
Adds worktree id to refs and worktree icons to the graph

axosoft-ramint:
Note: requires graph-side PR to merge first.

After that, will bump the library version in this PR and remove it from draft.


eamodio (approved):
Looks great!

PR #3497, opened 2024-08-28
Adds 'open in worktree' context menu command to branches and prs in views, and branches in graph

axosoft-ramint:
Adds the 'open in worktree' command to branches in views, prs in views (including launchpad view), and branches in the commit graph.

Note some restrictions:
- Is hidden for the current branch (attempting a switch there gives an error because it's your active worktree)
- Is hidden for branches that already have a worktree (there is already a command 'Open Worktree in New Window' that does the exact same thing)

### DIFF
--- a/package.json
+++ b/package.json
@@ -7415,6 +7415,12 @@
 				"icon": "$(empty-window)"
 			},
 			{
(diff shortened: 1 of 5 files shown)


d13 (approved):
LGTM

PR #3501, opened 2024-08-29
Fix connecting rich integrations on login to GK Dev (GLVSC-631)

sergeibbb:
# Description

When you logout from GK Dev and then login back, all configured cloud integrations should become connected.

# Checklist

### DIFF
--- a/src/plus/integrations/integration.ts
+++ b/src/plus/integrations/integration.ts
@@ -193,8 +193,9 @@ export abstract class IntegrationBase<
 		this._session = null;
 
 		if (connected) {
-			// Don't store the disconnected flag if this only for this current VS Code session (will be re-connected on next restart)
-			if (!options?.currentSessionOnly) {
+			// Don't store the disconnected flag if silently disconnecting or disconnecting this only for
+			// this current VS Code session (will be re-connected on next restart)
+			if (!options?.currentSessionOnly && !options?.silent) {
 				void this.container.storage.storeWorkspace(this.connectedKey, false);
 			}
 
@@ -261,7 +262,7 @@ export abstract class IntegrationBase<
 				await this.ensureSession({ createIfNeeded: forceSync });
 				break;
 			case 'disconnected':
-				await this.reset();
+				await this.disconnect({ silent: true });
 				break;
 		}
 	}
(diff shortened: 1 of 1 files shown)


src/plus/integrations/integrationService.ts
```diff
@@ -127,7 +127,7 @@ export class IntegrationService implements Disposable {
 	}
 
 	private onUserCheckedIn() {
-		void this.syncCloudIntegrations(false);
+		void this.syncCloudIntegrations(true);
```
axosoft-ramint, line 130: @sergeibbb This will cause check-ins to always force reconnection to providers, even when the user manually disconnects and disables them.

check-in is considered a passive sync case and should never force reconnections, because it happens automatically (in my repro, I can trigger it by refreshing the account view).

checkin-forces-reconnection

Instead, we should check the logout flow and ensure that integrations can be disconnected "softly" when a logout or some other passive sync case triggers the disconnect. Then those integrations would reconnect upon login even when `forceConnect` is `false` on `syncCloudIntegrations`.

axosoft-ramint (changes requested):

src/plus/integrations/integrationService.ts
sergeibbb: fixed

axosoft-ramint (changes requested):
The side-effect of this reset is similar to the previous issue - by clearing the "connected" key, we lose context on manual disconnects the user did, so any integrations the user manually disabled are automatically reconnected when logging back in (which is unexpected).

To repro:
- Go to remotes view and disconnect GitHub, then choose "disable and sign out"
- Log out of your GitKraken account in the client
- Log back in to your GitKraken account in the client
- Notice that GitHub is automatically reconnected (should still be disconnected)

I think it's fine to just use `this.disconnect` here, but we may need a new flag to use (not `currentSessionOnly`, because setting that causes `this.container.integrations.disconnected(this, this.key)` not to fire). The flag would just ensure that we do not store "false" for the connected key of this integration in storage.

axosoft-ramint:
@sergeibbb Hope you can forgive me committing to your branch. Since we want this in the release next week, I went ahead and updated the PR: it is back to using `disconnect({ silent: true })` with the added condition that using the `silent` flag causes us not to "remember" your disconnect, so that it automatically reconnects the next time you log in. Unlike `reset`, this does not clear any stored value you have for remembering disconnected state - it simply does not touch the storage.

axosoft-ramint (approved):

PR #3506, opened 2024-09-02
Adds a sidebar to the Commit Graph

eamodio:


src/webviews/apps/shared/app.ts
```diff
@@ -0,0 +1,114 @@
...
+import type { Disposable } from './events';
+import { HostIpc } from './ipc';
+
+export abstract class GlApp<
+	State extends { webviewId: CustomEditorIds | WebviewIds | WebviewViewIds; timestamp: number } = {
+		webviewId: CustomEditorIds | WebviewIds | WebviewViewIds;
+		timestamp: number;
+	},
+> extends GlElement {
+	@property({ type: String }) name!: string;
```
d13, line 20: Let's add `delegatesFocus` this since the App itself doesn't need to be focusable:

```typescript
	static override shadowRootOptions: ShadowRootInit = {
		...LitElement.shadowRootOptions,
		delegatesFocus: true,
	};
```

src/webviews/apps/shared/app.ts
```diff
@@ -0,0 +1,114 @@
...
+		this._sendWebviewFocusChangedCommandDebounced = debounce((params: WebviewFocusChangedParams) => {
+			this._ipc.sendCommand(WebviewFocusChangedCommand, params);
+		}, 150);
+
+		// Removes VS Code's default title attributes on <a> tags
+		document.querySelectorAll('a').forEach(a => {
+			if (a.href === a.title) {
+				a.removeAttribute('title');
+			}
+		});
```
d13, line 71: I wonder if we should just do this in `GlElement`? Though it could add up to a lot of DOM work.

src/webviews/apps/shared/app.ts
eamodio: Will do

src/webviews/apps/shared/app.ts
eamodio: Ugh, I hate to need every component to do this, and vscode only does it once on load

d13: Wondering if we even need this in Shadow DOM

PR #3509, opened 2024-09-02
Deleting a branch calls dialog of deleting its worktree (GLVSC-632)

sergeibbb:
# Description

Or instead, the Delete Worktree flow could just be triggered as an optional step in the Delete Branch flow.

# Checklist

### DIFF
--- a/src/commands/git/worktree.ts
+++ b/src/commands/git/worktree.ts
@@ -103,7 +103,7 @@ interface CreateState {
 	skipWorktreeConfirmations?: boolean;
 }
 
-type DeleteFlags = '--force' | '--delete-branches';
+type DeleteFlags = '--force' | '--delete-branches' | '--deleting-of-selected-branches';
 
 interface DeleteState {
 	subcommand: 'delete';
@@ -953,6 +953,12 @@ export class WorktreeGitCommand extends QuickCommand<State> {
 	private *deleteCommandConfirmStep(state: DeleteStepState, context: Context): StepResultGenerator<DeleteFlags[]> {
 		context.title = state.uris.length === 1 ? 'Delete Worktree' : 'Delete Worktrees';
 		const label = state.uris.length === 1 ? 'Delete Worktree' : 'Delete Worktrees';
+		const deletingOfSelectedBranches = state.flags.includes('--deleting-of-selected-branches');
+		const selectedBranchesLabelSuffix = !deletingOfSelectedBranches
+			? ''
+			: state.uris.length === 1
+			  ? ' of corresponding branch'
+			  : ' of corresponding branches';
 		const branchesLabel = state.uris.length === 1 ? 'Branch' : 'Branches';
 		const description =
 			state.uris.length === 1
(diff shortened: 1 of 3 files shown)


axosoft-ramint (commented):
Just commenting on the user experience, but when I choose to delete a branch, I find it a bit stange that it jumps me into this screen:

image

There is no indication here that "this is a branch with a worktree, so you need to delete the worktree first". It just says "delete worktree". So it feels like I clicked the wrong command, even though I did not.

I think we can do a better job with messaging to make this less confusing. Either the title of the "delete worktree" step can be updated, or the two available options can say "Delete Worktree & Branch" and "Force Delete Worktree & Branch", to make it clear that we are deleting the worktree and then the branch.

src/commands/git/worktree.ts
```diff
@@ -110,6 +110,7 @@ interface DeleteState {
 	repo: string | Repository;
 	uris: Uri[];
 	flags: DeleteFlags[];
+	doNotSuggestDeletingBranches?: boolean;
```
axosoft-ramint, line 113: Maybe we can just pass `branch` or `branches` as state when coming from "delete branch" and then we don't need to calculate them on the delete steps, and also can use their existence in state as a flag to restrict the options?

sergeibbb: @axosoft-ramint 

What do you think about this:
image

sergeibbb: @axosoft-ramint 

>  then we don't need to calculate them on the delete steps

It's not clear how we can use the passed branches on the "worktrees" step. We calculate branches only if "Delete Worktrees&Branches" is selected, but those are hidden.

axosoft-ramint: Fair enough - if the state isn't needed then a boolean should suffice 👍🏼

sergeibbb: @axosoft-ramint 

I'm thinking about renaming because it affects the view in a very specific case. "Deleting of the worktree of a selected branch". So, it can be called like this: `deletingOfSelectedBranches`, or moreover I'm thinking of replacing the field by a flag:

```
type DeleteFlags = '--force' | '--delete-branches' | '--deleting-of-selected-branches';
```

What do you think?

sergeibbb: @axosoft-ramint 

I've just pushed how it can be with a `--deleting-of-selected-branches` flag. Let me know if you prefers it being a boolean field.

sergeibbb: The UI:

image

image

Unfortunately, I need to go now. I can do any other updates tomorrow.

axosoft-ramint: @sergeibbb No preference here. Up to you. Doing it with a flag like that sounds fine. My main hope is that the messaging on the worktree step allows me to understand that we are deleting the worktree because it is necessary to delete the branch.

sergeibbb: @axosoft-ramint 

Ok. I'm leaving it with the flag. Feel free to suggest updates for the wording, so it would be more clearer for user. You are the native speaker, so, you know it better.

axosoft-ramint: @sergeibbb I think we can take out the word "corresponding" in the options since it becomes a bit wordy. We should also make sure the capitalization is consistent in the options.

What do you think about "Delete Worktree of Branch"/"Delete Worktrees of Branch" etc.?

axosoft-ramint: @sergeibbb I updated the wording on your branch. Thanks for working on this!

axosoft-ramint (approved):
LGTM

src/commands/git/worktree.ts
sergeibbb: Great! Thank you!

PR #3522, opened 2024-09-06
GLVSC-593: Check expiration of paid accounts offline

### LINKED ISSUE #3569: Check expiration of paid accounts even when user cannot check in (i.e. offline)

sergiolms:
# Description

Paid licenses can expire while offline, so with this PR we introduce some checks and a new `Paid expired` state to users that own an expired license, so we can customize messages and promotions for them if we want to.

Adding @justinrobots as CC to this PR because I'd like him to review the messages I set in case we want to change them or if I made a typo. 😄 

Here's a screenshot of how it would look like for users that have a pro/enterprise account that has expired:
Captura de pantalla 2024-09-06 a las 11 12 01

# Checklist

---
<a href="https://gitkraken.dev/review/github/gitkraken/vscode-gitlens/pull/3522">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://gitkraken.dev/images/figures/gitkraken-review-badge-dark.svg">
    Open with GitKraken
  </picture>
</a>

### DIFF
--- a/package.json
+++ b/package.json
@@ -17846,6 +17846,15 @@
 						"when": "gitlens:plus:state == 6"
 					},
 					{
+						"id": "pro-expired-reactivate",
+						"title": "Reactivate Pro Power-up",
+						"description": "Reactivate your Pro account and experience all the new [Pro features](https://gitkraken.com/gitlens/pro-features?utm_source=gitlens-extension&utm_medium=in-app-links) and the full [GitKraken DevEx platform](https://gitkraken.com/devex?utm_source=gitlens-extension&utm_medium=in-app-links).\n\n[Reactivate Pro](command:gitlens.plus.upgrade?%7B%22source%22%3A%22walkthrough%22%7D)\n\n**Pro Features**\n$(gitlens-graph)  [Commit Graph](command:gitlens.openWalkthrough?%7B%22step%22%3A%22visualize%22,%22source%22%3A%22walkthrough%22%7D) — visualize your repository and keep track of all work in progress\n$(rocket)  [Launchpad](command:gitlens.openWalkthrough?%7B%22step%22%3A%22launchpad%22,%22source%22%3A%22walkthrough%22%7D) — stay focused and keep your team unblocked\n$(gitlens-code-suggestion)  [Code Suggest](command:gitlens.openWalkthrough?%7B%22step%22%3A%22code-collab%22,%22source%22%3A%22walkthrough%22%7D) — free your code reviews from unnecessary restrictions\n$(gitlens-cloud-patch)  [Cloud Patches](command:gitlens.openWalkthrough?%7B%22step%22%3A%22code-collab%22,%22source%22%3A%22walkthrough%22%7D) — easily and securely share code with your teammates\n$(gitlens-worktrees-view)  **Worktrees** — work on multiple branches simultaneously\n$(gitlens-workspaces-view)  **Workspaces** — group and manage multiple repositories together\n$(graph-scatter)  [Visual File History](command:gitlens.openWalkthrough?%7B%22step%22%3A%22visualize%22,%22source%22%3A%22walkthrough%22%7D) — visualize the evolution of a file and quickly identify when the most impactful changes were made and by whom",
+						"media": {
+							"markdown": "walkthroughs/welcome/pro-reactivate.md"
+						},
+						"when": "gitlens:plus:state == 6"
+					},
+					{
 						"id": "visualize",
 						"title": "Visualize with Commit Graph & Visual File History",
 						"description": "**Commit Graph**\nEasily visualize your repository and keep track of all work in progress.\nUse the rich commit search to find exactly what you're looking for. Its powerful filters allow you to search by a specific commit, message, author, a changed file or files, or even a specific code change. [Learn more](https://gitkraken.com/solutions/commit-graph?utm_source=gitlens-extension&utm_medium=in-app-links)\n\n[Open Commit Graph](command:gitlens.showGraph)\n\n💡Quickly toggle the Graph via the [Toggle Commit Graph](command:gitlens.toggleGraph) command or maximize it using the [Toggle Maximized Commit Graph](command:gitlens.toggleMaximizedGraph) command.\n**Visual File History**\nVisualize the evolution of a file and quickly identify when the most impactful changes were made and by whom\n\n[Open Visual File History](command:gitlens.showTimelineView)",
--- a/package.json
+++ b/package.json
@@ -17548,7 +17548,7 @@
 			{
 				"view": "gitlens.views.worktrees",
 				"contents": "[Upgrade to Pro](command:gitlens.plus.upgrade?%7B%22source%22%3A%22worktrees%22%7D)",
-				"when": "gitlens:plus:required && gitlens:plus:state == 4"
+				"when": "gitlens:plus:required && (gitlens:plus:state == 4 || gitlens:plus:state == 7)"
 			},
 			{
 				"view": "gitlens.views.worktrees",
@@ -17558,12 +17558,12 @@
 			{
 				"view": "gitlens.views.worktrees",
 				"contents": "Launchpad sale: Save 75% or more on GitLens Pro",
-				"when": "gitlens:plus:required && gitlens:plus:state == 4 && gitlens:promo == launchpad-extended"
+				"when": "gitlens:plus:required && (gitlens:plus:state == 4 || gitlens:plus:state == 7) && gitlens:promo == launchpad-extended"
 			},
 			{
 				"view": "gitlens.views.worktrees",
 				"contents": "Limited-time sale: Save up to 80% on GitLen Pro",
-				"when": "gitlens:plus:required && gitlens:plus:state == 4 && gitlens:promo == devexdays"
+				"when": "gitlens:plus:required && (gitlens:plus:state == 4 ||  gitlens:plus:state == 7) && gitlens:promo == devexdays"
 			},
 			{
 				"view": "gitlens.views.worktrees",
@@ -17852,7 +17852,7 @@
 						"media": {
 							"markdown": "walkthroughs/welcome/pro-reactivate.md"
 						},
-						"when": "gitlens:plus:state == 6"
+						"when": "gitlens:plus:state == 7"
 					},
 					{
 						"id": "visualize",
--- a/src/plus/gk/account/promos.ts
+++ b/src/plus/gk/account/promos.ts
@@ -30,6 +30,7 @@ const promos: Promo[] = [
 			SubscriptionState.ProTrial,
 			SubscriptionState.ProTrialExpired,
 			SubscriptionState.ProTrialReactivationEligible,
+			SubscriptionState.PaidExpired,
 		],
 		startsOn: new Date('2024-09-27T06:59:00.000Z').getTime(),
 		expiresOn: new Date('2024-10-14T06:59:00.000Z').getTime(),
@@ -48,6 +49,7 @@ const promos: Promo[] = [
 			SubscriptionState.ProTrial,
 			SubscriptionState.ProTrialExpired,
 			SubscriptionState.ProTrialReactivationEligible,
+			SubscriptionState.PaidExpired,
 		],
 		startsOn: new Date('2024-10-13T06:59:00.000Z').getTime(),
 		expiresOn: new Date('2024-11-05T06:59:00.000Z').getTime(),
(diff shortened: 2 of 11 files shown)


sergiolms:
Note: Build is failing in `main`. I'm not sure of the real fix here, but it comes from something related to worktrees. In the meantime I added a [commit](https://github.com/gitkraken/vscode-gitlens/pull/3522/commits/613b2b99801e2e91c358a7e20935c229a2c0506b) to temporarily fix this error so it compiles, but feel free to remove it once it's fixed and I can rebase this branch. Or let me know and I can remove it myself.

src/webviews/apps/shared/components/feature-badge.ts
```diff
@@ -269,6 +269,13 @@ export class GlFeatureBadge extends LitElement {
 					${this.renderUpgradeActions(html`<p>Please upgrade for full access to Pro features:</p>`)}`;
 				break;
 
+			case SubscriptionState.PaidExpired:
+				content = html`<p>
+						Your Pro license as ended. You can now only use Pro features on publicly-hosted repos.
```
axosoft-ramint, line 274: > 
```suggestion
						Your Pro license has ended. You can now only use Pro features on publicly-hosted repos.
```

package.json
```diff
@@ -17880,6 +17880,15 @@
...
+						"media": {
+							"markdown": "walkthroughs/welcome/pro-reactivate.md"
+						},
+						"when": "gitlens:plus:state == 6"
```
axosoft-ramint, line 17890:
```suggestion
						"when": "gitlens:plus:state == 7"
```

You will want to do a search for `gitlens:plus:state` across `package.json` and find all cases where we need to account for the new state you have made (7). For example, you'll need a new addition for text content in worktrees view.

axosoft-ramint: In addition, this will cause a new value to be sent out in telemetry so we will want to make sure all consumers of GL telemetry are set up to handle and understand the new state, and the difference between "free" and "paidexpired" if the distinction is important.

ramin-t (commented):
Some initial things I noticed. Will test e2e next week (will take a bit to set up/mimic an account with a paid expired license).

src/webviews/apps/shared/components/feature-badge.ts
eamodio: I'd suggest "Your Pro license has expired..."

package.json
sergiolms: @axosoft-ramint 
> In addition, this will cause a new value to be sent out in telemetry so we will want to make sure all consumers of GL telemetry are set up to handle and understand the new state, and the difference between "free" and "paidexpired" if the distinction is important.

What changes do we need to do in order to have this? Do I need to update something else in the GitLens side or is this related exclusively to other projects consuming GL telemetry? Thanks!

axosoft-ramint: @sergiolms I don't think we need to change anything in the telemetry, just that we should communicate it so that the projects, filters, views, etc. that use telemetry are aware of the new state and can adjust appropriately.

package.json
```diff
@@ -17589,7 +17589,12 @@
 			{
 				"view": "gitlens.views.worktrees",
 				"contents": "DevEx Days 24 Sale: Save up to 80% on GitLens Pro - lowest price of the year!",
-				"when": "gitlens:plus:required && gitlens:plus:state == 4 && gitlens:promo == devexdays24"
+				"when": "gitlens:plus:required && (gitlens:plus:state == 4 || gitlens:plus:state == 7) && gitlens:promo == devexdays24"
```
axosoft-ramint, line 17592: I think we would want to apply `gitlens:plus:state == 7` case to the "pro 50% off" promo just above this as well.

src/plus/gk/account/promos.ts
```diff
@@ -21,6 +21,7 @@ const promos: Promo[] = [
 			SubscriptionState.FreePlusInTrial,
 			SubscriptionState.FreePlusTrialExpired,
 			SubscriptionState.FreePlusTrialReactivationEligible,
+			SubscriptionState.PaidExpired,
```
axosoft-ramint, line 24: We probably want to include this state on the `pro50` promo as well.

ramin-t (commented):
Looks good. I can only really simulate having an expired license in testing, but it showed the right messaging for me on Pro features and account view. Only suggestion would be to apply this state to our default pro50 promo as well.

package.json
sergiolms: I thought it was only for free accounts that converted to pro accounts for the first time. Since you come from an expired pro account, I thought it wouldn't make sense to make a promo on your first seat, as you're not new here. But I may be missing some information so you may be right :+1:

axosoft-ramint: Ah yes, good point! I hadn't thought it through. Then let's leave it as you had it, without the pro50 promo included. 👍🏼 Sorry for the unnecessary change.

src/plus/gk/account/promos.ts
axosoft-ramint: **Edit:** I was wrong - as you mentioned, the promo is only for first seat: https://github.com/gitkraken/vscode-gitlens/pull/3522#discussion_r1758517486

ramin-t (approved):

axosoft-ramint:
Heads-up: Just noticed after talking with @eamodio that we're currently filtering out expired paid licenses during the check-in computations: https://github.com/gitkraken/vscode-gitlens/blob/main/src/plus/gk/checkin.ts#L69-L71

This might mean that those paid licenses with expired date never make it into someone's license data, which means they would never enter the new state being set up here.

We might need to update those computations:
1. If the best license available is an expired paid license, we should probably let it through so that we can recognize this "expired paid" state rather than defaulting to "free", which may offer options like "start trial" that the user isn't eligible for yet.
2. But we should be careful not to have the paid, expired license data replace the trial data, in case someone does reactivate trial with an expired license: https://github.com/gitkraken/vscode-gitlens/blob/main/src/plus/gk/checkin.ts#L179-L181

sergiolms:
> Heads-up: Just noticed after talking with @eamodio that we're currently filtering out expired paid licenses during the check-in computations: https://github.com/gitkraken/vscode-gitlens/blob/main/src/plus/gk/checkin.ts#L69-L71
> 
> This might mean that those paid licenses with expired date never make it into someone's license data, which means they would never enter the new state being set up here.

That's right, they get filtered out when you log in. My changes only apply if you were already logged in and while offline your account expires.

> We might need to update those computations:
> 
> 1. If the best license available is an expired paid license, we should probably let it through so that we can recognize this "expired paid" state rather than defaulting to "free", which may offer options like "start trial" that the user isn't eligible for yet.
> 2. But we should be careful not to have the paid, expired license data replace the trial data, in case someone does reactivate trial with an expired license: https://github.com/gitkraken/vscode-gitlens/blob/main/src/plus/gk/checkin.ts#L179-L181

Gotcha, so instead of filtering out now we will take it into account and only replace the effective license if the actual license isn't a paid expired.

Good catch

src/plus/gk/checkin.ts
```diff
@@ -176,7 +180,18 @@ export function getSubscriptionFromCheckIn(
 		);
 	}
 
-	if (effective == null || getSubscriptionPlanPriority(actual.id) >= getSubscriptionPlanPriority(effective.id)) {
+	const isActualLicenseExpired =
```
axosoft-ramint, line 183: We could just check `latestStatus` and/or `expiresOn` date of the `actual` here, right? Then no need to go through `computeSubscriptionState`.

src/plus/gk/checkin.ts
```diff
@@ -66,9 +72,7 @@ export function getSubscriptionFromCheckIn(
 
 	let effectiveLicenses = Object.entries(data.licenses.effectiveLicenses) as [GKLicenseType, GKLicense][];
 	let paidLicenses = Object.entries(data.licenses.paidLicenses) as [GKLicenseType, GKLicense][];
-	paidLicenses = paidLicenses.filter(
-		license => license[1].latestStatus !== 'expired' && license[1].latestStatus !== 'cancelled',
-	);
+	paidLicenses = paidLicenses.filter(license => license[1].latestStatus !== 'cancelled');
```
axosoft-ramint, line 75: I could be wrong, but I *think* that `cancelled` is Chargebee's version of `expired`, so we may want to remove both of these conditions from the filter (i.e. get rid of this line entirely, as long as `computeSubscriptionState` can ultimately get us the right license state.

src/plus/gk/checkin.ts
```diff
@@ -176,7 +180,18 @@ export function getSubscriptionFromCheckIn(
...
+			plan: {
+				actual: actual,
+				effective: actual,
+			},
+			account: account,
+		}) === SubscriptionState.PaidExpired;
+	if (
+		effective == null ||
+		(getSubscriptionPlanPriority(actual.id) >= getSubscriptionPlanPriority(effective.id) && !isActualLicenseExpired)
+	) {
```
axosoft-ramint, line 194: One other issue I am seeing with `computeSubscriptionState` that could happen if we let expired subscriptions through:

1. You have a paid, but expired, Pro license.
2. You are on a (reactivated) Pro trial.

Note that this conditional is not triggered so `effective = { ...actual }` does not happen, but...

`effective` is still set in 171 to a trial license with `effective.id === SubscriptionPlanId.Pro`. And `actual` is an expired license with `actual.id === SubscriptionPlanId.Pro`.

So when we hit `computeSubscriptionState`, it will match the `actual.id === `effective.id` check (which previously wouldn't happen unless we hit this `effective = { ...actual }` line, and that would only happen if it was an active paid license since we were filtering out paid licenses that were expired.

Ultimately, this would cause your license to show as a paid Pro license rather than an active Pro trial.

To avoid that regression, I suggest we just attach an `isTrial` property to these and we can add that to the condition in `computeSubscriptionState`.

src/plus/gk/checkin.ts
sergiolms: Didn't know that, I'll remove that as well

src/plus/gk/checkin.ts
sergiolms: Updated with all feedback except this one

src/plus/gk/checkin.ts
```diff
@@ -176,6 +177,7 @@ export function getSubscriptionFromCheckIn(
 			new Date(license.latestEndDate),
 			license.latestStatus === 'cancelled',
 			license.nextOptInDate ?? data.nextOptInDate,
+			license.latestStatus === 'trial' || license.latestStatus === 'in_trial',
```
axosoft-ramint, line 180: This is *probably* fine, but I'm not 100% sure if we can rely on `latestStatus` coming through as 'trial' for trial licenses the way we are hoping for it to here. Since effective licenses from the api check-in response were always intended to be for trials, and paid licenses were always intended to be non-trial, we could probably just get away with using `true` in effective licenses and `false` for paid ones when setting them up from the check-in response.

sergiolms: hmmm, okay, I just uploaded a commit with this

axosoft-ramint:
Going to give this one more review/test pass today and we can merge if it looks good.

src/plus/gk/checkin.ts
```diff
@@ -176,10 +186,17 @@ export function getSubscriptionFromCheckIn(
...
+				latestEndDate > today &&
+				today > latestStartDate,
 		);
 	}
 
-	if (effective == null || getSubscriptionPlanPriority(actual.id) >= getSubscriptionPlanPriority(effective.id)) {
+	if (
+		effective == null ||
+		(getSubscriptionPlanPriority(actual.id) >= getSubscriptionPlanPriority(effective.id) &&
+			!isSubscriptionExpired(actual))
```
axosoft-ramint, line 198: I'm not sure how this passed the build check on the PR, but `isSubscriptionExpired` is expecting a different type than it is being sent here.

image

Maybe you can just check the expiration date directly on it since its a `SubscriptionPlan`?

ramin-t (changes requested):

src/plus/gk/checkin.ts
sergiolms: You're right, not sure why it didn't error before 🤔 
Fixed it, thanks for the catch!

axosoft-ramint:
When I am in the "expired paid" state, the account view is confusing:

image

There is no indication here that my license is expired. Furthermore, it provides links to sign in/sign up, when it should be asking me to renew my license.

Any chance you can take a pass at the content here for this state?

d13:
Let's pause on this PR until after the 16.0 release, so we can spend more time and focus on the priority 16.0 items. Going to bump this back to draft in the meantime.

d13:
Closing as stale

PR #3558, opened 2024-09-09
Graduates Launchpad out of preview

eamodio:
This still needs finalized messaging and promo testing.

### DIFF
--- a/src/constants.telemetry.ts
+++ b/src/constants.telemetry.ts
@@ -466,3 +466,5 @@ type SubscriptionEventData = {
 		Record<`previous.subscription.${string}`, string | number | boolean | undefined> &
 		Record<`previous.subscription.previewTrial.${string}`, string | number | boolean | undefined>
 >;
+
+export type TrackingContext = 'graph' | 'launchpad' | 'visual_file_history' | 'worktrees';
--- a/src/constants.telemetry.ts
+++ b/src/constants.telemetry.ts
@@ -467,4 +467,5 @@ type SubscriptionEventData = {
 		Record<`previous.subscription.previewTrial.${string}`, string | number | boolean | undefined>
 >;
 
+/** Used to provide a "source context" to gk.dev for both tracking and customization purposes */
 export type TrackingContext = 'graph' | 'launchpad' | 'visual_file_history' | 'worktrees';
--- a/src/plus/focus/focus.ts
+++ b/src/plus/focus/focus.ts
@@ -1,5 +1,5 @@
 import type { QuickInputButton, QuickPick, QuickPickItem } from 'vscode';
-import { commands, Uri } from 'vscode';
+import { commands, ThemeIcon, Uri } from 'vscode';
 import { getAvatarUri } from '../../avatars';
 import type {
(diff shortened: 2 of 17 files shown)


src/plus/focus/focus.ts
```diff
@@ -258,6 +263,11 @@ export class FocusCommand extends QuickCommand<State> {
 				newlyConnected = Boolean(connected);
 			}
 
+			const result = yield* ensureAccessStep(state, context, PlusFeatures.Focus);
+			if (result === StepResultBreak) continue;
+
+			context.showGraduationPromo = getApplicablePromo(result.subscription.current.state, 'launchpad') != null;
```
axosoft-ramint, line 269: Is it intentional that we are doing this access check/step after the "connect an integration" step? Once this becomes a Pro feature, should we be doing the access confirmation first? That way we don't let a user go all the way through with creating an account and connecting an integration, only to block them afterward?

src/constants.telemetry.ts
```diff
@@ -466,3 +466,5 @@ type SubscriptionEventData = {
 		Record<`previous.subscription.${string}`, string | number | boolean | undefined> &
 		Record<`previous.subscription.previewTrial.${string}`, string | number | boolean | undefined>
 >;
+
+export type TrackingContext = 'graph' | 'launchpad' | 'visual_file_history' | 'worktrees';
```
axosoft-ramint, line 470: This is being called `TrackingContext` now, and we moved it to the telemetry constants, but it's for more than just telemetry. Sending a context key to the login route actually changes the content that is shown on the landing page.

For example, here is the landing page for `https://gitkraken.dev/login?source=gitlens&context=graph`:

image

axosoft-ramint (commented):
Looking good, just a couple of questions/suggestions.

I'm also wondering how we want the indicator to work once this is live. Would it still show up, grab and summarize PRs in the hover, etc.?

Also, would it be weird that the hover says "connect an integration to get started" when really the user needs to upgrade to Pro first?

src/webviews/apps/welcome/welcome.html
```diff
@@ -626,7 +626,10 @@ <h3 class="t-eyebrow sticky">Popular</h3>
 							><gl-tooltip hoist content="Show Commit Graph"><span>Commit Graph</span></gl-tooltip
 							><gl-feature-badge class="super small"></gl-feature-badge>
 						</a>
-						<a href="command:gitlens.showFocusPage" data-requires="repo" aria-label="Open Launchpad"
+						<a
+							href="command:gitlens.showLaunchpad?%7B%22source%22%3A%22welcome%22%7D"
+							data-requires="repo"
```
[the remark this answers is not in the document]
eamodio: Good catch!

src/plus/focus/focus.ts
eamodio: Yeah it's intentional, as for most they will connect an integration which starts the trial.

axosoft-ramint (approved):
Tested as many cases as I could think of - LGTM

PR #3559, opened 2024-09-09
Refines 'Create Worktree' Flow

### LINKED ISSUE #3542: "Create Worktree" flow improvements
When using “create worktree”, there is a step where you have two options: “create worktree for branch” and “create worktree for new branch named…”

Just have one option: “Create worktree for branch”. If the branch you selected in a previous step is already a worktree, or is main, etc., “Create worktree” for branch” functions just like clicking “create worktree for new branch named…” did. Otherwise, it functions just like “Create worktree for branch” did.

Remove the option to “choose a different folder”. <-- **EDIT: We will do this later, not in this scope**

There is a bug where you enter the worktree flow after creating a new branch, where the option says “create worktree for new branch named…” even though you already created the branch. That should be fixed by this update too.

axosoft-ramint:
Closes #3542

Automatically chooses the `-b` flag or not based on if your chosen branch is already in a worktree.

Also, if the `-b` flag does apply, shows the "new branch name" prompt before the confirm step.

### DIFF
--- a/src/commands/git/worktree.ts
+++ b/src/commands/git/worktree.ts
@@ -717,7 +717,7 @@ export class WorktreeGitCommand extends QuickCommand<State> {
 		type StepType = FlagsQuickPickItem<CreateFlags, CreateConfirmationChoice>;
 		const defaultOption = createFlagsQuickPickItem<CreateFlags, Uri>(
 			state.flags,
-			state.worktree ? ['-b'] : [],
+			state.createBranch ? ['-b'] : [],
 			{
 				label: isRemoteBranch
 					? 'Create Worktree for New Local Branch'
@@ -726,7 +726,7 @@ export class WorktreeGitCommand extends QuickCommand<State> {
 					  : context.title,
 				description: '',
 				detail: `Will create worktree in $(folder) ${
-					state.worktree ? recommendedNewBranchFriendlyPath : recommendedFriendlyPath
+					state.createBranch ? recommendedNewBranchFriendlyPath : recommendedFriendlyPath
 				}`,
 			},
 			recommendedRootUri,
--- a/src/commands/git/worktree.ts
+++ b/src/commands/git/worktree.ts
@@ -26,6 +26,7 @@ import {
 } from '../../git/models/reference';
 import type { Repository } from '../../git/models/repository';
 import type { GitWorktree } from '../../git/models/worktree';
+import { getWorktreeForBranch } from '../../git/models/worktree';
 import { showGenericErrorMessage } from '../../messages';
 import type { QuickPickItemOfT } from '../../quickpicks/items/common';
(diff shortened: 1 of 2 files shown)


axosoft-ramint:
To test:

1. Use "create worktree" on a local branch that doesn't have a worktree. It should not prompt you for a branch name, and should just use the branch's name.
2. Use "create worktree" on a local branch which already has a worktree. It should automatically prompt for branch name and create a new branch and worktree.
3. Use "create branch" and then choose the "create branch and worktree" option. It should prompt for branch name and then let you create a worktree for the branch.

sergeibbb:
Hi @axosoft-ramint 

Re #1. I'm not sure that I get it correctly. I see the same behavior in pre-release. It creates and opens a worktree without a question:

https://github.com/user-attachments/assets/f875517c-7577-470d-a686-f18e3eb085cc

sergeibbb:
@axosoft-ramint  Same for create:

https://github.com/user-attachments/assets/32957d00-2b48-45c3-adfb-b7643d418d31

sergeibbb:
@axosoft-ramint 
Ah! That's clear now. Thank you!

sergeibbb:
Re test scenario 3. Looks like no difference:

main:
image
image


feature:
image
image

sergeibbb:
@axosoft-ramint 
Re 3.
> It should work like scenario 1 after using "Create branch in New Worktree". It should offer "Create Worktree for Branch" which just creates the worktree for the new branch you have created.

Ah, I see. it has replaced "Create worktree for new branch named" with a "create worktree for branch". Now I see it. But it behaves identically. In both cases (main and feature branch) I come to the same result: a worktree for a new branch named after it.

axosoft-ramint:
> @axosoft-ramint Re 3.
> 
> > It should work like scenario 1 after using "Create branch in New Worktree". It should offer "Create Worktree for Branch" which just creates the worktree for the new branch you have created.
> 
> Ah, I see. it has replaced "Create worktree for new branch named" with a "create worktree for branch". Now I see it. But it behaves identically. In both cases (main and feature branch) I come to the same result: a worktree for a new branch named after it.

Yes, this one is mainly a naming change because the previous naming of the option was confusing.

axosoft-ramint:
@sergeibbb I have fixed case 4 (and a separate bug with case 3 where it does not create the new branch). Would you mind testing all four cases once more?

sergeibbb:
@axosoft-ramint 

I confirm. #.4 fixed

sergeibbb:
scenario 2 works well.

axosoft-ramint:
> @axosoft-ramint
> 
> re scenario 3 When I have several repos in my workspace, it falls to selecting a repository step:

Good catch. This is an existing issue that this PR does not address, and we should write a follow-up ticket to fix it. Since it's not a new issue introduced by these changes, I will merge them for now.

PR #3524, opened 2024-09-11
Change GK account authentication to no longer use the VS Code authentication system

sergeibbb:
Currently we leverage the VS Code authentication system for our GK authentication, but this causes unnecessary prompts and limits our control over the experience.

We already implement both sides of the authentication process within GitLens, so we just need to avoid using VS Code as the middleman

### Roadmap
- [x] stop registering AccountAuthenticationProvider in VSCode
- [x] fix get-or-create behavior and honoring `createIfNeeded` and `silent` flags.
- ❌ ~~probably stop implementing AuthenticationProvider~~ ➡️ probably but in a follow-up
- [x] ~~stop relying on authentication events, e.g. stop handling them and make sure that we loos nothing by doing that (or fix the cases that are broken)~~ I've checked that don't relying on such events for GK authentication.

### Testing notes
- Be unauthenticated before the update:
  - general behavior
    - login
    - restart VS Code: you should keep your session
    - logout
    - login again
    - logout
    - restart VS Code: you should keep being unauthenticated.
  - integrations
    - you're able to connect-disconnect rich GitHub integration,
    - they survive restarting the application.
    - disconect GitHub, then go to the Account View, press "Integrations" button, come back to VS Code. Integrations become connected.
- Be authenticated before the update
  - check the integrations behavior from the previous section
  - check the general behavior from the previous section
- Test both "sign-in" and "sign-up" flows, when you're authenticated and unauthenticated in your browser

### DIFF
--- a/src/plus/gk/account/authenticationProvider.ts
+++ b/src/plus/gk/account/authenticationProvider.ts
@@ -3,7 +3,7 @@ import type {
 	AuthenticationProviderAuthenticationSessionsChangeEvent,
 	AuthenticationSession,
 } from 'vscode';
-import { authentication, Disposable, EventEmitter, window } from 'vscode';
+import { Disposable, EventEmitter, window } from 'vscode';
 import { uuid } from '@env/crypto';
 import type { Container, Environment } from '../../../container';
 import { CancellationError } from '../../../errors';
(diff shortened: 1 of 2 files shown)


axosoft-ramint (approved):
LGTM

sergeibbb:
Hi @eamodio @axosoft-ramint 

I've created a follow up ticket: https://github.com/gitkraken/vscode-gitlens/issues/3574 I suppose, it needs to be prioritized, milestonized, etc.

PR #3583, opened 2024-09-15
Fixed #1651 - Adds an option to open a file on click in Search & Compare

### LINKED ISSUE #1651: Allow opening file instead of diff

EhabY:
# Description

Closes https://github.com/gitkraken/vscode-gitlens/issues/1651. This PR adds an option in the GitLens settings to open the file when clicking on its entry in the Search & Compare instead of opening the diff view.

Since this issue was only for the Search & Compare, I only added this settings for this view. If there is need to add it to the other views then I can do that as well.

# Checklist


EhabY:
@eamodio Merged main into this branch, ready for review :pray:

src/views/nodes/commitFileNode.ts
```diff
@@ -166,6 +167,21 @@ export abstract class CommitFileNodeBase<
 			range = this.commit.file?.range ?? selectionToDiffRange(this.options?.selection);
 		}
 
+		const showOptions: TextDocumentShowOptions = { preserveFocus: true, preview: true };
+
+		const filesConfig = this.view.config.files;
+		if ('openDiffOnClick' in filesConfig && filesConfig.openDiffOnClick === false) {
+			return createCommand<[CommitFileNodeBase<Type, TView>, OpenFileAtRevisionCommandArgs]>(
+				'gitlens.views.openFileRevision',
+				'Open File at Revision',
```
EhabY, line 176: This relies on reusing the logic from the context menu item "Open File at Revision". The alternative is to use the generic command `gitlens.openFileRevision` and copy the logic to get the revision URI:

```ts
const revisionUri =
	node.commit.file?.status === 'D'
		? this.container.git
				.getRepositoryService(node.commit.repoPath)
				.getRevisionUri(
					(await node.commit.getPreviousSha()) ?? deletedOrMissing,
					node.commit.file.path,
				)
		: this.container.git.getRevisionUriFromGitUri(node.uri);
```

However this would require making this method `async` which wouldn't work :/

src/views/nodes/commitFileNode.ts
```diff
@@ -166,6 +167,21 @@ export abstract class CommitFileNodeBase<
 			range = this.commit.file?.range ?? selectionToDiffRange(this.options?.selection);
 		}
 
+		const showOptions: TextDocumentShowOptions = { preserveFocus: true, preview: true };
+
+		const filesConfig = this.view.config.files;
+		if ('openDiffOnClick' in filesConfig && filesConfig.openDiffOnClick === false) {
+			return createCommand<[CommitFileNodeBase<Type, TView>, OpenFileAtRevisionCommandArgs]>(
+				'gitlens.views.openFileRevision',
```
EhabY, line 175: Should this open the file at the revision or open current file? The context menu item has both "Open File" and "Open File at Revision" and same for the icons, though clicking the icon without [Alt] would open the file at the revision so I opted to make that the default for a single click

PR #3585, opened 2024-09-18
Upgrades eslint to v9 & flat config

eamodio:


src/env/browser/providers.ts
```diff
@@ -1,11 +1,12 @@
-import { Container } from '../../container';
-import { GitCommandOptions } from '../../git/commandOptions';
+import type { Container } from '../../container';
+import type { GitCommandOptions } from '../../git/commandOptions';
 // Force import of GitHub since dynamic imports are not supported in the WebWorker ExtensionHost
-import { GitHubGitProvider } from '../../plus/integrations/providers/github/githubGitProvider';
+// eslint-disable-next-line @typescript-eslint/consistent-type-imports
 import { GitProvider } from '../../git/gitProvider';
```
nzaytsev, line 5: Why the rule is disabled here?

nzaytsev (commented):
looks good, but there is one question about rule ignore and the code of e2e tests is not linted now
Screenshot 2024-09-18 at 14 22 58

tests/e2e/specs/baseTest.ts
```diff
@@ -60,6 +61,7 @@ export const test = base.extend<TestFixtures>({
 			await fs.promises.cp(logPath, logOutputPath, { recursive: true });
 		}
 	},
+	// eslint-disable-next-line no-empty-pattern
 	createTmpDir: async ({}, use) => {
```
nzaytsev, line 65: Maybe use underscore for unused param?
```suggestion
	// eslint-disable-next-line no-empty-pattern
	createTmpDir: async (_, use) => {
```

PR #3596, opened 2024-09-20
Change `branch create` and `branch rename` into normal commands

sergiolms:
# Description

This is the first part of #3528 that changes `branch create` and `branch rename` into normal commands.
Since GitLens performs most of the validation for branch names, there wasn't really much to error check here.

Preparing next PR with `branch delete` which will probably contain more validations.

# Checklist

- [x] I have followed the guidelines in the Contributing document
- [x] My changes follow the coding style of this project
- [x] My changes build without any errors or warnings
- [x] My changes have been formatted and linted
- [x] My changes include any required corresponding changes to the documentation (including CHANGELOG.md and README.md)
- [x] My changes have been rebased and squashed to the minimal number (typically 1) of relevant commits
- [x] My changes have a descriptive commit message with a short title, including a `Fixes $XXX -` or `Closes #XXX -` prefix to auto-close the issue that your PR addresses

PR #3597, opened 2024-09-20
3528 Change `branch` terminal-run commands into normal commands w/ proper error handling

### LINKED ISSUE #3528: Change `branch` terminal-run commands into normal commands w/ proper error handling
Refs: #3527

We may also want to consider #1828 when converting branch delete

Identify all the cases and create issues to track each one

```[tasklist]
### Tasks
- [ ] Add a draft title or issue reference here
```

sergiolms:
# Description

Part 2 including changes to `branch delete`.
Resolves #3528 

# Checklist

### DIFF
--- a/src/commands/git/branch.ts
+++ b/src/commands/git/branch.ts
@@ -393,7 +393,12 @@ export class BranchGitCommand extends QuickCommand {
 			if (state.flags.includes('--switch')) {
 				await state.repo.switch(state.reference.ref, { createBranch: state.name });
 			} else {
-				state.repo.branch(...state.flags, state.name, state.reference.ref);
+				await state.repo.branch({
+					create: {
+						name: state.name,
+						startRef: state.reference.ref,
+					},
+				});
 			}
 		}
 	}
@@ -614,7 +619,12 @@ export class BranchGitCommand extends QuickCommand {
 			state.flags = result;
 
 			endSteps(state);
-			state.repo.branch(...state.flags, state.reference.ref, state.name);
+			await state.repo.branch({
+				rename: {
+					old: state.reference.ref,
+					new: state.name,
+				},
+			});
 		}
 	}
 
--- a/src/env/node/git/git.ts
+++ b/src/env/node/git/git.ts
@@ -506,6 +506,10 @@ export class Git {
 		}
 	}
 
+	branch(repoPath: string, ...args: string[]) {
+		return this.git<string>({ cwd: repoPath }, 'branch', ...args);
+	}
+
 	branch__set_upstream(repoPath: string, branch: string, remote: string, remoteBranch: string) {
 		return this.git<string>({ cwd: repoPath }, 'branch', '--set-upstream-to', `${remote}/${remoteBranch}`, branch);
 	}
--- a/src/env/node/git/localGitProvider.ts
+++ b/src/env/node/git/localGitProvider.ts
@@ -1226,6 +1226,16 @@ export class LocalGitProvider implements GitProvider, Disposable {
 	}
 
 	@log()
+	async branch(repoPath: string, ...args: string[]): Promise<void> {
+		try {
+			await this.git.branch(repoPath, ...args);
+		} catch (ex) {
+			Logger.error(ex, getLogScope());
+			void showGenericErrorMessage(`Unable to create branch`);
+		}
+	}
+
+	@log()
 	async checkout(
 		repoPath: string,
 		ref: string,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 
 ## [Unreleased]
 
+### Changed
+- Change `branch create` and `branch rename` terminal-run commands into normal commands
+
 ### Fixed
 
 - Fixes [#3582](https://github.com/gitkraken/vscode-gitlens/issues/3582) - "Delete Branch" option is sometimes unexpectedly missing
--- a/src/git/gitProvider.ts
+++ b/src/git/gitProvider.ts
@@ -167,6 +167,7 @@ export interface GitProvider extends Disposable {
 			stash?: boolean | 'prompt';
 		},
 	): Promise<void>;
+	branch(repoPath: string, ...args: string[]): Promise<void>;
 	checkout(
 		repoPath: string,
 		ref: string,
--- a/src/git/gitProviderService.ts
+++ b/src/git/gitProviderService.ts
@@ -1326,6 +1326,12 @@ export class GitProviderService implements Disposable {
 	}
 
 	@log()
+	branch(repoPath: string | Uri, ...args: string[]): Promise<void> {
+		const { provider, path } = this.getProvider(repoPath);
+		return provider.branch(path, ...args);
+	}
+
+	@log()
 	checkout(
 		repoPath: string | Uri,
 		ref: string,
--- a/src/git/models/repository.ts
+++ b/src/git/models/repository.ts
@@ -47,6 +47,17 @@ export interface RepositoriesSortOptions {
 	orderBy?: RepositoriesSorting;
 }
 
+type GitBranchOptions = {
+	rename?: {
+		old: string;
+		new: string;
+	};
+	create?: {
+		name: string;
+		startRef: string;
+	};
+};
+
 const emptyArray = Object.freeze([]) as unknown as any[];
 
 const millisecondsPerMinute = 60 * 1000;
@@ -568,8 +579,17 @@ export class Repository implements Disposable {
 	}
 
 	@log()
-	branch(...args: string[]) {
-		void this.runTerminalCommand('branch', ...args);
+	branch(options: GitBranchOptions) {
+		const { rename, create } = options;
+		if (rename != null) {
+			return this.container.git.branch(this.uri, '-m', rename.old, rename.new);
+		}
+
+		if (create != null) {
+			return this.container.git.branch(this.uri, create.name, create.startRef);
+		}
+
+		throw new Error('Invalid branch options');
 	}
 
 	@log()
--- a/src/plus/integrations/providers/github/githubGitProvider.ts
+++ b/src/plus/integrations/providers/github/githubGitProvider.ts
@@ -453,6 +453,9 @@ export class GitHubGitProvider implements GitProvider, Disposable {
 	async applyChangesToWorkingFile(_uri: GitUri, _ref1?: string, _ref2?: string): Promise<void> {}
 
 	@log()
+	async branch(_repoPath: string, ..._args: string[]): Promise<void> {}
+
+	@log()
 	async branchContainsCommit(_repoPath: string, _name: string, _ref: string): Promise<boolean> {
 		return false;
 	}


eamodio:
@sergiolms Please apply similar changes from https://github.com/gitkraken/vscode-gitlens/pull/3596 to this PR

src/env/node/git/localGitProvider.ts
```diff
@@ -1240,6 +1250,57 @@ export class LocalGitProvider implements GitProvider, Disposable {
 		await this.git.branch(repoPath, '-m', oldName, newName);
 	}
 
+	@log()
+	async deleteBranch(
```
axosoft-ramint, line 1254: Nitpick: since this takes an array of branches as a param, shouldn't we call this function `deleteBranches`? Same with the provider service function for consistency, though that one can take either a single or multiple branches.

axosoft-ramint (commented):
Since I don't have too much context going into this, could you help me test it by:

1. Describing how to initiate the commands being changed (all the different ways we can get into the delete branch flow that you are aware of)
2. Describing what happens before this PR and what happens after the PR changes (essentially, what behavior should I expect to be different when using this PR)
3. Describing what areas may have regressions as a result of the PR changes that I should spot-check, if any

I did a quick code-scan and it looks good, just had one nitpick below:

src/env/node/git/localGitProvider.ts
sergiolms: Hey @axosoft-ramint ! Sure, let me go over your questions:
> 1. Describing how to initiate the commands being changed (all the different ways we can get into the delete branch flow that you are aware of)
This behavior hasn't changed. I usually trigger the delete branch action from the context menu in the graph. But since the context menu opens the command palette, you can directly execute the `GitLens: Git Delete Branch` action from the command palette as well.
Note that if you do the latter you will be asked to select the branch since it has no context.

> 2. Describing what happens before this PR and what happens after the PR changes (essentially, what behavior should I expect to be different when using this PR)
> 3. Describing what areas may have regressions as a result of the PR changes that I should spot-check, if any

Before this PR, executing the git action would run a terminal with the commands needed to delete the branch. 
See the screenshot, where I ran a create branch action to create a branch called `test_branch` and then I ran a delete branch action to delete the same branch:
Captura de pantalla 2024-10-01 a las 10 45 22

However, now we run the commands through our git binary and wrap any possible error. See the following example (it was forced to have a bad name, this is not possible because we filter invalid characters out):
Captura de pantalla 2024-09-30 a las 16 10 45

So the expected behavior is that when a `git branch` command is run (create, rename or delete), it doesn't open a terminal anymore and performs the action. If anything happens, it should display a toast with the error information.


> Nitpick: since this takes an array of branches as a param, shouldn't we call this function `deleteBranches`? Same with the provider service function for consistency, though that one can take either a single or multiple branches.

Yeah makes sense. I followed the naming convention of `createBranch` and `renameBranch` thinking of the `git branch` command, so I would group them all under the `branch` name if that makes sense. But it makes more sense for a method that may take multiple branches to be called `deleteBranches` instead. Good catch!

src/commands/git/branch.ts
```diff
@@ -516,10 +517,19 @@ export class BranchGitCommand extends QuickCommand {
...
+
+			try {
+				await state.repo.git.deleteBranch(state.references, {
+					force: state.flags.includes('--force'),
+					remote: state.flags.includes('--remotes'),
+				});
+			} catch (ex) {
+				Logger.error(ex);
+				// TODO likely need some better error handling here
+				return showGenericErrorMessage(
```
eamodio, line 529: We should be handling resumable/retriable errors. For example, if not merged, we should offer to retry with force.

src/commands/git/branch.ts
```diff
@@ -516,10 +517,19 @@ export class BranchGitCommand extends QuickCommand {
...
+					new BranchError(ex.reason, ex, state.references.map(r => r.name).join(', ')).message,
```
eamodio, line 530: Why are we creating a BranchError here? Isn't the one being thrown a BranchError?

src/commands/git/branch.ts
```diff
@@ -627,7 +637,7 @@ export class BranchGitCommand extends QuickCommand {
 			} catch (ex) {
 				Logger.error(ex);
 				// TODO likely need some better error handling here
-				return showGenericErrorMessage('Unable to rename branch');
+				return showGenericErrorMessage(new BranchError(ex.reason, ex, state.name).message);
```
eamodio, line 640: Same here

src/env/node/git/localGitProvider.ts
eamodio: The naming is one issue here, and maybe it should be broken into 2 methods, a `deleteBranch` and `deleteBranches`, but imo the bigger issue is the error handling. The first error breaks the whole operation. Unless we want to create a good AggregateError model so that a caller can handle all the issue, I would suggest that we leave the deleting multiple branches up to the calling code (we can wrap it in its own function to be shared -- in `git/actions/branch` probably), and just have the method here deal with a single branch (and optionally its remote) so that errors can be handled, prompted and maybe resumes independently.

eamodio (changes requested):

src/commands/git/branch.ts
sergiolms: Indeed, I don't really like this solution either. When we return the BranchError at git level we don't know the branch names, we just have an array of args. We could either guess the branch names, or change the number of args of `git.branch` to be more specific so we can have easier access. In this case we're reusing the BranchError but adding the branch name for context.
Another option is to turn BranchError into a Builder pattern that we can always call like `showGenericErrorMessage(ex.WithBranch(branchName))` or something along those lines to provide context on top layers.

sergiolms:
Now when an error such as unmerged branch happens, we display a prompt to retry it:

Captura de pantalla 2024-10-03 a las 12 59 07

We can turn this prompt into something reusable like a confirmation prompt we can parametrize as we find more cases

src/env/node/git/git.ts
```diff
@@ -160,6 +166,15 @@ function getStdinUniqueKey(): number {
 type ExitCodeOnlyGitCommandOptions = GitCommandOptions & { exitCodeOnly: true };
 export type PushForceOptions = { withLease: true; ifIncludes?: boolean } | { withLease: false; ifIncludes?: never };
 
+const branchErrorAndReason = [
+	[GitErrors.noRemoteReference, BranchErrorReason.NoRemoteReference],
+	[GitErrors.invalidBranchName, BranchErrorReason.InvalidBranchName],
+	[GitErrors.branchAlreadyExists, BranchErrorReason.BranchAlreadyExists],
+	[GitErrors.branchNotFullyMerged, BranchErrorReason.BranchNotFullyMerged],
+	[GitErrors.branchNotYetBorn, BranchErrorReason.BranchNotYetBorn],
+	[GitErrors.branchFastForwardRejected, BranchErrorReason.BranchFastForwardRejected],
```
axosoft-ramint, line 175: These don't exist yet - you'll need to define them. For my own curiosity, what is "branch not yet born?"

src/env/node/git/localGitProvider.ts
```diff
@@ -1252,13 +1253,87 @@ export class LocalGitProvider implements GitProvider, Disposable {
 	}
 
 	@log()
-	async createBranch(repoPath: string, name: string, ref: string): Promise<void> {
-		await this.git.branch(repoPath, name, ref);
+	createBranch(repoPath: string, name: string, ref: string): Promise<void> {
+		try {
+			return void this.git.branch(repoPath, name, ref);
```
axosoft-ramint, line 1258: This should just be `return this.git....`

src/env/node/git/localGitProvider.ts
```diff
@@ -1252,13 +1253,87 @@ export class LocalGitProvider implements GitProvider, Disposable {
...
+		} catch (ex) {
+			if (ex instanceof BranchError) {
+				throw ex.WithBranch(branch.name);
```
axosoft-ramint, line 1261: `branch` doesn't exist in this context. Should just be `name`

src/env/node/git/localGitProvider.ts
```diff
@@ -1252,13 +1253,87 @@ export class LocalGitProvider implements GitProvider, Disposable {
...
+			}
+
+			throw ex;
+		}
+	}
+
+	@log()
+	renameBranch(repoPath: string, oldName: string, newName: string): Promise<void> {
+		try {
+			return void this.git.branch(repoPath, '-m', oldName, newName);
```
axosoft-ramint, line 1271: This should just be `return this.git....`

src/env/node/git/localGitProvider.ts
```diff
@@ -1252,13 +1253,87 @@ export class LocalGitProvider implements GitProvider, Disposable {
...
+		} catch (ex) {
+			if (ex instanceof BranchError) {
+				throw ex.WithBranch(branch.name);
```
axosoft-ramint, line 1274: `branch` doesn't exist in this context. Should just be `oldName`

src/env/node/git/localGitProvider.ts
```diff
@@ -1252,13 +1253,87 @@ export class LocalGitProvider implements GitProvider, Disposable {
...
+				args.push('--force');
+			}
+
+			if (!options.remote || !branch.upstream) {
+				await this.git.branch(repoPath, ...args, branch.ref);
+				return;
+			}
+
+			const remote = getRemoteNameFromBranchName(branch.upstream.name);
+			remoteCommit = await this.git.rev_list(repoPath, `refs/remotes/${remote}/${branch.ref}`, {
```
axosoft-ramint, line 1309: `remoteCommit` needs to be declared/defined either here or earlier

src/env/node/git/localGitProvider.ts
```diff
@@ -1252,13 +1253,87 @@ export class LocalGitProvider implements GitProvider, Disposable {
...
+			});
+
+			await this.git.branch(repoPath, '--delete', '--remotes', `${remote}/${branch.ref}`);
+
+			try {
+				await this.git.branch(repoPath, ...args, branch.ref);
+			} catch (ex) {
+				// If it fails, restore the remote branch
+				await this.git.update_ref(repoPath, `refs/remotes/${remote}/${branch.ref}`, commit);
+				await this.git.branch__set_upstream(repoPath, branch, remote, branch);
```
axosoft-ramint, line 1320: `commit` doesn't exist in this context.

`branch` is a `GitBranchReference` and `branch__set_upstream` is looking for a string

axosoft-ramint (changes requested):
Several build errors here, as well as conflicts with main. Will test once it's rebased and the errors are resolved

axosoft-ramint (approved):
Not sure what commands are affected, and the only one I noticed was branch delete, so I tried it with a few cases and looks like it works and doesn't open my terminal.

LGTM - let me know if there are other cases I should test

axosoft-ramint (commented):
When attempting to delete a remote branch that no longer exists, I do not get the right error message in the notification area. I get a push error that doesn't describe the issue well:

image

It is true that the operation happening under the hood is `git push -d origin cool-branch` but I expected an error relating to "unable to delete branch because the remote branch was already deleted" or however you have the messaging set up.

The error I get in the logs is more descriptive/accurate:

image

sergiolms:
> When attempting to delete a remote branch that no longer exists, I do not get the right error message in the notification area. I get a push error that doesn't describe the issue well:
> [...]
> It is true that the operation happening under the hood is `git push -d origin cool-branch` but I expected an error relating to "unable to delete branch because the remote branch was already deleted" or however you have the messaging set up.

True. Since this action is running under git push it wasn't covered by the git branch errors. I added an additional case to git push to handle that case:

Captura de pantalla 2024-11-11 a las 15 22 58

eamodio:
This has been integrated in other changes

Issue #3536, opened 2024-09-06
Integrate accounts view into home


d13:
It would be great to remove the accounts view and incorporate its functionality directly into the Home view as maybe a footer on the view.

Here is a [Cloud Patch](https://gitkraken.dev/link/drafts/3b34c140-0b60-4f01-b611-36eab2884e02?type=patch) that almost completely implements this, but needs some more functionality (e.g. updating sub/integrations based on click) and we need to revisit the UX as what we have currently is far too big as a footer. So either we can dramatically slim it down, or add a “slim” mode, that auto-expands when you click into it.

PR #3609, opened 2024-09-23
Removes account view

d13:
The home view now displays account information, so the account view is no longer needed.

2nd piece of https://github.com/gitkraken/vscode-gitlens/issues/3536


eamodio (changes requested):
Looks good, but haven't cleaned up the commands and registrations in the package.json

eamodio (approved):

PR #3610, opened 2024-09-24
fix E2E tests

sergiolms:
# Description
E2E tests were giving the following error:
Captura de pantalla 2024-09-24 a las 13 25 23

This is caused because Playwright requires that parameter to be a destructured object, but our eslint rules collide with that because we don't allow empty patterns. The solution is to ignore that eslint line for this case.

In addition, I set the maximum timeout from 5s to 10s, as in some cases if VSCode takes too long to load, the test would timeout before even starting GitLens.

# Checklist

### DIFF
--- a/tests/e2e/playwright.config.ts
+++ b/tests/e2e/playwright.config.ts
@@ -22,4 +22,5 @@ export default defineConfig({
 			},
 		},
 	],
+	testMatch: 'specs/*.test.ts',
 });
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -9,3 +9,4 @@ node_modules
 images/settings
 gitlens-*.vsix
 tsconfig*.tsbuildinfo
+test-results
```
eamodio, line 12: Shouldn't need this because I moved the test-results into our `out` folder

eamodio (approved):
Thanks!

.gitignore
sergiolms: Ooh you're right, perhaps I had it from before that change. Removing it now

PR #3613, opened 2024-09-24
Slims `Repository` to use proxied `GitProviderService`

eamodio:
This avoids reimplementing many `GitProviderService` methods on `Repository`


axosoft-ramint (approved):
Conflicts need to be resolved before merge

PR #3624, opened 2024-09-27
Adds debug commands and utilities to mock/restore subscription state

axosoft-ramint:
Utility functions for debugging purposes only.

Also included in this PR are some fixes to types in the license response of check-in, since all license keys are not guaranteed to be in the response.

### DIFF
--- a/src/plus/gk/utils.ts
+++ b/src/plus/gk/utils.ts
@@ -1,5 +1,6 @@
 import type { MessageItem } from 'vscode';
 import { window } from 'vscode';
+import { createFromDateDelta } from '../../system/date';
 import { configuration } from '../../system/vscode/configuration';
 import { getContext } from '../../system/vscode/context';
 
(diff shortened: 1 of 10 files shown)


src/plus/gk/utils.ts
```diff
@@ -24,3 +25,29 @@ export async function ensurePlusFeaturesEnabled(): Promise<boolean> {
...
+
+export function getPreviewTrialAndDays(isDebugging: boolean) {
+	const startedOn = new Date();
+
+	let days: number;
+	let expiresOn = new Date(startedOn);
+	if (isDebugging) {
+		expiresOn = createFromDateDelta(expiresOn, { minutes: 1 });
+		days = 0;
+	} else {
```
eamodio, line 37: We can probably nuke this now and the "reset" command too?

eamodio (approved):
:shipit:

src/plus/gk/utils.ts
axosoft-ramint: Good point. Going to go ahead and remove them.

axosoft-ramint: Actually, let's keep the "reset" command for subscription because it wipes the stored subscription, which is something these debug utils don't mess with. So it feels like it still has separate utility. Removing this special case date logic for preview trials seems fine though.

PR #3629, opened 2024-09-28
Improves debug account simulator

eamodio:
- Enhanced the UI/UX to be more "stateful" and holistic
- Changes the non-account states to "remove" the account
- Removed the restore command as I added the functionality directly to the main command


src/plus/integrations/integrationService.ts
```diff
@@ -126,6 +126,7 @@ export class IntegrationService implements Disposable {
 	}
 
 	private onDidChangeSubscription(e: SubscriptionChangeEvent) {
+		// TODO is this correct?
```
axosoft-ramint, line 129: This ensures that when we log out of our GK account, any connected cloud integrations are disconnected.

axosoft-ramint (approved):
LGTM, thanks!

src/plus/integrations/integrationService.ts
eamodio: Ah -- Can you update that comment then please?

PR #3633, opened 2024-09-30
Adds discovered/found repos to local path map

### LINKED ISSUE #3555: Update shared repo mapping file when you open a repo
When GitLens discovers an open repo, it should automatically add it as an entry to the shared repo disk mapping file (using all of its remotes, its first commit sha, etc.) if it’s not in there already.

Currently it sometimes prompts you to locate a repo even though you already opened it in the past, which feels weird.

axosoft-ramint:
Closes #3555

### DIFF
--- a/src/git/gitProviderService.ts
+++ b/src/git/gitProviderService.ts
@@ -2448,10 +2448,7 @@ export class GitProviderService implements Disposable {
 		const scope = getLogScope();
 		for (const repo of repos) {
 			try {
-				await this.container.repositoryIdentity.addFoundRepositoryToMap(
-					repo,
-					await this.container.repositoryIdentity.getRepositoryIdentity(repo),
-				);
+				await this.container.repositoryIdentity.addFoundRepositoryToMap(repo);
 			} catch (ex) {
 				Logger.error(ex, scope);
 			}
--- a/src/git/gitProviderService.ts
+++ b/src/git/gitProviderService.ts
@@ -655,16 +655,7 @@ export class GitProviderService implements Disposable {
 			if (added.length) {
 				// Defer the event trigger enough to let everything unwind
 				queueMicrotask(() => this.fireRepositoriesChanged(added));
-				for (const addedRepo of added) {
-					try {
-						await this.container.repositoryIdentity.addFoundRepositoryToMap(
-							addedRepo,
-							await this.container.repositoryIdentity.getRepositoryIdentity(addedRepo),
-						);
-					} catch (ex) {
-						Logger.error('Could not add discovered repository to identity map: ', ex);
-					}
-				}
+				queueMicrotask(() => this.writeAddedReposToPathMap(added));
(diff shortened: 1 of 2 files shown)


src/git/gitProviderService.ts
```diff
@@ -655,6 +655,16 @@ export class GitProviderService implements Disposable {
...
+				for (const addedRepo of added) {
+					try {
+						await this.container.repositoryIdentity.addFoundRepositoryToMap(
+							addedRepo,
+							await this.container.repositoryIdentity.getRepositoryIdentity(addedRepo),
+						);
+					} catch (ex) {
+						Logger.error('Could not add discovered repository to identity map: ', ex);
+					}
+				}
```
axosoft-ramint, line 667: @eamodio What are your thoughts on me putting this logic in its own function and just using `queueMicrotask` for it here and in the other place below?

axosoft-ramint: Ended up doing it, and added gating: https://github.com/gitkraken/vscode-gitlens/pull/3633/commits/7130c687ca4c996a609dfa59c39d7772c392ad49

src/git/gitProviderService.ts
```diff
@@ -2440,6 +2442,22 @@ export class GitProviderService implements Disposable {
...
+	@gate()
+	@log()
+	async writeAddedReposToPathMap(repos: Repository[]): Promise<void> {
+		const scope = getLogScope();
+		for (const repo of repos) {
+			try {
+				await this.container.repositoryIdentity.addFoundRepositoryToMap(
+					repo,
+					await this.container.repositoryIdentity.getRepositoryIdentity(repo),
+				);
```
eamodio, line 2454: Can we hide the need to call `getRepositoryIdentity` into the `addFoundRepositoryToMap` call, or at least allow it, but have that method do it if missing.

src/plus/repos/repositoryIdentityService.ts
```diff
@@ -25,6 +25,16 @@ export class RepositoryIdentityService implements Disposable {
 		return this.locateRepository(identity, options);
 	}
 
+	async getRepositoryIdentity(repository: Repository): Promise<RepositoryIdentityDescriptor> {
+		const bestRemote = await this.container.git.getBestRemoteWithProvider(repository.uri);
+		return {
+			name: repository.name,
+			initialCommitSha: await this.container.git.getFirstCommitSha(repository.uri),
```
eamodio, line 32: Please execute these 2 in parallel

eamodio (changes requested):
Looks good -- minor tweaks requested

axosoft-ramint:
@eamodio Improvements you mentioned are in. Ready to review again when you're able. Thanks!

eamodio (approved):
Added some tweaks (which can be squashed). So, if it looks good to you, it's ready to go.

PR #3635, opened 2024-10-01
Attempts to pre-open PR branch when using Switch or Open in Worktree

### LINKED ISSUE #3627: Improve repo location in Deep Link service for PRs
The "add repo" case in the deep link service, which the Launchpad "Open in Worktree" command uses, doesn't leverage everything we know about the PR in the Launchpad. It should be able to leverage functions like `getOrOpenPullRequestRepository` from pullRequest.ts (using whatever we can pull through over the link like head and base repo URLs).

Right now the "add repo" case only cares about the remote url, which is not as in-depth. We should expand it to do more to get to the opened repo when the target is a PR.

axosoft-ramint:
Closes #3627

Leverages `getOrOpenPullRequestRepository` to try to pre-identify the PR's repo, which then skips the "repo match" step in the deep link flow.

Note: this has undesirable effects with virtual repos, so I added a `skipVirtual` open to skip opening virtual uris for now. A bigger refactor of target opening logic (moving it out of the deep link service into its own service) will probably include better virtual repo support in the flow, but for now it doesn't work for most cases of `Open in Worktree`.

### DIFF
--- a/src/git/models/pullRequest.ts
+++ b/src/git/models/pullRequest.ts
@@ -309,7 +309,7 @@ export function getVirtualUriForPullRequest(pr: PullRequest): Uri | undefined {
 export async function getOrOpenPullRequestRepository(
 	container: Container,
 	pr: PullRequest,
-	options?: { promptIfNeeded?: boolean },
+	options?: { promptIfNeeded?: boolean; skipVirtual?: boolean },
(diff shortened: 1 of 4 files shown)


eamodio (approved):
:shipit:

PR #3636, opened 2024-10-01
Completes the cloud integration reauth flow by disconnecting cloud token

### LINKED ISSUE #3618: Implement reauthentication for Cloud Integrations
We have flows in GitLens for reauthing a token (for example, notification of 'your <provider> token is invalid or expired. Would you like to try reauthenticating?'). When using cloud integrations, this currently just opens the `/connect` flow and doesn't change anything about the token.

To fully reauth, the client should:
1. Wipe the local session for the cloud integration
2. Call the backend route to delete the token
3. Call `/connect` which would then allow the user to re-auth with that provider and use the new token.

axosoft-ramint:
Closes #3618

When doing a reauth with a cloud integration, disconnects the token from the backend before initiating a connect flow.

### DIFF
--- a/src/constants.telemetry.ts
+++ b/src/constants.telemetry.ts
@@ -65,6 +65,12 @@ export type TelemetryEvents = {
 		'integration.connected.ids': string | undefined;
 	};
 
+	/** Sent when disconnecting a provider from the api fails*/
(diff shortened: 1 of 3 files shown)


eamodio (approved):
:shipit:

PR #3640, opened 2024-10-03
Adds refresh to account section

axosoft-ramint:


package.json
```diff
@@ -8091,6 +8091,12 @@
 				"category": "GitLens",
 				"icon": "$(refresh)"
 			},
+			{
+				"command": "gitlens.views.home.account.refresh",
+				"title": "Refresh",
+				"category": "GitLens",
+				"icon": "$(refresh)"
+			},
```
eamodio, line 8099: Unless a command needs to show up in VS Code controlled UI (command palette, menus) we don't need to register them here.

So in this case we can nuke this.

package.json
```diff
@@ -11482,6 +11488,10 @@
 					"command": "gitlens.views.home.refresh",
 					"when": "false"
 				},
+				{
+					"command": "gitlens.views.home.account.refresh",
+					"when": "false"
+				},
```
eamodio, line 11494: Same as above

src/webviews/apps/plus/shared/components/home-account-content.ts
```diff
@@ -219,6 +219,13 @@ export class GLHomeAccountContent extends LitElement {
 					this.hasAccount,
 					() => html`
 						<span class="header__actions">
+							<gl-button
+								appearance="toolbar"
+								href="command:gitlens.views.home.account.refresh"
+								tooltip="Refresh"
+								aria-label="Refresh"
+								><code-icon icon="refresh"></code-icon
+							></gl-button>
```
eamodio, line 228: Should we think of this as a "refresh" or a "re-sync"? @justinrobots 

If "refresh" then this is good, but I'd change the tooltip to "Refresh Account Status"
If "re-sync" then maybe, use the sync icon and the tooltip to be "Synchronize Account Status"

eamodio (changes requested):

src/webviews/apps/plus/shared/components/home-account-content.ts
d13: Sync does seem _a little_ more concise and it would add a better distinction from the view's refresh.

axosoft-ramint:
@eamodio @d13 Updated the command name, tooltip and icon, and removed it from package.JSON.

d13 (approved):
LGTM

PR #3641, opened 2024-10-03
Adds contribution telemetry

eamodio:


axosoft-ramint (approved):
Seems to work well. I like that we capture average too - can be combined with total contributors to convert to percentages in the distributions.

LGTM!

PR #3658, opened 2024-10-09
Uses root repo path for worktree create

### LINKED ISSUE #3657: Creating a worktree from within a worktree chooses the wrong path

eamodio:
Fixes #3657

This needs some vetting of all the different worktree create options (default folder, choosing a new root, etc) to make sure it doesn't regress any behavior.

I also added an additional commit to honor a chosen root folder rather than always adding the "trailer" (e.g. `<repo>.worktrees` ) onto the root


sergeibbb:
my testing:
- created a worktree from a worktree that is inside of the `<root>` ✅
- created a worktree from a worktree with complex branch name, such as `<root>/featrue/feature-name` ✅
- created a worktree from a worktree in a detached folder and check that it uses the same `<root>` as all other worktrees. ✅
- New root:
	- open a worktree in the old root
	- create a worktree choosing a new root
	- then use it to open a new worktree with the same root ⚠️ Alarm!!! It uses the old root!
	- then use it to open a new worktree with changing the root again ⏺️
	- Then use the opened worktree of the old root to create a new worktree in the root. Check that it uses the new root. ⏺️

PR #3663, opened 2024-10-09
Avoids extra repo creation during discovery

eamodio:


axosoft-ramint:
Can I get some context into this one? What problem is it solving, how do I reproduce the issue and verify that it fixes it, etc. At minimum, a reference to an issue would be nice.

eamodio:
Sorry, there isn't any issue. This is just an avoidance of unnecessary work when we are discovering repositories. Stops creating extra repo objects that will ultimately just get discarded later in the flow.

Just wanted another pair of eyes on it to see if there was anything glaring. The part I'm worried about the most is the behavior if a "closed" repo exists but we are discovering it and it should be opened. We are triggering and even during discovery which should be fine but timing could be odd

axosoft-ramint:
Gotcha, thanks!

axosoft-ramint (approved):
Looks good on a quick scan/pass. It's a bit weird that we're doing the logging in `isUntrackedRepo` but I can see why we wanted to put it there to consolidate the calls.

eamodio:
@axosoft-ramint I updated it to change the function to be a `maybeAddRepo`, so it should feel less odd about the logging

d13 (approved):
LGTM

PR #3664, opened 2024-10-10
Add support to localhost urls while debugging

sergiolms:
# Description

These changes will allow testing local development on APIs and backend services of gk-dev. With this, you can now set the following settings in `settings.json` (made up urls for the example):
```json
{
	"gitkraken.env": "local",
	"gitkraken.url": {
		"api": "http://localhost:1314/api",
		"gkdev": {
			"base": "http://localhost:1314",
			"api": "http://localhost:1314/gkdev/api"
		}
	}
}
```

NOTE: URLs will only be applied when `gitkraken.env` is set to `local`. Also, the `local` environment has the same restrictions as `dev` and `staging` (only available if it's pre-release or debugging).

# Checklist

### DIFF
--- a/src/container.ts
+++ b/src/container.ts
@@ -98,7 +98,7 @@ import type { WebviewViewProxy } from './webviews/webviewsController';
 import { WebviewsController } from './webviews/webviewsController';
 import { registerWelcomeWebviewPanel } from './webviews/welcome/registration';
 
-export type Environment = 'local' | 'dev' | 'staging' | 'production';
+export type Environment = 'dev' | 'staging' | 'production';
 
 export class Container {
 	static #instance: Container | undefined;
@@ -527,7 +527,6 @@ export class Container {
 	get env(): Environment {
 		if (this.prereleaseOrDebugging) {
 			const env = configuration.getAny('gitkraken.env');
-			if (env === 'local') return 'local';
 			if (env === 'dev') return 'dev';
 			if (env === 'staging') return 'staging';
 		}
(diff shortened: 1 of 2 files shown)


src/container.ts
```diff
@@ -919,6 +919,11 @@ export class Container {
 
 	@memoize()
 	private get baseGkDevUri(): Uri {
+		if (this.prereleaseOrDebugging) {
```
axosoft-ramint, line 922: I suggest instead of `this.prereleaseOrDebugging`, to allow the value `local` for `gitkraken.env` and then checking that here.

And if it is `local` and this url is not provided, we should throw because we don't want someone targeting prod and thinking they were on a local setup.

Same with the other cases.

sergiolms: I started with that `local` value for `env`, but if I do that and you only want to overwrite 1 URL, you still have to manually put the other ones because `env` won't tell you anymore if you're on dev, staging or prod. Which is totally okay if we want to do that and make the developer aware they have to manually set all URLs in config. In fact, my first commit does that, and I could retake it if we want to go down this path and throw instead of setting a default URL, as you said.

However, I found it unconvenient that you had to manually set all URLs if you wanted to overwrite just one, so I switched to a `useLocal` boolean that would "enable" the use of URLs, but again it seemed like setting an URL should be enough to specify that it should use it. With that, you would overwrite the desired URL, and leave all the rest up to `env` to decide where to point at.

Again, I'm totally up for both options so I appreciate the point of view

PR #3670, opened 2024-10-11
3529 Change `tag` terminal-run commands into normal commands w/ proper error handling

### LINKED ISSUE #3529: Change `tag` terminal-run commands into normal commands w/ proper error handling
Refs: #3527

Identify all the cases and create issues to track each one

```[tasklist]
### Tasks
- [ ] Add a draft title or issue reference here
```

sergiolms:
# Description

Resolves #3529 

Here's what an error looks like now:

Captura de pantalla 2024-10-11 a las 14 09 02

# Checklist

### DIFF
--- a/src/env/node/git/git.ts
+++ b/src/env/node/git/git.ts
@@ -22,6 +22,8 @@ import {
 	PushErrorReason,
 	StashPushError,
 	StashPushErrorReason,
+	TagError,
+	TagErrorReason,
 	WorkspaceUntrustedError,
 } from '../../../git/errors';
 import type { GitDir } from '../../../git/gitProvider';
@@ -32,7 +34,6 @@ import type { GitUser } from '../../../git/models/user';
 import { parseGitBranchesDefaultFormat } from '../../../git/parsers/branchParser';
 import { parseGitLogAllFormat, parseGitLogDefaultFormat } from '../../../git/parsers/logParser';
 import { parseGitRemoteUrl } from '../../../git/parsers/remoteParser';
-import { parseGitTagsDefaultFormat } from '../../../git/parsers/tagParser';
 import { splitAt } from '../../../system/array';
 import { log } from '../../../system/decorators/log';
 import { join } from '../../../system/iterable';
@@ -100,6 +101,10 @@ export const GitErrors = {
 	tagConflict: /! \[rejected\].*\(would clobber existing tag\)/m,
 	unmergedFiles: /is not possible because you have unmerged files/i,
 	unstagedChanges: /You have unstaged changes/i,
+	tagAlreadyExists: /tag .* already exists/i,
+	tagNotFound: /tag .* not found/i,
+	invalidTagName: /invalid tag name/i,
+	remoteRejected: /rejected because the remote contains work/i,
 };
 
 const GitWarnings = {
@@ -159,6 +164,14 @@ function getStdinUniqueKey(): number {
 
 type ExitCodeOnlyGitCommandOptions = GitCommandOptions & { exitCodeOnly: true };
 export type PushForceOptions = { withLease: true; ifIncludes?: boolean } | { withLease: false; ifIncludes?: never };
+
+const tagErrorAndReason = [
+	[GitErrors.tagAlreadyExists, TagErrorReason.TagAlreadyExists],
+	[GitErrors.tagNotFound, TagErrorReason.TagNotFound],
+	[GitErrors.invalidTagName, TagErrorReason.InvalidTagName],
+	[GitErrors.permissionDenied, TagErrorReason.PermissionDenied],
+	[GitErrors.remoteRejected, TagErrorReason.RemoteRejected],
+];
 
 export class Git {
 	/** Map of running git commands -- avoids running duplicate overlaping commands */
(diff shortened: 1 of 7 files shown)


src/git/errors.ts
```diff
@@ -488,3 +488,68 @@ export class WorktreeDeleteError extends Error {
...
+export const enum TagErrorReason {
+	TagAlreadyExists,
+	TagNotFound,
+	InvalidTagName,
+	PermissionDenied,
+	RemoteRejected,
+	Other,
+}
+
+export class TagError extends Error {
```
axosoft-ramint, line 501: What do you think about either splitting this up into one class per tag action (create/delete/get) like we do above this with worktrees/stashes/etc. or passing in the action so that the message is more descriptive to the action being done to the tag?

src/env/node/git/git.ts
```diff
@@ -2116,8 +2129,19 @@ export class Git {
 		return this.git<string>({ cwd: repoPath }, 'symbolic-ref', '--short', ref);
 	}
 
-	tag(repoPath: string) {
-		return this.git<string>({ cwd: repoPath }, 'tag', '-l', `--format=${parseGitTagsDefaultFormat}`);
+	async tag(repoPath: string, ...args: string[]) {
+		try {
+			const output = await this.git<string>({ cwd: repoPath }, 'tag', ...args);
```
axosoft-ramint, line 2134: In the removed lines I see a `-l` and a  `--format=${parseGitTagsDefaultFormat}` being passed as arguments, and we've replaced those with an `args[]` parameter that leaves the caller responsible for passing them in. A few questions/thoughts here:

1. Before it seems we were always using these two params, and now it looks like we're being more flexible. Are they not always needed?
2. I wonder if we can get away with having more readable params like an `options` objects param that has something like a `useDefaultFormat?: boolean` for example, among other things dependening on which flags we use, and then just construct the args in here based on what is flagged on in the options? I know we do use an open-ended `args` param in other git commands in this file, so maybe that pattern is better and more flexible - just wanted to pitch the idea and get your thoughts.

axosoft-ramint (commented):
A couple of thoughts I wanted to get your input on before testing

src/git/errors.ts
sergiolms: I'm not entirely sure if it's worth splitting into one class per action, as it may cause a lot of duplicated code. In fact, I wanted to revisit the errors once all git actions are done, because it may be interesting to have a more generic `GitError` class that receives information about the git command and action.
I agree on the passing the action as an argument to the constructor (and maybe even with a builder method as well). I'll make the changes, thanks for the feedback!

src/env/node/git/git.ts
sergiolms: > 1. Before it seems we were always using these two params, and now it looks like we're being more flexible. Are they not always needed?
> 2. I wonder if we can get away with having more readable params like an `options` objects param that has something like a `useDefaultFormat?: boolean` for example, among other things dependening on which flags we use, and then just construct the args in here based on what is flagged on in the options? I know we do use an open-ended `args` param in other git commands in this file, so maybe that pattern is better and more flexible - just wanted to pitch the idea and get your thoughts.

I'll try to answer both, let me know if you want me to get more in detail with anything. Before, the `tag` method was only used from `localGitProvider:getTags` method to obtain the list of tags. However, with these changes, I've split the responsibility to the different layers:

- `commands/git/tag`:  Validates the command steps and arguments (for most cases), such as an invalid name, non-empty strings, etc. This layer also call's the corresponding git provider service method with options (if necessary).
- `git/gitProviderService`: At this layer we only retrieve the git provider and call the method with the same args. This layer is only intended to validate the provider has the requested method.
- `env/node/git/localGitProvider` (or the equivalent for remotes): Is in charge of doing the pertinent git calls and operations given a set of parameters and options. If we're talking about remotes, it will perform API calls. In this case, it runs git with a certain set of arguments given some parameters. It is also in charge of catching and adding context to errors that may happen during the execution of those commands.
- `env/node/git/git`: Lastly, at this layer is where the most atomic action happens, that is calling git with a set of arguments and catching any error that may happen to transform it into manageable errors.

Yes, we could move the arguments and options to the lower git layer, but then it wouldn't be an atomic action. It's a little hard to distinguish when something should happen at `localGitProvider` and when in `git`. But I went with this distinction and I think it's okay so far. You shouldn't be calling `git` from anywhere else but git provider service, though. Right? So it makes sense to leave control of the git arguments to `localGitProvider`.

As for the `useDefaultFormat` I think it's definitely something we can put into `getTags`' options arg. :+1: 

I tried to explain myself, but if you still think we could organize it different, please reach out to me and we can go over this 😄

axosoft-ramint: Sounds good - thanks for the detailed explanation!

sergiolms: I just reviewed what `parseGitTagsDefaultFormat` is used for, and without it `parseGitTags` method doesn't make much sense because it expects the tag data in a given format. I agree that providing a custom format to the `getTags` method makes it more flexible, but it would require the method to receive a custom tag format & parser.

This is doable, but since it goes out of the PR scope I would leave it for a separate task if you're okay with it

axosoft-ramint: That's fine with me, thanks for looking into it.

axosoft-ramint (changes requested):
Build errors:

image

I think most of them are solvable by setting up a type on `tagErrorAndReason`: `const tagErrorAndReason: [RegExp, TagErrorReason][] = ...`

Not sure on that last one though.

sergiolms:
@axosoft-ramint 
> I think most of them are solvable by setting up a type on `tagErrorAndReason`: `const tagErrorAndReason: [RegExp, TagErrorReason][] = ...`

Yup... That was mostly it. I'm not sure why it didn't pop up before. 🤷‍♂️  Thank you

axosoft-ramint (approved):
Tested creating and deleting a tag and it LGTM. Let me know if you want me to test other cases.

Issue #3543, opened 2024-09-06
Launchpad: Search for another PR


axosoft-ramint:
Add ability to search for a PR not in the Launchpad list. UX needs to be talked through.

Normally our algorithm provides the Launchpad with some subset of all PRs. So, the full list of PRs accessible to the user is wider. The current search looks only for PRs selected for the Launchpad and  we want user to be able to search among all other available PRs.

The current idea, but needs discussion and may be changed:
- User can use the text input in the Launchpad quickpick, either in the main Launchpad list/step or through some other/new step, to search for a PR using its url. For example: `https://github.com/gitkraken/vscode-gitlens/pull/3500`. Or if that fails, we could just do a PR search with the integration using the user's search query.
- If no item in the existing categorized list matches the search, we check other PRs in the repo (or repos) matching the user's text to see if we can find a matching PR, and show a Focus item (default to "other" category) for it.
- The user can then open that item and take actions on it like any other Launchpad item.

## Notes
- Should first discuss design/UX with the team. For example: should there instead be a "search for a different PR" button or action item in the Launchpad list? Would that make it more discoverable?
- Thinking of using `integration.getPullRequest` for the PR search when the user uses PR url, and `integration.searchPullRequests` for more general search, but the exact function/query may change depending on final design.
- Note that if we use `integration.getPullRequest` or `integration.searchPullRequests`, they must be implemented for GitLab since Launchpad supports GitLab as well (currently it is only implemented for GitHub). This might be a good opportunity to add this functionality to the shared provider library, but for now we can implement them locally for GitLab as needed.

### PR1. Search by URL GitHub
If I copy and paste a PR url directly into the input box on the main step, it should search for that PR
     
1. ✅ filter the current list for the specified PR if it's there
2. ✅ otherwise search for the specified PR by its ID
3. ✅ make sure it's debounced properly
4. Fix few search bugs:
  - ✅ Fix bug of not clearing the found item
  - ✅ Fix bug of incorrect local search by PR-id
5. ✅ be friendly to incomplete URLs or URLs with additional parts
6. 💡 waiting for code review feedback

### PR2. Add full search mode

Add an always visible item such as "Search everywhere". If it's clicked, switch to the mode that searches across all PRs the user has access to using the search terms.

1. ⏺️ ...
2. ⏺️ Bug: Local search cannot find items of collapsed groups.

### PR3. GitLab
Support GitLab MR urls and search across all PRs on GitLab.

[...]

PR #3684, opened 2024-10-18
Search in Launchpad by PR URL. GitHub (#3543)

sergeibbb:
# Description

When I copy and paste a PR URL directly into the input box on the main step, it should search for that PR

# Checklist

- [x] My changes have been rebased and squashed to the minimal number of relevant commits (~~typically 1~~ , **don't squash my lovely commits,** this PR is too big too be squashed).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 
 ## [Unreleased]
 
+### Added
+
+- Adds new ability to search for a GitHub PR in the _Launchpad;_ closes [#3543](https://github.com/gitkraken/vscode-gitlens/issues/3543)
+
 ## [15.6.2] - 2024-10-17
 
 ### Fixed
--- a/src/plus/launchpad/launchpad.ts
+++ b/src/plus/launchpad/launchpad.ts
@@ -419,6 +419,7 @@ export class LaunchpadCommand extends QuickCommand<State> {
 			i: LaunchpadItem,
 			ui: LaunchpadGroup,
 			topItem: LaunchpadItem | undefined,
+			alwaysShow: boolean | undefined,
 		): LaunchpadItemQuickPickItem => {
 			const buttons = [];
 
@@ -449,6 +450,7 @@ export class LaunchpadCommand extends QuickCommand<State> {
 					i.actionableCategory === 'other' ? '' : `${actionGroupMap.get(i.actionableCategory)![0]} \u2022  `
 				}${fromNow(i.updatedDate)} by @${i.author!.username}`,
 
+				alwaysShow: alwaysShow,
 				buttons: buttons,
 				iconPath: i.author?.avatarUrl != null ? Uri.parse(i.author.avatarUrl) : undefined,
 				item: i,
@@ -457,7 +459,7 @@ export class LaunchpadCommand extends QuickCommand<State> {
 			};
 		};
 
-		const getItems = (result: LaunchpadCategorizedResult) => {
+		const getItems = (result: LaunchpadCategorizedResult, isSearching?: boolean) => {
 			const items: (LaunchpadItemQuickPickItem | DirectiveQuickPickItem | ConnectMoreIntegrationsItem)[] = [];
 
 			if (result.items?.length) {
@@ -472,18 +474,21 @@ export class LaunchpadCommand extends QuickCommand<State> {
 				for (const [ui, groupItems] of uiGroups) {
 					if (!groupItems.length) continue;
 
-					items.push(...buildGroupHeading(ui, groupItems.length));
+					if (!isSearching) {
+						items.push(...buildGroupHeading(ui, groupItems.length));
+						if (context.collapsed.get(ui)) {
+							continue;
+						}
+					}
 
-					if (context.collapsed.get(ui)) continue;
-
-					items.push(...groupItems.map(i => buildLaunchpadQuickPickItem(i, ui, topItem)));
+					items.push(...groupItems.map(i => buildLaunchpadQuickPickItem(i, ui, topItem, isSearching)));
 				}
 			}
 
 			return items;
 		};
 
-		function getItemsAndPlaceholder() {
+		function getItemsAndPlaceholder(isSearching?: boolean) {
 			if (context.result.error != null) {
 				return {
 					placeholder: `Unable to load items (${
@@ -506,17 +511,18 @@ export class LaunchpadCommand extends QuickCommand<State> {
 
 			return {
 				placeholder: 'Choose an item to focus on',
-				items: getItems(context.result),
+				items: getItems(context.result, isSearching),
 			};
 		}
 
 		const updateItems = async (
 			quickpick: QuickPick<LaunchpadItemQuickPickItem | DirectiveQuickPickItem | ConnectMoreIntegrationsItem>,
 		) => {
+			const search = quickpick.value;
 			quickpick.busy = true;
 
 			try {
-				await updateContextItems(this.container, context, { force: true });
+				await updateContextItems(this.container, context, { force: true, search: search });
 
 				const { items, placeholder } = getItemsAndPlaceholder();
 				quickpick.placeholder = placeholder;
@@ -527,8 +533,7 @@ export class LaunchpadCommand extends QuickCommand<State> {
 		};
 
 		const { items, placeholder } = getItemsAndPlaceholder();
-
-		let groupsHidden = false;
+		const nonGroupedItems = items.filter(i => !isDirectiveQuickPickItem(i));
 
 		const step = createPickStep({
 			title: context.title,
@@ -543,29 +548,27 @@ export class LaunchpadCommand extends QuickCommand<State> {
 				LaunchpadSettingsQuickInputButton,
 				RefreshQuickInputButton,
 			],
-			onDidChangeValue: quickpick => {
+			onDidChangeValue: async quickpick => {
 				const { value } = quickpick;
-				const hideGroups = Boolean(quickpick.value?.length);
-
-				if (groupsHidden !== hideGroups) {
-					groupsHidden = hideGroups;
-					quickpick.items = hideGroups ? items.filter(i => !isDirectiveQuickPickItem(i)) : items;
-				}
-				const activeLaunchpadItems = quickpick.activeItems.filter(
-					(i): i is LaunchpadItemQuickPickItem => 'item' in i && !i.alwaysShow,
-				);
+				const hideGroups = Boolean(value?.length);
+				const consideredItems = hideGroups ? nonGroupedItems : items;
 
 				let updated = false;
-				for (const item of quickpick.items) {
+				for (const item of consideredItems) {
 					if (item.alwaysShow) {
 						item.alwaysShow = false;
 						updated = true;
 					}
 				}
-				if (updated) {
-					// Force quickpick to update by changing the items object:
-					quickpick.items = [...quickpick.items];
-				}
+
+				// By doing the following we make sure we operate with the PRs that belong to Launchpad initially.
+				// Also, when we re-create the array, we make sure that `alwaysShow` updates are applied.
+				quickpick.items =
+					updated && quickpick.items === consideredItems ? [...consideredItems] : consideredItems;
+
+				const activeLaunchpadItems = quickpick.activeItems.filter(
+					(i): i is LaunchpadItemQuickPickItem => 'item' in i,
+				);
 
 				if (!value?.length || activeLaunchpadItems.length) {
 					// Nothing to search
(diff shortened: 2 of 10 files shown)


src/plus/integrations/providers/github.ts
```diff
@@ -317,3 +318,52 @@ export class GitHubEnterpriseIntegration extends GitHubIntegrationBase<SelfHoste
...
+		prNumber = match[2];
+	}
+
+	if (prNumber == null) {
+		match = search.match(/([^/]+\/[^/]+)\/pull\/(\d+)/); // without domain
+		if (match != null) {
+			ownerAndRepo = match[1];
+			prNumber = match[2];
+		}
+	}
```
sergiolms, line 343: Is there a reason for checking with the `github.com` domain first? The matching groups don't seem to be different, you could directly run the search without the domain, right? If the URL has github domain in it, it will work anyway with just the second query.
Captura de pantalla 2024-10-21 a las 10 51 46

sergiolms (commented):
There are two odd things I've found while testing this. I haven't looked at the whole code yet:

1) I tried to paste this link: `https://github.com/gitkraken/vscode-gitlens/pull/3684/files#diff-e4b3bb605be44151fa3b67bc6a53b6178113c32d7f5ae627fd1b8fb4ba3e1a61R331-R343` but it didn't work until I removed the #diff... part:

https://github.com/user-attachments/assets/dc4847d3-fda2-4cb2-8a92-f26fd9d048dd

However it should be getting the correct parts 🤔 
Captura de pantalla 2024-10-21 a las 11 15 56


2) After opening a URL in the browser, some items disappear but they still count, creating some empty items you cannot interact with:



https://github.com/user-attachments/assets/36cc57e0-6b4a-40ca-87ca-852485bea085

src/plus/integrations/providers/github.ts
sergeibbb: @sergiolms 
correct. fixed.

sergeibbb:
@sergiolms 

> I tried to paste this link: https://github.com/gitkraken/vscode-gitlens/pull/3684/files#diff-e4b3bb605be44151fa3b67bc6a53b6178113c32d7f5ae627fd1b8fb4ba3e1a61R331-R343 but it didn't work until I removed the #diff... part:

Thank you. Fixed

sergeibbb:
@sergiolms 

> After opening a URL in the browser, some items disappear but they still count, creating some empty items you cannot interact with:

Good catch! I should not search on the next step.

sergeibbb:
@sergiolms

> After opening a URL in the browser, some items disappear but they still count, creating some empty items you cannot interact with:

It looks like existing behavior. Got to an item and put a query that doesn't match any command and these two item appear.


https://github.com/user-attachments/assets/1983a2e7-21e4-4453-aab8-a392873253fa

src/system/vscode/asyncDebouncer.ts
```diff
@@ -0,0 +1,134 @@
...
+/**
+ * This is similar to `src/system/function.ts: debounce` but it's for async tasks.
+ * The old `debounce` function does not awaits for promises, so it's not suitable for async tasks.
+ *
+ * Here the debouncer returns a promise that awaits task for completion.
+ * Also we can let tasks know if they are cancelled by passing a cancellation token.
+ *
+ * Despite being able to accept synchronous tasks, we always return a promise here. It's implemeted this way for simplicity.
+ */
+export function createAsyncDebouncer<T>(delay: number): Disposable & Deferrable<(task: AsyncTask<T>) => Promise<T>> {
```
axosoft-ramint, line 21: Would it work to move this to the same `src/system/function` file? It appears to me like this is another utility for working with functions.

src/system/function.ts
```diff
@@ -98,13 +97,6 @@ export function debounce<T extends (...args: any[]) => ReturnType<T>>(
 		lastThis = this;
 		lastCallTime = time;
 
-		if (isInvoking) {
-			if (timer == null) {
-				// Start the timer for the trailing edge.
-				timer = setTimeout(timerExpired, wait);
-				return result;
-			}
-		}
```
axosoft-ramint, line 107: Can you elaborate why this logic is being removed?

src/plus/launchpad/launchpadProvider.ts
```diff
@@ -609,15 +644,52 @@ export class LaunchpadProvider implements Disposable {
 		return repoRemotes;
 	}
 
+	mergeSearchedCategorizedItems(
+		permanent: LaunchpadCategorizedResult,
```
axosoft-ramint, line 648: What do we mean by `permanent` here?

src/plus/launchpad/launchpadProvider.ts
```diff
@@ -609,15 +644,52 @@ export class LaunchpadProvider implements Disposable {
...
+				return permanent;
+			}
+			return {
+				items: permanent.items.filter(i => !i.isSearched),
+				timings: permanent.timings,
+			};
+		}
+
+		const result = [];
+		const ids1 = new Set();
```
axosoft-ramint, line 662: What is this set representing? Can we get a clearer name for it?

src/plus/integrations/providers/github.ts
```diff
@@ -317,3 +318,53 @@ export class GitHubEnterpriseIntegration extends GitHubIntegrationBase<SelfHoste
...
+		if (match != null) {
+			prNumber = match[1];
+		}
+	}
+
+	return { ownerAndRepo: ownerAndRepo, prNumber: prNumber };
+}
+
+export function doesPullRequestSatisfyGitHubRepositoryURLIdentity(
+	pr: LaunchpadPullRequest | undefined,
```
axosoft-ramint, line 355: Including Launchpad-specific types here muddles the purity and intent of the provider - instead, since this appears to be a utility fn, we can probably move it to `pullRequest.ts`, and either now or in the near future, make the parameters less GitHub-specific.

src/plus/integrations/providers/github.ts
```diff
@@ -317,3 +318,53 @@ export class GitHubEnterpriseIntegration extends GitHubIntegrationBase<SelfHoste
...
+
+	if (prNumber == null) {
+		match = search.match(/(?:#|\/)(\d+)/); // any number starting with # or /
+		if (match != null) {
+			prNumber = match[1];
+		}
+	}
+
+	return { ownerAndRepo: ownerAndRepo, prNumber: prNumber };
+}
```
axosoft-ramint, line 352: Same as with the function below, I suggest moving these to `pullRequest.ts` and, as soon as we start considering other providers, making them more generic/flexible.

Perhaps then the GitHub-specific regexes we are using can stay in GitHub-domain files like this one, and be imported there.

axosoft-ramint (commented):
I didn't manage to get through all the logic in `launchpad.ts` yet, so these comments are not yet complete, but cover all the other changes. Mostly just nitpicks and questions at this point - will test more thoroughly on Monday.

src/system/function.ts
sergeibbb: @axosoft-ramint 

This piece of code is removed because it did nothing. It didn't break thing, though. Looks like there used to be a special edging case handling that have been removed later. And this code looks like a piece of logic that just hasn't been removed but without special execution for edging case it does nothing.

src/system/vscode/asyncDebouncer.ts
sergeibbb: @axosoft-ramint 

I wanted to do this. But `src/system/function.ts` is supposed to be used both in vscode and in web-view. While `asyncDebouncer` imports `CancellationTokenSource` from `vscode`, so placing it to `function.ts` makes the project non-compilable.

Therefore I've put it under `src/system/vscode`.

src/plus/launchpad/launchpadProvider.ts
sergeibbb: fixed: https://github.com/gitkraken/vscode-gitlens/pull/3684/commits/d44f3443cb11d938c01a71a65e24c20be09db999

src/plus/integrations/providers/github.ts
sergeibbb: @axosoft-ramint 

Fixed. And it's possible to use `EnrichablePullRequest` here instead of `LaunchpadPullRequest`.

src/plus/integrations/providers/github.ts
sergeibbb: @axosoft-ramint 

fixed

src/system/vscode/asyncDebouncer.ts
axosoft-ramint: I see. Thanks for explaining 👍🏼

axosoft-ramint:
Accidentally requested my own review on this PR. I removed it - feel free to re-request my review whenever it's ready to go.

sergeibbb:
Hi @axosoft-ramint 

I think it's ready to review.

CHANGELOG.md
```diff
@@ -34,6 +34,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 - Adds a new `${authorFirst}` and `${authorLast}` commit formatting tokens that can be used in inline blame, commit hovers, etc &mdash; closes [#2980](https://github.com/gitkraken/vscode-gitlens/issues/2980)
 - Adds a new _Create New Branch_ button to the _Commit Graph_ toolbar &mdash; closes [#3553](https://github.com/gitkraken/vscode-gitlens/issues/3553)
 - Adds new ability to force push from the _Commit Graph_ toolbar&mdash; closes [#3493](https://github.com/gitkraken/vscode-gitlens/issues/3493)
+- Adds new ability to search for a GitHub PR in the _Launchpad;_ closes [#3543](https://github.com/gitkraken/vscode-gitlens/issues/3543)
```
axosoft-ramint, line 37: This changelog line should go in the "Unreleased" section of the changelog. If you need to add an "Added" subheading there, you can.

src/plus/launchpad/launchpad.ts
```diff
@@ -493,30 +513,51 @@ export class LaunchpadCommand extends QuickCommand<State> {
...
 
 			return {
-				placeholder: 'Choose an item to focus on',
-				items: getItems(context.result),
+				placeholder: 'Choose an item, type a term to search, or paste in a PR URL',
+				items: getItems(context.result, isSearching),
 			};
 		}
 
+		const mergeQuickpickItems = <T extends { item: { id: string } } | object>(arr: readonly T[], items: T[]) => {
```
axosoft-ramint, line 521: Nitpick: let's call it `combineQuickpickItemsWithSearchResults` since 1. it's clearer and 2. we want to avoid using `merge` because it's associated elsewhere in this file with merging PR branches. And `arr` param can be `searchResults`.

Also, we discussed this elsewhere but, it would be great if the existing launchpad items can stay at the top, and the search results can be added below them. Currently they get mixed in with the results:

search mixup

@eamodio not sure how we can get around it as it feels like we can't really control that ordering when you have input in the box. Unless you have any suggestions I don't think we should spent too much time on it as there are other priorities for this release.

axosoft-ramint (commented):
This is getting close! I found one issue while testing: on the PR items that are search results, we should not have "pin" and "snooze" buttons/options, neither in this stage nor in the next stage after choosing the item (you will need to rebase this on `main` to see the "pin" and "snooze" options in the next step since that is new).

Currently, clicking "pin" or "snooze" will do effectively nothing, as we cannot show categories until the search is empty, nor should those search results persist. I suggest hiding all pin and snooze options on launchpad items that are search results.

src/plus/launchpad/launchpad.ts
sergeibbb: > let's call it combineQuickpickItemsWithSearchResults

Fixed.

> Currently they get mixed in with the results

Currently VS Code does not give us a way to control it.

CHANGELOG.md
sergeibbb: @axosoft-ramint 
My bad. I just didn't expected to meet a fresh "Unreleased" section. 

Fixed.

sergeibbb:
@axosoft-ramint 

> on the PR items that are search results, we should not have "pin" and "snooze" buttons/options, neither in this stage nor in the next stage after choosing the item (you will need to rebase this on main to see the "pin" and "snooze" options in the next step since that is new).

1. Do you mean the state of searching? Such as removing "pin" and "snooze" from **every item.**
2. Or **only items that are "additional"** so they don't belong to the current Launchpad set of PRs?

If # 1. 

What if I want to pin an item but I don't see it in the launchpad, so I start searching it, then when it becomes clearly visible I click "pin". So, it becomes pinned.

> Currently, clicking "pin" or "snooze" will do effectively nothing, as we cannot show categories until the search is empty, nor should those search results persist

Yes. We don't see the effect immediately, but the item has been affected, so in the non-searching state of the Launchpad we'll see the result of our action.

If # 2.
Probably the main problem is that we get "involves:@me" PRs from the GitHub. So, even if I've pinned a PR, so its ID is stored on GK.dev as "pinned", I won't see it in the Launchpad, because it's outside of the set of PRs where I'm involved.

However, it would make sense if we let users pin any PR.

But now I agree, it makes sense to remove these action from "additional" items that have been found.

Thoughts?
cc @eamodio @justinrobots

axosoft-ramint (approved):
LGTM

Issue #3621, opened 2024-09-27
Start Work operation in Home view


justinrobots:
[Figma Flow Diagram](https://www.figma.com/board/einSHVwbXU9NeHOnnY0NMT/GitLens%3A-Start-Work-Flow?node-id=0-1&t=stWxxPzS1QMFUONU-1)

Add a "Start work" button to the Home view that opens a quick pick to initiate different flows for starting work (creating a branch).

Users should be able to create a branch, create a branch on a worktree (pro), and do either of those from an issue on a connected service (e.g, create a branch from a Jira issue).

New branches associated with issues will follow whatever naming conventions the service (e.g, Jira) requires to associate them.
- if users have specific autolinks setup to make these associations, we may want to consider those when creating branches for issues.

## Roadmap

- ✅ collect GitHub issues
- ✅ if we don't have an issue integration, suggest user to connect it. (can we have many issue providers? I believe, yes).
<img src="https://github.com/user-attachments/assets/2308f7b4-c036-42fa-8a99-96f3040e12ea" width="160"> <img src="https://github.com/user-attachments/assets/f666bafe-ef9e-4663-97b6-befe19d4c724" width="500">

- ✅ build compatible branch name based on issue title and id
  <img src="https://github.com/user-attachments/assets/8a6a230f-ee76-44b8-aed5-008622b240d0" width="400">

- ✅ Remove "read more about start work" section from "connect integrations" step:
  <img src="https://github.com/user-attachments/assets/84e36953-3650-437d-b066-9f68ba1eb6ed" width="600">

- ✅ when user selects a GitHub issue, highlight the corresponding repository on the "select repo" step

- ❌ ~~feature gate:~~
  <img src="https://github.com/user-attachments/assets/ecc4b797-4e69-4c17-adb0-2baaaf90373d" width="400">

- ✅ let user create a branch without selecting an issue (just send them to "create-branch" flow)
  <img src="https://github.com/user-attachments/assets/52062d8e-a4f4-4854-a14a-0b1ebabe313d" width="400">

- ✅ Jira issues
[...]

PR #3698, opened 2024-10-25
PR on Start Work operation in Home view (#3621)

sergeibbb:
# Description

This PR introduces initial Start Work flow that:
- adds a "Start Work" shortcut to the home view
- lets user to select a GitHub issue and then starts a create-branch flow
- suggests the repository of the selected issue as default repository for creating a branch
- suggests a default branch name based on issue title

# Checklist

### DIFF
--- a/src/commands/git/branch.ts
+++ b/src/commands/git/branch.ts
@@ -63,6 +63,11 @@ interface CreateState {
 	flags: CreateFlags[];
 
 	suggestNameOnly?: boolean;
+	suggestRepoOnly?: boolean;
+}
+
+function isCreateState(state: Partial<State> | undefined): state is Partial<CreateState> {
+	return state?.subcommand === 'create';
 }
 
 type DeleteFlags = '--force' | '--remotes';
@@ -172,6 +177,10 @@ export class BranchGitCommand extends QuickCommand {
 						counter++;
 					}
 
+					if (args.state.suggestRepoOnly && args.state.repo != null) {
+						counter--;
+					}
+
 					break;
 				case 'delete':
 				case 'prune':
@@ -251,7 +260,12 @@ export class BranchGitCommand extends QuickCommand {
 				state.subcommand,
 			);
 
-			if (state.counter < 2 || state.repo == null || typeof state.repo === 'string') {
+			if (
+				state.counter < 2 ||
+				state.repo == null ||
+				typeof state.repo === 'string' ||
+				(isCreateState(state) && state.suggestRepoOnly)
+			) {
 				skippedStepTwo = false;
 				if (context.repos.length === 1) {
 					skippedStepTwo = true;
--- a/src/commands/git/branch.ts
+++ b/src/commands/git/branch.ts
@@ -373,7 +373,7 @@ export class BranchGitCommand extends QuickCommand {
 				state.reference = result;
 			}
 
-			if (state.counter < 4 || state.name == null) {
+			if (state.counter < 4 || state.name == null || state.suggestNameOnly) {
 				const result = yield* inputBranchNameStep(state, context, {
 					titleContext: ` from ${getReferenceLabel(state.reference, {
 						capitalize: true,
--- a/src/commands/quickWizard.utils.ts
+++ b/src/commands/quickWizard.utils.ts
@@ -1,6 +1,7 @@
 import type { StoredRecentUsage } from '../constants.storage';
 import type { Container } from '../container';
 import { LaunchpadCommand } from '../plus/launchpad/launchpad';
+import { StartWorkCommand } from '../plus/startWork/startWork';
 import { configuration } from '../system/vscode/configuration';
 import { getContext } from '../system/vscode/context';
 import { BranchGitCommand } from './git/branch';
@@ -111,6 +112,10 @@ export class QuickWizardRootStep implements QuickPickStep<QuickCommand> {
 		this.hiddenItems = [];
 		if (args?.command === 'launchpad') {
 			this.hiddenItems.push(new LaunchpadCommand(container, args));
+		}
+
+		if (args?.command === 'startWork') {
+			this.hiddenItems.push(new StartWorkCommand(container));
 		}
 	}
 
--- a/src/plus/startWork/startWork.ts
+++ b/src/plus/startWork/startWork.ts
@@ -16,6 +16,7 @@ import {
 	QuickCommand,
 	StepResultBreak,
 } from '../../commands/quickCommand';
+import { getSteps } from '../../commands/quickWizard.utils';
 import { proBadge } from '../../constants';
 import type { IntegrationId } from '../../constants.integrations';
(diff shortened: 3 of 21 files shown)


Meganbrentshandacrain:
> # Description
> This PR introduces initial Start Work flow that:
> 
> * adds a "Start Work" shortcut to the home view
> * lets user to select a GitHub issue and then starts a create-branch flow
> * suggests the repository of the selected issue as default repository for creating a branch
> * suggests a default branch name based on issue title
> 
> # Checklist
> * [x]  I have followed the guidelines in the Contributing document
> * [x]  My changes follow the coding style of this project
> * [x]  My changes build without any errors or warnings
> * [x]  My changes have been formatted and linted
> * [ ]  My changes include any required corresponding changes to the documentation (including CHANGELOG.md and README.md)
> * [ ]  My changes have been rebased and squashed to the minimal number (typically 1) of relevant commits
> * [x]  My changes have a descriptive commit message with a short title, including a `Fixes $XXX -` or `Closes #XXX -` prefix to auto-close the issue that your PR addresses

Megan Brianna CrainGracie adalynn Donald Joseph chiasson iv Y Brent crain  and shanda spears 27487 benard crain rd Angie la 70426_3391

src/commands/quickWizard.utils.ts
Meganbrentshandacrain, line 1: src/commands/quickWizard.ts

src/plus/startWork/startWork.ts
```diff
@@ -0,0 +1,462 @@
...
+		);
+
+		return connected;
+	}
+}
+
+async function updateContextItems(container: Container, context: Context) {
+	context.result = {
+		items:
+			(await container.integrations.getMyIssues([HostingIntegrationId.GitHub]))?.map(i => ({
```
axosoft-ramint, line 458: Perhaps we can do one of the following:
1. Create a `startWorkProvider`
2. Rename the `launchpadProvider` to be a bit more flexible outside the domain of Launchpad itself and extend it

And then we can add functions to the provider to:
1. Get searched issues, with caching and error handling like we do for Launchpad items, and the ability to filter out stale ones and other post-processing like we do with PRs for Launchpad (not that we have to implement any post-processing now, just the caching and error handling parts)
2. Complete actions on "start work" items, like creating a branch, etc.
3. Set up other utilities that we want in the future for Start Work so that the command file doesn't get out of hand.
4. Use those functions here as needed, like `launchpad` does with `launchpadProvider`.

axosoft-ramint (commented):
Mostly nitpicks - the only major suggestion would be to set up a provider that manages the issue querying and includes caching and error handling as we have with Launchpad. Most of this would be for future-proofing purposes - we don't automatically query for issues with this feature like we do for Launchpad - but the need for that may arise sooner than we think (for example - some UX on the Home view showing a count of how many issues we have, etc.)

sergeibbb:
@axosoft-ramint @ericf-axosoft 

Please review it one more time. All notes related to implementing a provider haven't been fixed yet, because I've been focusing on the functionality and Eric has suggested to implement optimizations as a follow-up.

So, now you're welcome to suggest any urgent fixes that I can make on Monday.

sergeibbb:
@ericf-axosoft 

> I assume you mean @eamodio instead of me, right?

True. Sorry for bothering

src/commands/git/branch.ts
```diff
@@ -251,7 +264,12 @@ export class BranchGitCommand extends QuickCommand {
 				state.subcommand,
 			);
 
-			if (state.counter < 2 || state.repo == null || typeof state.repo === 'string') {
+			if (
+				state.counter < 2 ||
+				state.repo == null ||
+				typeof state.repo === 'string' ||
+				castCreateState(state)?.suggestRepoOnly
```
axosoft-ramint, line 271: Rather than using a casting function, what do you think about just asserting first and then using the property:
```suggestion
				(isCreateState(state) && state.suggestRepoOnly)
```

src/plus/startWork/startWork.ts
```diff
@@ -0,0 +1,452 @@
...
+				connected.set(integrationId, hasAccess);
+			}),
+		);
+
+		return connected;
+	}
+}
+
+async function updateContextItems(container: Container, context: Context) {
+	const conneectedIntegrationMap = context.connectedIntegrations;
```
axosoft-ramint, line 442:
```suggestion
	const connectedIntegrationsMap = context.connectedIntegrations;
```

axosoft-ramint (changes requested):
I'm noticing a few issues/bugs when testing:

1. If I don't have integrations connected, when using the command, it immediately shows the "connect integration" step. This step should only show after I choose one of the "from an issue" options in the initial step.

image

2. The issue description/body is not showing in the hover tooltip of each issue. We used to have this at the "confirm" step where you could choose an action on that issue, but since that step is gone now, there is no way to read the issue description without opening the issue. We should have the current description, and then a blank line, and then the issue body, so that the current description lines show as they are now in the quickpick but if you hover over one of them, you can see the full issue body in the tooltip.

image

3. There is a bug it seems where the command disappears/is no longer available after choosing to open the worktree we created in a new window:

start work command disappears

4. For Jira, I'm only seeing issues that I created/authored. It should show, and prioritize, issues I am assignee on, [like this one for example](https://gitkraken.atlassian.net/browse/GLVSC-427). Instead I am seeing issues I reported even if they have no assignee.

5. When clicking the "globe" icon on an issue to open that issue in the browser, we should not hide the quickpick. I'm seeing the same issue with Launchpad items as well, so perhaps there is a regression here or some issue with the debug window? (in the pre-release, if I click the globe to open a Launchpad PR in the browser, the quickpick stays open).

sergeibbb:
@axosoft-ramint 

> There is a bug it seems where the command disappears/is no longer available after choosing to open the worktree we created in a new window:

I think it's a problem with VS Code debugging. If you launched the current code in dev mode and open a worktree in a new window, the new window is not the new code anymore. It runs with your current version of GitLens.

I faced this problem even when I open a blank space and "open a folder". To avoid this I've been adding a folder to the current workspace.

src/plus/startWork/startWork.ts
sergeibbb: fixed

src/commands/git/branch.ts
sergeibbb: fixed

sergeibbb:
@axosoft-ramint 

> 5. (in the pre-release, if I click the globe to open a Launchpad PR in the browser, the quickpick stays open).

disappears for me in pre-releas. It disappears every time when the window looses the focus, but it stays if (and after) the quickpick has a text in the search field.

https://github.com/user-attachments/assets/e54e8152-7184-4475-b52f-f0cbc1891e89

axosoft-ramint (approved):
Some of the issues from previous review may be due to debugging or pre-existing regressions. We will address those and any other unaddressed issues above post-merge.

PR #3701, opened 2024-10-28
Opens all changes after open in worktree flow finished for PRs

nzaytsev:
# Description

https://github.com/user-attachments/assets/a4ebac45-b003-47de-9e44-4ad742b697b9

# Checklist

### DIFF
--- a/src/commands/git/worktree.ts
+++ b/src/commands/git/worktree.ts
@@ -102,7 +102,7 @@ interface CreateState {
 		title?: string;
 	};
 
-	onWorkspaceChanging?: (() => Promise<void>) | (() => void);
+	onWorkspaceChanging?: ((isNewWorktree?: boolean) => Promise<void>) | ((isNewWorktree?: boolean) => void);
 	skipWorktreeConfirmations?: boolean;
 }
 
@@ -139,7 +139,8 @@ interface OpenState {
 		};
 	};
 
-	onWorkspaceChanging?: (() => Promise<void>) | (() => void);
+	onWorkspaceChanging?: ((isNewWorktree?: boolean) => Promise<void>) | ((isNewWorktree?: boolean) => void);
+	isNewWorktree?: boolean;
 	skipWorktreeConfirmations?: boolean;
 }
 
@@ -628,6 +629,7 @@ export class WorktreeGitCommand extends QuickCommand<State> {
 						confirm: action === 'prompt',
 						openOnly: true,
 						overrides: { disallowBack: true },
+						isNewWorktree: true,
 						skipWorktreeConfirmations: state.skipWorktreeConfirmations,
 						onWorkspaceChanging: state.onWorkspaceChanging,
 					} satisfies OpenStepState,
@@ -1081,7 +1083,7 @@ export class WorktreeGitCommand extends QuickCommand<State> {
 
 				const location = convertOpenFlagsToLocation(state.flags);
 				if (location === 'currentWindow' || location === 'newWindow') {
-					await state.onWorkspaceChanging?.();
+					await state.onWorkspaceChanging?.(state.isNewWorktree);
 				}
 
 				openWorkspace(state.worktree.uri, { location: convertOpenFlagsToLocation(state.flags), name: name });
--- a/src/plus/launchpad/launchpadProvider.ts
+++ b/src/plus/launchpad/launchpadProvider.ts
@@ -540,7 +540,13 @@ export class LaunchpadProvider implements Disposable {
 				? item.openRepository.localBranch.name
 				: item.headRef.name;
 
-		return getPullRequestBranchDeepLink(this.container, branchName, item.repoIdentity.remote.url, action);
+		return getPullRequestBranchDeepLink(
+			this.container,
+			branchName,
+			item.repoIdentity.remote.url,
+			action,
+			item.underlyingPullRequest,
+		);
 	}
 
 	private async getMatchingOpenRepository(
@@ -1023,17 +1029,31 @@ export function getPullRequestBranchDeepLink(
 	headRefBranchName: string,
 	remoteUrl: string,
 	action?: DeepLinkActionType,
+	pr?: PullRequest,
 ) {
 	const schemeOverride = configuration.get('deepLinks.schemeOverride');
 	const scheme = typeof schemeOverride === 'string' ? schemeOverride : env.uriScheme;
+
+	const searchParams = new URLSearchParams({
+		url: ensureRemoteUrl(remoteUrl),
+	});
+	if (action) {
+		searchParams.set('action', action);
+	}
+	if (pr) {
+		searchParams.set('prId', pr.id);
+		searchParams.set('prTitle', pr.title);
+	}
+	if (pr?.refs) {
+		searchParams.set('prBaseRef', pr.refs.base.sha);
+		searchParams.set('prHeadRef', pr.refs.head.sha);
+	}
 	// TODO: Get the proper pull URL from the provider, rather than tacking .git at the end of the
 	// url from the head ref.
 	return Uri.parse(
 		`${scheme}://${container.context.extension.id}/${'link' satisfies UriTypes}/${DeepLinkType.Repository}/-/${
 			DeepLinkType.Branch
-		}/${encodeURIComponent(headRefBranchName)}?url=${encodeURIComponent(ensureRemoteUrl(remoteUrl))}${
-			action != null ? `&action=${action}` : ''
-		}`,
+		}/${encodeURIComponent(headRefBranchName)}?${searchParams.toString()}`,
 	);
 }
 
--- a/src/uris/deepLinks/deepLink.ts
+++ b/src/uris/deepLinks/deepLink.ts
@@ -391,7 +391,6 @@ export const deepLinkStateTransitionTable: Record<string, Record<string, DeepLin
 		[DeepLinkServiceAction.OpenGraph]: DeepLinkServiceState.OpenGraph,
 		[DeepLinkServiceAction.OpenFile]: DeepLinkServiceState.OpenFile,
 		[DeepLinkServiceAction.OpenSwitch]: DeepLinkServiceState.SwitchToRef,
-		[DeepLinkServiceAction.OpenAllPrChanges]: DeepLinkServiceState.OpenAllPrChanges,
 		[DeepLinkServiceAction.OpenComparison]: DeepLinkServiceState.OpenComparison,
 		[DeepLinkServiceAction.DeepLinkErrored]: DeepLinkServiceState.Idle,
 		[DeepLinkServiceAction.DeepLinkCancelled]: DeepLinkServiceState.Idle,
--- a/src/uris/deepLinks/deepLink.ts
+++ b/src/uris/deepLinks/deepLink.ts
@@ -100,6 +100,7 @@ export interface DeepLink {
 	secondaryTargetId?: string;
 	secondaryRemoteUrl?: string;
 	action?: string;
+	prId?: string;
 	params?: URLSearchParams;
 }
 
@@ -178,6 +179,7 @@ export function parseDeepLinkUri(uri: Uri): DeepLink | undefined {
 				secondaryRemoteUrl: secondaryRemoteUrl,
 				action: action,
 				params: urlParams,
+				prId: urlParams.get('prId') ?? undefined,
 			};
 		}
 		case DeepLinkType.Draft: {
@@ -239,6 +241,7 @@ export const enum DeepLinkServiceState {
 	OpenInspect,
 	SwitchToRef,
 	RunCommand,
+	OpenAllPrChanges,
 }
 
 export const enum DeepLinkServiceAction {
@@ -271,6 +274,7 @@ export const enum DeepLinkServiceAction {
 	OpenFile,
 	OpenInspect,
 	OpenSwitch,
+	OpenAllPrChanges,
 }
 
 export type DeepLinkRepoOpenType = 'clone' | 'folder' | 'workspace' | 'current';
@@ -387,6 +391,7 @@ export const deepLinkStateTransitionTable: Record<string, Record<string, DeepLin
 		[DeepLinkServiceAction.OpenGraph]: DeepLinkServiceState.OpenGraph,
 		[DeepLinkServiceAction.OpenFile]: DeepLinkServiceState.OpenFile,
 		[DeepLinkServiceAction.OpenSwitch]: DeepLinkServiceState.SwitchToRef,
+		[DeepLinkServiceAction.OpenAllPrChanges]: DeepLinkServiceState.OpenAllPrChanges,
 		[DeepLinkServiceAction.OpenComparison]: DeepLinkServiceState.OpenComparison,
 		[DeepLinkServiceAction.DeepLinkErrored]: DeepLinkServiceState.Idle,
 		[DeepLinkServiceAction.DeepLinkCancelled]: DeepLinkServiceState.Idle,
@@ -417,6 +422,12 @@ export const deepLinkStateTransitionTable: Record<string, Record<string, DeepLin
 		[DeepLinkServiceAction.DeepLinkCancelled]: DeepLinkServiceState.Idle,
 	},
 	[DeepLinkServiceState.OpenInspect]: {
+		[DeepLinkServiceAction.OpenAllPrChanges]: DeepLinkServiceState.OpenAllPrChanges,
+		[DeepLinkServiceAction.DeepLinkResolved]: DeepLinkServiceState.Idle,
+		[DeepLinkServiceAction.DeepLinkErrored]: DeepLinkServiceState.Idle,
+		[DeepLinkServiceAction.DeepLinkCancelled]: DeepLinkServiceState.Idle,
+	},
+	[DeepLinkServiceState.OpenAllPrChanges]: {
 		[DeepLinkServiceAction.DeepLinkResolved]: DeepLinkServiceState.Idle,
 		[DeepLinkServiceAction.DeepLinkErrored]: DeepLinkServiceState.Idle,
 		[DeepLinkServiceAction.DeepLinkCancelled]: DeepLinkServiceState.Idle,
--- a/src/uris/deepLinks/deepLinkService.ts
+++ b/src/uris/deepLinks/deepLinkService.ts
@@ -241,8 +241,9 @@ export class DeepLinkService implements Disposable {
 
 	@debug()
 	private async processPendingDeepLink(pendingDeepLink: StoredDeepLinkContext | undefined) {
-		if (pendingDeepLink?.url == null) return;
+		if (pendingDeepLink == null) return;
 		void this.container.storage.delete('deepLinks:pending');
+		if (pendingDeepLink?.url == null) return;
 		const link = parseDeepLinkUri(Uri.parse(pendingDeepLink.url));
 		if (link == null) return;
 
--- a/src/uris/deepLinks/deepLinkService.ts
+++ b/src/uris/deepLinks/deepLinkService.ts
@@ -1321,9 +1321,25 @@ export class DeepLinkService implements Disposable {
 							state: this._context.state,
 						};
 
+						// Form a new link URL with PR info stripped out in case we are opening an existing PR worktree,
+						// in which case we do not want to advance to the  "Open All PR Changes" step in the flow.
+						// We should only advance to that step if a worktree is newly created in the flow.
+						const oldUrl = Uri.parse(this._context.url ?? '').toString(true);
+						const urlParams = new URL(oldUrl).searchParams;
+						urlParams.delete('prId');
+						urlParams.delete('prTitle');
+						urlParams.delete('prBaseRef');
+						urlParams.delete('prHeadRef');
+						const newUrlParams = urlParams.toString();
+						const nonPrUrl =
+							newUrlParams.length > 0 ? `${oldUrl.split('?')[0]}?${newUrlParams}` : oldUrl.split('?')[0];
+
 						// Storing link info in case the switch causes a new window to open
-						const onWorkspaceChanging = async () =>
-							this.container.storage.store('deepLinks:pending', pendingDeepLink);
+						const onWorkspaceChanging = async (isNewWorktree?: boolean) =>
+							this.container.storage.store(
+								'deepLinks:pending',
+								isNewWorktree ? pendingDeepLink : { ...pendingDeepLink, url: nonPrUrl },
+							);
 
 						await executeGitCommand({
 							command: 'switch',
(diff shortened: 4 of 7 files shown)


package.json
```diff
@@ -17924,7 +17924,7 @@
 		"lint:fix": "pnpm run clean:lint && eslint . --fix",
 		"lint:webviews": "pnpm run clean:lint && eslint \"src/webviews/apps/**/*.ts?(x)\"",
 		"package": "vsce package --no-dependencies",
-		"package-pre": "pnpm run patch-pre && pnpm run package --pre-release",
+		"//package-pre": "pnpm run patch-pre && pnpm run package --pre-release",
```
axosoft-ramint, line 17927: This change should not be in this PR

package.json
```diff
@@ -17946,7 +17946,7 @@
 		"update-emoji": "node ./scripts/generateEmojiShortcodeMap.mjs",
 		"update-licenses": "node ./scripts/generateLicenses.mjs",
 		"pretest": "pnpm run build:tests",
-		"vscode:prepublish": "pnpm run bundle"
+		"//vscode:prepublish": "pnpm run bundle"
```
axosoft-ramint, line 17949: Same here

axosoft-ramint:
Do we have an issue that this PR closes? I want to make sure I understand the requirements completely, because it's possible that these changes will cause other PR workflows like "switch" actions from Launchpad to also open the PR changes (which sounds like what we would want, but it is outside of the context of worktrees as the PR title describes).

src/commands/git/worktree.ts
```diff
@@ -102,6 +102,8 @@ interface CreateState {
 		title?: string;
 	};
 
+	prId?: string;
```
axosoft-ramint, line 105: Where is this property being used? I see we are attaching `prId` to the "create" and "open" states but I'm having trouble finding where we use it. From what I can tell we only use it in the deep link service to open the multi-diff, but that is from the url itself, not the `prId` being set here.

src/git/actions/worktree.ts
```diff
@@ -56,11 +56,15 @@ export function copyChangesToWorktree(
...
-export function open(worktree: GitWorktree, options?: { location?: OpenWorkspaceLocation; openOnly?: boolean }) {
+export function open(
+	worktree: GitWorktree,
+	options?: { location?: OpenWorkspaceLocation; openOnly?: boolean; prId?: string },
+) {
 	return executeGitCommand({
 		command: 'worktree',
 		state: {
 			subcommand: 'open',
+			prId: options?.prId,
```
axosoft-ramint, line 67: This is set here, but not sure where it is used in the command flow (see prev. comment)

src/plus/launchpad/launchpadProvider.ts
```diff
@@ -1033,7 +1034,7 @@ export function getPullRequestBranchDeepLink(
 			DeepLinkType.Branch
 		}/${encodeURIComponent(headRefBranchName)}?url=${encodeURIComponent(ensureRemoteUrl(remoteUrl))}${
 			action != null ? `&action=${action}` : ''
-		}`,
+		}${prId != null ? `&prId=${prId}` : ''}`,
```
axosoft-ramint, line 1037: I'm not sure if a PR id can ever have characters that would need to be encoded, but just in case, I suggest using `encodeURIComponent` on it.

src/uris/deepLinks/deepLinkService.ts
```diff
@@ -241,8 +242,10 @@ export class DeepLinkService implements Disposable {
 		}
 	}
 
-	private async processPendingDeepLink(pendingDeepLink: StoredDeepLinkContext) {
-		if (pendingDeepLink.url == null) return;
+	@debug()
+	private async processPendingDeepLink(pendingDeepLink: StoredDeepLinkContext | undefined) {
+		if (pendingDeepLink?.url == null) return;
+		void this.container.storage.delete('deepLinks:pending');
```
axosoft-ramint, line 248: I don't mind that we moved the storage deletion here, and a pending deep link in storage _should_ always have a url, so this is probably not a big deal, but to protect ourselves, I suggest deleting the stored value regardless of whether it has a `url` property or not. That would be the case of bad data in storage, but we would want that deleted or else it will stick around and kick off this logic on every start of the service.

src/uris/deepLinks/deepLinkService.ts
```diff
@@ -312,6 +315,19 @@ export class DeepLinkService implements Disposable {
...
+		const pullRequests = remote
+			? await this.container.integrations.getMyPullRequestsForRemotes(remote)
+			: await this.container.integrations.getMyPullRequests();
+
+		if (!pullRequests?.value) {
+			return;
+		}
+		return pullRequests.value.find(x => x.pullRequest.id === targetId)?.pullRequest;
+	}
+
```
axosoft-ramint, line 330: Since this is only for the deep link service, and since the PR ID lives in the context, I suggest just making this a `getPullRequestFromContext`, remove the `targetId` parameter, and just pull the `prId` directly from context as we are already doing for `repo` and `remote`. And if it is not defined, return `undefined` like we are already doing here for `repo`.

Furthermore, `getMyPullRequests` with no input parameters sounds like potentially a LOT of work for the small task of opening a PR's changes. Is there any reason we would arrive in this function without a defined `remote`?

src/uris/deepLinks/deepLinkService.ts
```diff
@@ -1386,6 +1402,25 @@ export class DeepLinkService implements Disposable {
...
+					if (this._context.prId) {
+						action = DeepLinkServiceAction.OpenAllPrChanges;
+					} else {
+						action = DeepLinkServiceAction.DeepLinkResolved;
+					}
+					break;
+				}
+				case DeepLinkServiceState.OpenAllPrChanges: {
+					const pr = await this.getPullRequest(this._context.prId!);
+					const refs = getComparisonRefsForPullRequest(this._context.repoPath!, pr!.refs!);
```
axosoft-ramint, line 1414: We should not be forcing assertions on the existence of these values. Following the pattern we use in other states, we should check the existence of the values and properties we need, and if we cannot find them, kick out into the error state with a message.

axosoft-ramint (changes requested):

package.json
nzaytsev: yep, my bad. it's all because of pnpm issues on MacOS

src/uris/deepLinks/deepLinkService.ts
nzaytsev: done

src/plus/launchpad/launchpadProvider.ts
nzaytsev: I suggest using native URLSearchParams class as more readable and correct way to add search parameters to the url string

src/uris/deepLinks/deepLinkService.ts
nzaytsev: done

src/plus/launchpad/launchpadProvider.ts
nzaytsev: It looks like the link generation should be moved on deeplinkink service side. I would also make the `UriTypes` enum instead of using satisfies

axosoft-ramint: Both of those suggestions sound good, thanks!

src/uris/deepLinks/deepLinkService.ts
nzaytsev: made it shorter, so I think there is no need to keep it as a separate method

src/commands/git/worktree.ts
nzaytsev: removed

src/git/actions/worktree.ts
nzaytsev: removed

src/uris/deepLinks/deepLinkService.ts
axosoft-ramint: Follow-up after the change: we should not try to delete from storage every time, only if we retrieved something from storage in line 73. That's why the original was encased in an `if (pendingDeepLink != null) {` call. I suggest encasing this line in the same `if` check, and then proceed to verify that the `url` property exists.

src/uris/deepLinks/deepLinkService.ts
```diff
@@ -1077,7 +1082,11 @@ export class DeepLinkService implements Disposable {
 					return;
 				}
 				case DeepLinkServiceState.GoToTarget: {
-					// Need to re-fetch the remotes in case we opened in a new window
+					// Re-fetching the remotes in case we opened in a new window for specific actions
+					if (this._context.action === DeepLinkActionType.OpenAllPrChanges) {
+						action = DeepLinkServiceAction.OpenAllPrChanges;
+						break;
+					}
```
axosoft-ramint, line 1089: Question here: If we skip straight to opening PR changes, wouldn't it fail to open the inspect view?

Logically, it should always open inspect regardless if a new window was opened or not, and then the action of opening PR changes can follow as a next step.

I would avoid modifying or using `this._context.action` directly because it uses that property outside of its intended purpose.

src/plus/launchpad/launchpadProvider.ts
```diff
@@ -1023,17 +1022,27 @@ export function getPullRequestBranchDeepLink(
...
+	}
+	if (prId) {
+		searchParams.set('prId', prId);
+	}
+
 	// TODO: Get the proper pull URL from the provider, rather than tacking .git at the end of the
 	// url from the head ref.
 	return Uri.parse(
-		`${scheme}://${container.context.extension.id}/${'link' satisfies UriTypes}/${DeepLinkType.Repository}/-/${
+		`${scheme}://${container.context.extension.id}/${UriTypes.link}/${DeepLinkType.Repository}/-/${
```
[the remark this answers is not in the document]
nzaytsev: The question is: should I revert the changes I've made?

IMO, to not increase the bundle size, all enums should be placed in separate filed from logical code. Just enum importing must not be to expensive. But it may be inconvenient from the code structure side

src/commands/git/worktree.ts
```diff
@@ -128,6 +128,8 @@ interface OpenState {
 	worktree: GitWorktree;
 	flags: OpenFlags[];
 
+	prId?: string;
+
```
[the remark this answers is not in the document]
nzaytsev: removed

src/plus/launchpad/launchpadProvider.ts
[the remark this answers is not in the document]
nzaytsev: reverted

src/uris/deepLinks/deepLink.ts
```diff
@@ -5,7 +5,9 @@ import type { GitRemote } from '../../git/models/remote';
 import type { Repository } from '../../git/models/repository';
 import type { OpenWorkspaceLocation } from '../../system/vscode/utils';
 
-export type UriTypes = 'link';
+export enum UriTypes {
+	link = 'link',
+}
```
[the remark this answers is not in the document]
nzaytsev: reverted

src/uris/deepLinks/deepLinkService.ts
```diff
@@ -1522,9 +1567,7 @@ export class DeepLinkService implements Disposable {
 
 		// Start with the prefix, add the repo prefix and the repo ID to the URL, and then add the target tag and target ID to the URL (if applicable)
 		const deepLink = new URL(
-			`${scheme}://${this.container.context.extension.id}/${'link' satisfies UriTypes}/${
-				DeepLinkType.Repository
-			}/${repoId}${target}`,
+			`${scheme}://${this.container.context.extension.id}/${UriTypes.link}/${DeepLinkType.Repository}/${repoId}${target}`,
```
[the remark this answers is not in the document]
nzaytsev: reverted

src/uris/deepLinks/deepLinkService.ts
```diff
@@ -1575,9 +1618,7 @@ export class DeepLinkService implements Disposable {
 		}
 
 		const deepLink = new URL(
-			`${scheme}://${this.container.context.extension.id}/${'link' satisfies UriTypes}/${
-				DeepLinkType.Repository
-			}/${repoId}/${targetType}/${targetId}`,
+			`${scheme}://${this.container.context.extension.id}/${UriTypes.link}/${DeepLinkType.Repository}/${repoId}/${targetType}/${targetId}`,
```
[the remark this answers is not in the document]
nzaytsev: reverted

src/uris/deepLinks/deepLinkService.ts
nzaytsev: >  If we skip straight to opening PR changes, wouldn't it fail to open the inspect view?

I don't understand your question. Current behavior is exactly how you described: open inspect, than if needed - open all changes (of course through ensuring repo)

> I would avoid modifying or using this._context.action

agree, but there is a new mechanic required. not sure if I'm able to suggest something reliable now, because I'm not well familiar with the deeplinking service. What do you think about a flag (such as `openedInNewWindow`) in the context, then we'll detect the flag and choose the right target. I'll push it now

nzaytsev: or there should be a middle state between ensureRepoMatch and goToTarget (e.g. repoMatchFinished) the goal of that is to redirect to goToTarget or to proceedNewWindowLink (the new one). If you like it, I'll implement. Here is a rude example:

```ts

case DeepLinkServiceState.GoToTarget: {
  /*leave as is*/
}
case DeepLinkServiceState.RepoMatchFinished: {
    if (this._context.isNewWindowOpened) {
       action = DeepLinkServiceAction.ProceedNewWindowLink
    } else {
       action = DeepLinkServiceAction.GoToTarget
    }
}
case DeepLinkServiceState.ProceedNewWindowLink: {
    if (this._context.action === DeepLinkActionType.SwitchToPullRequestWorktree) {
       action = DeepLinkServiceAction.OpenAllPrChanges
       break
    }
    action = DeepLinkServiceAction.DeepLinkResolved
}
```

src/plus/launchpad/launchpadProvider.ts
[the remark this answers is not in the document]
nzaytsev: reverted

src/views/viewCommands.ts
```diff
@@ -808,6 +808,7 @@ export class ViewCommands {
 				pr.refs.head.branch,
 				repoIdentity.remote.url,
 				DeepLinkActionType.SwitchToPullRequestWorktree,
+				pr.id,
```
axosoft-ramint, line 811: We need to also pass PR ID in other places where we use "open in workspace" actions, like from the Launchpad and Launchpad view (`switchTo` of `launchpadProvider`): https://github.com/gitkraken/vscode-gitlens/blob/features/3549-open-multi-diff-editor-when-opening-a-new-worktree-from-a-pr/src/plus/launchpad/launchpadProvider.ts#L470-L471

This explains why I was not able to see the PR changes when I was testing your PR (I was using Launchpad).

src/uris/deepLinks/deepLinkService.ts
```diff
@@ -241,9 +241,10 @@ export class DeepLinkService implements Disposable {
 		}
 	}
 
-	private async processPendingDeepLink(pendingDeepLink: StoredDeepLinkContext) {
-		if (pendingDeepLink.url == null) return;
-
+	@debug()
+	private async processPendingDeepLink(pendingDeepLink: StoredDeepLinkContext | undefined) {
+		if (pendingDeepLink?.url == null) return;
+		void this.container.storage.delete('deepLinks:pending');
```
axosoft-ramint, line 247:
```suggestion
		if (pendingDeepLink == null) return;
		void this.container.storage.delete('deepLinks:pending');
		if (pendingDeepLink?.url == null) return;
```

axosoft-ramint: This is what I meant. We want to delete the stored value whether or not it has the url property, but we don't want to proceed beyond that without the url.

src/uris/deepLinks/deepLinkService.ts
```diff
@@ -1107,12 +1112,17 @@ export class DeepLinkService implements Disposable {
...
+									? DeepLinkServiceAction.OpenAllPrChanges
+									: DeepLinkServiceAction.OpenSwitch;
 							} else {
 								action = DeepLinkServiceAction.OpenGraph;
 							}
 							break;
 					}
+					window.showInformationMessage(
+						`go to target: ${targetType} - ${this._context.action} - ${this._context.openedInNewWindow} -/ ${action}`,
+					);
```
axosoft-ramint, line 1125: This extra notification is unnecessary.

src/uris/deepLinks/deepLinkService.ts
```diff
@@ -1078,15 +1082,16 @@ export class DeepLinkService implements Disposable {
...
-
 					if (targetType === DeepLinkType.Repository) {
 						if (
 							this._context.action === DeepLinkActionType.Switch ||
 							this._context.action === DeepLinkActionType.SwitchToPullRequest ||
 							this._context.action === DeepLinkActionType.SwitchToPullRequestWorktree ||
 							this._context.action === DeepLinkActionType.SwitchToAndSuggestPullRequest
 						) {
-							action = DeepLinkServiceAction.OpenSwitch;
+							action = this._context.openedInNewWindow
```
axosoft-ramint, line 1092: I see now what we're trying to go for with this "opened in new window". The problem is that there are other cases (like "switch to branch" with a non-pr branch) where we would open a new window as a part of the deep link flow. In those cases:

1. If this context was not set, it would make the code confusing, based on how it is named.
2. If this context was set, the deep link would not work because no PR is involved.

I have an idea to make this entire thing a bit easier on ourselves (until I move this processing out of deep links entirely in the future).

To show the PR changes, ultimately we just need the repo path (string), the base ref (string), the head ref (string), and the PR's title (string) and then we run `openComparisonChanges` on that, right?

So why don't we just encode `repoPath`, `pullRequestBaseRef` and `pullRequestHeadRef` into the query of the deep link for the PR in `getPullRequestBranchDeepLink`, and then just use those directly in the "open all PR changes" case? The entire query (`URLSearchParams`) should already be in `this._context.params`.

Then we can just extract them (null check and move to error state if they don't exist) and we wouldn't need to send PR ID, nor would we need to re-fetch the pull request object, nor the remote, etc.

axosoft-ramint (changes requested):
I found out why I wasn't able to see the PR changes in my test cases using the Launchpad. See below.

Also have an overall suggestion for making this much simpler. Let me know if it would work or if we would run into issues.

src/uris/deepLinks/deepLinkService.ts
nzaytsev: Ah, ok

src/uris/deepLinks/deepLinkService.ts
nzaytsev: I left it here for you to see why couldn't you test it. I'll remove

src/uris/deepLinks/deepLink.ts
```diff
@@ -387,6 +391,7 @@ export const deepLinkStateTransitionTable: Record<string, Record<string, DeepLin
 		[DeepLinkServiceAction.OpenGraph]: DeepLinkServiceState.OpenGraph,
 		[DeepLinkServiceAction.OpenFile]: DeepLinkServiceState.OpenFile,
 		[DeepLinkServiceAction.OpenSwitch]: DeepLinkServiceState.SwitchToRef,
+		[DeepLinkServiceAction.OpenAllPrChanges]: DeepLinkServiceState.OpenAllPrChanges,
```
axosoft-ramint, line 394: This transition is no longer necessary since it would/should only be reachable from `OpenInspect`

src/uris/deepLinks/deepLinkService.ts
```diff
@@ -1386,6 +1385,34 @@ export class DeepLinkService implements Disposable {
 						repository: repo,
 						source: 'launchpad',
 					} satisfies ShowWipArgs);
+					action = DeepLinkServiceAction.OpenAllPrChanges;
```
axosoft-ramint, line 1388: Right now this would work because we really only open inspect on PR switches, in which case all the info is there to show the PR changes. Would be a nice additional touch though to check the action type in `this._context.action` though and maybe restrict/guard against any non-PR case.

axosoft-ramint (commented):
This works for me - though I was speaking with @eamodio this morning and thinking that it might not make sense to always load the changes. Will follow up on that in a bit.

axosoft-ramint:
Hey @nzaytsev, while reviewing your PR, I'd suggest the following code changes:

👉 [**Only open changes when worktree is created**](https://gitkraken.dev/drafts/8aec87fe-8402-491f-9634-faa549bf5163?source=PRComment)
>- Removes unneeded transition event
- Only opens changes when using "Open in Worktree"
- Only opens changes when a worktree is created in the flow

You can also review and apply these suggestions [locally on your machine](https://gitkraken.dev/link/drafts/8aec87fe-8402-491f-9634-faa549bf5163?type=suggested_pr_change&prEntityId=WyJnaXRodWIiLCJwciIsIjEiLCIiLCJQUl9rd0RPQS1NUnVNNkFDcmRVIl0=&source=PRComment).

<details><summary>Learn more about GitKraken Code Suggest</summary>

>[Code Suggest](https://gitkraken.dev/redirect/aHR0cHM6Ly93d3cuZ2l0a3Jha2VuLmNvbS9zb2x1dGlvbnMvY29kZS1zdWdnZXN0P3NvdXJjZT1naXRodWIuY29tJm1lZGl1bT1wcl9jb21tZW50?source=PRSuggest&event=redirectToCodeSuggestPage&provider=github) liberates your code reviews from GitHub's restrictive, comment-only feedback style. As simple as suggesting changes in a Google-doc, provide real code suggestions from where you code, e.g. your IDE, and on anything in your project — not just on the lines of code changed in the PR.

_Join your team on [GitKraken](https://gitkraken.dev/redirect/aHR0cHM6Ly93d3cuZ2l0a3Jha2VuLmNvbT9zb3VyY2U9Z2l0aHViLmNvbSZtZWRpdW09cHJfY29tbWVudA==?source=PRSuggest&event=redirectToIndexPage&provider=github) to speed up PR review._
</details>

axosoft-ramint:
Hey @nzaytsev, while reviewing your PR, I'd suggest the following code changes:

👉 [**Restricts to only opening on "Open in Worktree" and only when worktree is created**](https://gitkraken.dev/drafts/fef93b2c-3abe-4d0a-ba22-04bec8df61db?source=PRComment)
>

You can also review and apply these suggestions [locally on your machine](https://gitkraken.dev/link/drafts/fef93b2c-3abe-4d0a-ba22-04bec8df61db?type=suggested_pr_change&prEntityId=WyJnaXRodWIiLCJwciIsIjEiLCIiLCJQUl9rd0RPQS1NUnVNNkFDcmRVIl0=&source=PRComment).

<details><summary>Learn more about GitKraken Code Suggest</summary>

>[Code Suggest](https://gitkraken.dev/redirect/aHR0cHM6Ly93d3cuZ2l0a3Jha2VuLmNvbS9zb2x1dGlvbnMvY29kZS1zdWdnZXN0P3NvdXJjZT1naXRodWIuY29tJm1lZGl1bT1wcl9jb21tZW50?source=PRSuggest&event=redirectToCodeSuggestPage&provider=github) liberates your code reviews from GitHub's restrictive, comment-only feedback style. As simple as suggesting changes in a Google-doc, provide real code suggestions from where you code, e.g. your IDE, and on anything in your project — not just on the lines of code changed in the PR.

_Join your team on [GitKraken](https://gitkraken.dev/redirect/aHR0cHM6Ly93d3cuZ2l0a3Jha2VuLmNvbT9zb3VyY2U9Z2l0aHViLmNvbSZtZWRpdW09cHJfY29tbWVudA==?source=PRSuggest&event=redirectToIndexPage&provider=github) to speed up PR review._
</details>

axosoft-ramint:
@eamodio Asking for your help reviewing this one since I committed to it directly.

axosoft-ramint:
Going to go ahead and merge this - we can address further issues in follow-up

PR #3722, opened 2024-11-08
Adds ability to group views on the SCM side bar

eamodio:
Adds configuration option to enable/disable grouped views
Adds ability to move views in and out of the grouped "view"
Adds ability to set one of the grouped views as the default

image


d13 (approved):
LGTM

PR #3726, opened 2024-11-08
Updates launchpad and commit-graph images on README.md

jkelroy:


images/docs/commit-graph.png
d13, line 1: This appears to be missing the new filter button next to the search box.

Screenshot 2024-11-08 at 5 04 27 PM

eamodio: These images are also quite large (both in file size and resolution. I don't think we need a 1600x900 image for the readme

jkelroy: I will update the images with the image sizes halved (800w).

@eatmorespinach do we have a updated commit graph image with the new filter button next to the search box?

d13:
This is out of date and have been updated

PR #3775, opened 2024-11-18
Shows different "globe" tooltips on GitHub and Jira issues in StartWork

### LINKED ISSUE #3774: Start Work: show different "globe" tooltips on GitHub and Jira issues
This issue is a follow-up of #3621

Image

sergeibbb:
Closes #3774

# Checklist

### DIFF
--- a/src/constants.telemetry.ts
+++ b/src/constants.telemetry.ts
@@ -328,6 +328,7 @@ export type TelemetryEvents = {
 	'startWork/issue/action': StartWorkEventData & {
 		action: 'soft-open';
 		connected: boolean;
+		type: StartWorkType;
 	} & Partial<Record<`item.${string}`, string | number | boolean>>;
 	/** Sent when the user chooses an issue to start work in the second step */
 	'startWork/issue/chosen': StartWorkEventData & {
--- a/src/constants.telemetry.ts
+++ b/src/constants.telemetry.ts
@@ -324,6 +324,11 @@ export type TelemetryEvents = {
 		connected: boolean;
 		type: StartWorkType;
 	};
+	/** Sent when the user takes an action on a StartWork issue */
+	'startWork/issue/action': StartWorkEventData & {
+		action: 'soft-open';
+		connected: boolean;
+	} & Partial<Record<`item.${string}`, string | number | boolean>>;
 	/** Sent when the user chooses an issue to start work in the second step */
 	'startWork/issue/chosen': StartWorkEventData & {
 		connected: boolean;
--- a/src/plus/startWork/startWork.ts
+++ b/src/plus/startWork/startWork.ts
@@ -56,6 +56,9 @@ interface State {
 	type?: StartWorkType;
 	inWorktree?: boolean;
 }
+interface StateWithType extends State {
+	type: StartWorkType;
+}
 
 export type StartWorkType = 'branch' | 'branch-worktree' | 'issue' | 'issue-worktree';
 type StartWorkTypeItem = { type: StartWorkType; inWorktree?: boolean };
@@ -166,6 +169,7 @@ export class StartWorkCommand extends QuickCommand<State> {
 					}
 				}
 
+				assertsTypeStepState(state);
 				const result = yield* this.pickIssueStep(state, context, opened);
 				opened = true;
 				if (result === StepResultBreak) continue;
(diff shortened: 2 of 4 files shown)


src/constants.telemetry.ts
```diff
@@ -324,6 +324,13 @@ export type TelemetryEvents = {
 		connected: boolean;
 		type: StartWorkType;
 	};
+	/** Sent when the user takes an action on a StartWork issue */
+	'startWork/issue/action': StartWorkEventData & {
+		action: 'soft-open';
```
axosoft-ramint, line 329: This should be typed similar to `startWork/issue/chosen` below (add `connected` and `type` properties, and either leave `item.` flexible as is done there, or update that event with tighter typing on `item.` properties), with the addition of `action`.

src/plus/startWork/startWork.ts
```diff
@@ -569,3 +588,15 @@ function isStartWorkTypeItem(item: unknown): item is StartWorkTypeItem {
 export function getStartWorkItemIdHash(item: StartWorkItem) {
 	return md5(item.item.issue.id);
 }
+
+function getOpenOnWebQuickInputButton(integrationId: string): QuickInputButton | undefined {
+	switch (integrationId) {
+		case HostingIntegrationId.GitHub:
+		case SelfHostedIntegrationId.GitHubEnterprise:
+			return OpenOnGitHubQuickInputButton;
```
axosoft-ramint, line 596: Do we support GitHub enterprise? Will this interface show you issues from your GitHub Enterprise instance if you have it connected? If not, we shouldn't include it here yet.

src/plus/startWork/startWork.ts
```diff
@@ -518,6 +527,16 @@ export class StartWorkCommand extends QuickCommand<State> {
 		void openUrl(item.item.issue.url);
 	}
 
+	private sendItemActionTelemetry(action: 'soft-open', item: StartWorkItem, context: Context) {
+		this.container.telemetry.sendEvent('startWork/issue/action', {
+			...context.telemetryContext!,
+			action: action,
+			'item.id': getStartWorkItemIdHash(item),
+			'item.type': item.item.issue.type,
+			'item.provider': item.item.issue.provider.id,
```
axosoft-ramint, line 536: We should send at least the same properties of the chosen item as we do in `startWork/issue/chosen`. If there are others we can send (that don't include private information), feel free to expand both. Perhaps we should create a function for forming the single-item telemetry payload and use it in both places.

src/plus/startWork/startWork.ts
sergeibbb: fixed

src/plus/startWork/startWork.ts
sergeibbb: fixed

src/plus/startWork/startWork.ts
```diff
@@ -166,6 +172,10 @@ export class StartWorkCommand extends QuickCommand<State> {
 					}
 				}
 
+				if (!isStateWithType(state)) {
+					continue;
+				}
```
axosoft-ramint, line 177: We should probably throw here and possibly include a `debugger` line if we don't have a `type` by this point, because something must have gone very wrong. Maybe we can use a function similar to `assertsLaunchpadStepState`?

src/plus/startWork/startWork.ts
```diff
@@ -570,3 +594,30 @@ function isStartWorkTypeItem(item: unknown): item is StartWorkTypeItem {
 export function getStartWorkItemIdHash(item: StartWorkItem) {
 	return md5(item.item.issue.id);
 }
+
+function buildItemActionTelemetryData(item: StartWorkItem) {
```
axosoft-ramint, line 598: Nitpick: this is used outside of "action" contexts, and the `action` property is really only being set in `sendItemActionTelemetry`, so maybe we just call this `buildItemTelemetryData`?

axosoft-ramint (commented):
Two tiny, very quick suggestions. Works in testing, so I think once these are addressed we're good to merge it.

src/plus/startWork/startWork.ts
sergeibbb: fixed

src/plus/startWork/startWork.ts
sergeibbb: @axosoft-ramint fixed

PR #3780, opened 2024-11-19
Uses correct counts in rebase command condition/messaging

### LINKED ISSUE #3747: "Rebase Current Branch onto Branch" incorrectly shows that the current branch is caught up to the destination
Can be reproduced in the commit graph by having your current branch be behind some commits from a destination branch, then choosing the "Rebase Current Branch onto Branch" context menu option. The quickpick will show that there is nothing to rebase when in fact there is.

axosoft-ramint:
Fixes #3747

Updates property names to be clearer in rebase command, and uses the correct side of the left-right commit count to:

1. Determine if the current branch is caught up to the destination
2. Determine how many commits will be rebased on top of the destination

Repro steps:

1. Choose a rebase command: "rebase current branch onto branch", "rebase current branch onto tip", or "GitLens: Git Rebase" command palette command.

2. Before: Quickpick prevents rebase and states that the current branch is already up-to-date with the destination when it is not. After: Quickpick should allow the rebase, and give the correct number of commits that will be rebased on top of the destination.

### DIFF
--- a/src/commands/git/rebase.ts
+++ b/src/commands/git/rebase.ts
@@ -28,7 +28,7 @@ interface Context {
 	repos: Repository[];
 	associatedView: ViewsWithRepositoryFolders;
 	cache: Map<string, Promise<GitLog | undefined>>;
-	destination: GitBranch;
+	branch: GitBranch;
 	pickCommit: boolean;
 	pickCommitForItem: boolean;
 	selectedBranchOrTag: GitReference | undefined;
@@ -40,7 +40,7 @@ type Flags = '--interactive';
 
 interface State {
 	repo: string | Repository;
-	reference: GitReference;
+	destination: GitReference;
 	flags: Flags[];
 }
 
@@ -63,7 +63,7 @@ export class RebaseGitCommand extends QuickCommand<State> {
 			counter++;
 		}
 
-		if (args?.state?.reference != null) {
+		if (args?.state?.destination != null) {
 			counter++;
 		}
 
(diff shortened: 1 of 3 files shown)


src/commands/git/rebase.ts
```diff
@@ -219,24 +219,25 @@ export class RebaseGitCommand extends QuickCommand<State> {
...
+			createRevisionRange(state.destination.ref, context.branch.ref, '...'),
 			{ excludeMerges: true },
 		);
 
-		const title = `${context.title} ${getReferenceLabel(state.reference, { icon: false, label: false })}`;
-		const count = counts != null ? counts.left : 0;
-		if (count === 0) {
+		const title = `${context.title} ${getReferenceLabel(state.destination, { icon: false, label: false })}`;
+		const ahead = counts != null ? counts.right : 0;
+		const behind = counts != null ? counts.left : 0;
```
axosoft-ramint, line 228: The core of the bug that is being addressed lies in these lines. The ref that is passed into the rebase command via state is the **destination** of the rebase, not the **source**. The rebase command is rebasing the **current branch** (previously housed in `context.destination`, renamed now to `context.branch` onto the **chosen branch** (previously housed in `state.reference`, renamed here to `state.destination`).

When we get the counts in `container.git.getLeftRightCommitCount`, we consistently put the "destination" on the left of the git revision range. In this case, because of the mislabeling, we were putting the source (current branch) on the left side and calling it the "destination", and the actual destination (chosen branch) is on the right side. So `counts.left` was giving us "how many commits ahead the current branch is" when we really wanted to know "how many commits behind the current branch is". In the messaging further down though, we say "will rebase by applying X commits on top of". Here, we want to know "how many commits ahead the current branch is".

So I split the counts into clearly labeled `ahead` and `behind` variables, put the "destination" and "source" into the correct slots of the revision range, and renamed everywhere so that it's clear that the destination is the argument provided to the command in the state/initial state, and the source is the current branch that we load into the context.

Issue #3788, opened 2024-11-21
Launchpad: API search for PRs on GitLab


sergeibbb:
This issue is a follow-up of #3543 

---

#### Todo
 
- ✅ Search for PRs on GitLab
- ✅ Perform the PR search for all connected providers
- ✅ Group found repo+prID pairs by provider to perform the more precise search
- ✅ Search GitLab PR by PR URL
- ✅ Get head-ref id on each PR, to be able to show checkout-like actions

#### Follow-ups

- ⏺️ Search by number on current repositories

#### Testing notes

Connect both GitLab and GitHub. Search by:
- GitHub URL of a local PR (a PR from your repositories)
- GitLab URL of a local PR
- GitHub URL in internet (a random PR on internet)
- GitLab URL in internet
- text query

Pay attention to user avatars. GitLab's graph-ql returns them differently in different requests: sometimes it's an URL, but sometimes it's just a path.

Make sure that owner/repo is shown.

Disconnect GitHub. Search by a text query.

sergeibbb:
Hi @axosoft-ramint , @eamodio 

I have a following question. It looks like GitLab's GraphQL is not able to search for PRs in general. It can does the following:
- get for a user: one of the following
  - assignedMergeRequests
  - authoredMergeRequests
  - reviewRequestedMergeRequests
- get for a project/group

Now for the Launchpad "get for a user" set of requests is used but retrieving only  open ones: `state:opened`

Also there is a REST API endpoint that just searches everywhere, e.g.: https://gitlab.com/api/v4/search?scope=merge_requests&search=flight

What I can do:
1. use "get by user" requests withtout `state:opened` limitations. By doing that I loose PRs that are not associated with the current user.
2. use "get by project" request for currently opened repositories. By doing that I loose PRs that belong to other projects.
3. use REST API endpoint. The results can be too wide, but now we have same for GitHub.

For now I'm going to continue with #3 until we choose another option or find a better solution.

What do you think?

axosoft-ramint:
I feel like since what we're doing with GitHub on a search is "search everywhere", we should have equivalent behavior for GitLab. So 3 sounds right.

PR #3795, opened 2024-11-25
Search for GitLab PRs in Launchpad

sergeibbb:
#3788
# Description

# Checklist

### DIFF
--- a/src/git/models/pullRequest.ts
+++ b/src/git/models/pullRequest.ts
@@ -2,12 +2,10 @@ import { Uri, window } from 'vscode';
 import { Schemes } from '../../constants';
 import { Container } from '../../container';
 import type { RepositoryIdentityDescriptor } from '../../gk/models/repositoryIdentities';
-import type { EnrichablePullRequest } from '../../plus/integrations/providers/models';
 import { formatDate, fromNow } from '../../system/date';
 import { memoize } from '../../system/decorators/memoize';
 import type { LeftRightCommitCountResult } from '../gitProvider';
 import type { IssueOrPullRequest, IssueRepository, IssueOrPullRequestState as PullRequestState } from './issue';
-import type { PullRequestUrlIdentity } from './pullRequest.utils';
 import type { ProviderReference } from './remoteProvider';
 import type { Repository } from './repository';
 import { createRevisionRange, shortenRevision } from './revision.utils';
@@ -418,22 +416,4 @@ export async function getOpenedPullRequestRepo(
 
 	const repo = await getOrOpenPullRequestRepository(container, pr, { promptIfNeeded: true });
 	return repo;
-}
-
-export function doesPullRequestSatisfyRepositoryURLIdentity(
-	pr: EnrichablePullRequest | undefined,
-	{ ownerAndRepo, prNumber }: PullRequestUrlIdentity,
-): boolean {
-	if (pr == null) {
-		return false;
-	}
-	const satisfiesPrNumber = prNumber != null && pr.number === parseInt(prNumber, 10);
-	if (!satisfiesPrNumber) {
-		return false;
-	}
-	const satisfiesOwnerAndRepo = ownerAndRepo != null && pr.repoIdentity.name === ownerAndRepo;
-	if (!satisfiesOwnerAndRepo) {
-		return false;
-	}
-	return true;
 }
--- a/src/plus/integrations/integration.ts
+++ b/src/plus/integrations/integration.ts
@@ -16,6 +16,7 @@ import type {
 	PullRequestState,
 	SearchedPullRequest,
 } from '../../git/models/pullRequest';
+import type { PullRequestUrlIdentity } from '../../git/models/pullRequest.utils';
 import type { RepositoryMetadata } from '../../git/models/repositoryMetadata';
 import { showIntegrationDisconnectedTooManyFailedRequestsWarningMessage } from '../../messages';
 import { gate } from '../../system/decorators/gate';
@@ -1361,4 +1362,6 @@ export abstract class HostingIntegration<
 		repos?: T[],
 		cancellation?: CancellationToken,
 	): Promise<PullRequest[] | undefined>;
+
+	getPullRequestIdentityFromMaybeUrl?(search: string): PullRequestUrlIdentity | undefined;
 }
--- a/src/plus/integrations/integration.ts
+++ b/src/plus/integrations/integration.ts
@@ -556,7 +556,7 @@ export abstract class IntegrationBase<
 		const connected = this.maybeConnected ?? (await this.isConnected());
 		if (!connected) return undefined;
 
-		const pr = this.container.cache.getPullRequest(id, resource, this, () => ({
+		const pr = await this.container.cache.getPullRequest(id, resource, this, () => ({
 			value: (async () => {
 				try {
 					const result = await this.getProviderPullRequest?.(this._session!, resource, id);
--- a/src/plus/integrations/integration.ts
+++ b/src/plus/integrations/integration.ts
@@ -1363,5 +1363,9 @@ export abstract class HostingIntegration<
 		cancellation?: CancellationToken,
 	): Promise<PullRequest[] | undefined>;
 
-	getPullRequestIdentityFromMaybeUrl?(search: string): PullRequestUrlIdentity | undefined;
+	getPullRequestIdentityFromMaybeUrl(search: string): PullRequestUrlIdentity | undefined {
+		return this.getProviderPullRequestIdentityFromMaybeUrl?.(search);
+	}
+
+	protected getProviderPullRequestIdentityFromMaybeUrl?(search: string): PullRequestUrlIdentity | undefined;
 }
--- a/src/plus/launchpad/launchpadProvider.ts
+++ b/src/plus/launchpad/launchpadProvider.ts
@@ -214,13 +214,15 @@ export class LaunchpadProvider implements Disposable {
 	}
 
 	private async getSearchedPullRequests(search: string, cancellation?: CancellationToken) {
-		const prUrlIdentity: PullRequestUrlIdentity | undefined = await this.getPullRequestIdentityFromSearch(search);
+		const connectedIntegrations = await this.getConnectedIntegrations();
+		const prUrlIdentity: PullRequestUrlIdentity | undefined = await this.getPullRequestIdentityFromSearch(
+			search,
+			connectedIntegrations,
+		);
 		const result: { readonly value: SearchedPullRequest[]; duration: number } = {
 			value: [],
 			duration: 0,
 		};
-
-		const connectedIntegrations = await this.getConnectedIntegrations();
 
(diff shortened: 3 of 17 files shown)


sergeibbb:
@axosoft-ramint 

In the last commit I'm grouping detected PR-identity by providers, because thought that if we know that an URL matches to a GitLab then no need to search for it in GitHub. But if it looks too compicated, then then last commit can be rejected.

axosoft-ramint (changes requested):
As it stands, this doesn't quite feel ready. The utils and test files that are provider-specific should be organized into provider-specific directories. The url identity being a mapped object by provider id feels confusing - the identity itself should just come in on the function. Some of the changes to the PR utils need either explanation or rewrite to make more sense (why are we defining an undefined const and then returning it without touching it?). The same function being defined in the PR utils but then also implemented specifically for integrations doesn't quite make sense to me. The `checkoutable` logic in launchpad shouldn't be necessary - we should always have at least a defined head ref and remote url for a PR, or else we're not fetching enough info.

Can talk on this further on your return.

sergeibbb:
@axosoft-ramint 

> The utils and test files that are provider-specific should be organized into provider-specific directories.

fixed

> The url identity being a mapped object by provider id feels confusing - the identity itself should just come in on the function.

fixed

> why are we defining an undefined const and then returning it without touching it?

where?

> The same function being defined in the PR utils but then also implemented specifically for integrations doesn't quite make sense to me

I access the necessary implementation of the function through the integration instance that's why it's in method.
But I cannot keep the implementation in the integration because with our current test configuration it cannot be tested (our tests cannot handle importing the Context). To make the function testable as unit I move it to an external file.

> The checkoutable logic in launchpad shouldn't be necessary

I'll try to fix it. I think that in order to do that I have to rewrite retrieving GitLab PRs from REST to GraphQL.

sergeibbb:
@axosoft-ramint 

> The utils and test files that are provider-specific should be organized into provider-specific directories.

fixed

> The url identity being a mapped object by provider id feels confusing - the identity itself should just come in on the function.

fixed

> why are we defining an undefined const and then returning it without touching it?

where?

> The same function being defined in the PR utils but then also implemented specifically for integrations doesn't quite make sense to me

I access the necessary implementation of the function through the integration instance that's why it's in method.
But I cannot keep the implementation in the integration because with our current test configuration it cannot be tested (our tests cannot handle importing the Context). To make the function testable as unit I move it to an external file.

> The checkoutable logic in launchpad shouldn't be necessary

I'll try to fix it. I think that in order to do that I have to rewrite retrieving GitLab PRs from REST to GraphQL.

But initially I thought that it can be a follow-up:

image

sergeibbb:
@axosoft-ramint 

> The utils and test files that are provider-specific should be organized into provider-specific directories.

fixed

> The url identity being a mapped object by provider id feels confusing - the identity itself should just come in on the function.

fixed

> why are we defining an undefined const and then returning it without touching it?

fixed

> The same function being defined in the PR utils but then also implemented specifically for integrations doesn't quite make sense to me

I access the necessary implementation of the function through the integration instance that's why it's in method.
But I cannot keep the implementation in the integration because with our current test configuration it cannot be tested (our tests cannot handle importing the Context). To make the function testable as unit I move it to an external file.

But I'm not sure about one thing: https://github.com/gitkraken/vscode-gitlens/blob/89bd972aa29c196a2b179e7a08ffad45291913ab/src/plus/launchpad/launchpadProvider.ts#L584-L611


> The checkoutable logic in launchpad shouldn't be necessary

Fixed.

src/git/models/pullRequest.ts
```diff
@@ -272,7 +270,7 @@ export function getRepositoryIdentityForPullRequest(
 	if (headRepo) {
 		return {
 			remote: {
-				url: pr.refs?.head?.url,
+				url: pr.repository.url,
```
axosoft-ramint, line 273: This changes the url ref to the base repository. This would break the model for PRs where the PR is targeting a different repository and (like when the head ref is from someone's fork).

Commands that use this like "open the PR's repository" or "open in worktree" will break in those cases.

src/plus/launchpad/launchpadProvider.ts
```diff
@@ -210,44 +214,74 @@ export class LaunchpadProvider implements Disposable {
...
-			if (pr?.value != null) {
-				result = { value: [{ pullRequest: pr.value, reasons: [] }], duration: pr.duration };
-				return { prs: result, suggestionCounts: undefined };
+		const prUrlIdentity: PullRequestUrlIdentity | undefined = await this.getPullRequestIdentityFromMaybeUrl(search);
+		const result: { readonly value: SearchedPullRequest[]; duration: number } = {
+			value: [],
+			duration: 0,
+		};
+
+		const connectedIntegrations = await this.getConnectedIntegrations();
```
axosoft-ramint, line 223: If we're already getting the connected integrations below, maybe we can just pass them into `this.getPullRequestIdentityFromMaybeUrl`

src/plus/launchpad/launchpadProvider.ts
```diff
@@ -210,44 +214,74 @@ export class LaunchpadProvider implements Disposable {
...
-			const integration = await this.container.integrations.get(HostingIntegrationId.GitHub);
+			return undefined;
+		};
+
+		const findByQuery = async (
+			integration: HostingIntegration,
+		): Promise<undefined | TimedResult<SearchedPullRequest[] | undefined>> => {
 			const prs = await withDurationAndSlowEventOnTimeout(
-				integration?.searchPullRequests(search, undefined, cancellation),
+				integration?.searchPullRequests(search, undefined, cancellation), //
```
axosoft-ramint, line 253: Empty comment line at the end

axosoft-ramint (changes requested):
Seems to work so far on my end, but I noticed an issue with the url reference in the RP repo identity that will likely break functionality in some cases with other features.

src/git/models/pullRequest.ts
sergeibbb: @axosoft-ramint 
Thank you for catching it.

Question here. `pr.refs?.head?.url` is a link to a branch to to the repo. It fails matching to `uniqueRemoteUrls` in the LaunchpadProvider when I try to `swtichTo` the branch.
So I see 3 ways to fix it:

1. Add `pr.refs.head.repoUrl` on the specific provider level.
2. Fix the matching to `uniqueRemoteUrls` in launchpadProvider
3. Cut it right here: `url: someOperationOnRefUrl(pr.refs?.head?.url)`

I think that the first option is the best because it can respect specifics of each type of integration. What do you think?

axosoft-ramint: @sergeibbb `pr.hefs?.head?.url` should be the link to the remote (repository) that the head ref is hosted on, not a link to a branch - at least, with GitHub this is the case. Are you finding that with GitLab, it is linking to a branch? If so, whatever we are using to convert GitLab PRs into our PR model, I think we are plugging the wrong datapoint into that property and should start there.

src/plus/launchpad/launchpadProvider.ts
sergeibbb: fixed

src/plus/launchpad/launchpadProvider.ts
sergeibbb: fixed

src/git/models/pullRequest.ts
sergeibbb: @axosoft-ramint 

GitLab REST API does not provide necessary information about source and target repositories. So, I've fixed it by switching to GraphQL.  (26727ffce4a3546a1b618d38c7c15e4d615d2a4f) 

However I haven't found a good way to search MRs via GraphQL. So I've solved it with 2 requests. I still use REST to search and then I use PRs' IDs to retrieve the rest of the fields from GraphQL. (3cb31d915a128b051a77e026e7d43791f5fbe0a4) 

cc @nzaytsev

src/plus/integrations/integration.ts
```diff
@@ -1361,4 +1362,6 @@ export abstract class HostingIntegration<
 		repos?: T[],
 		cancellation?: CancellationToken,
 	): Promise<PullRequest[] | undefined>;
+
+	getPullRequestIdentityFromMaybeUrl?(search: string): PullRequestUrlIdentity | undefined;
```
axosoft-ramint, line 1366: Let's follow a similar pattern for generic and provider-specific function implementation and naming here as we do for other functions, in case we want to expand on the base logic later:

1. Have a `protected getProviderPullRequestIdentityFromMaybeUrl` that can be implemented by specific providers (i.e. GitHub, GitLab, etc.)
2. Have a `getPullRequestIdentityFromMaybeUrl` that, for now, just does `return getProviderPullRequestIdentityFromMaybeUrl` and later if we want, we can layer in caching there.

src/plus/launchpad/launchpadProvider.ts
```diff
@@ -547,6 +583,30 @@ export class LaunchpadProvider implements Disposable {
...
 
+	isMaybeSupportedLaunchpadPullRequestSearchUrl(search: string): boolean {
+		return (
+			isMaybeGitHubPullRequestUrl(search) ||
+			isMaybeGitLabPullRequestUrl(search) ||
+			isMaybeNonSpecificPullRequestSearchUrl(search)
+		);
+	}
+
+	async getPullRequestIdentityFromMaybeUrl(
```
axosoft-ramint, line 594: I suggest calling this one `getPullRequestIdentityFromSearch` to avoid making it seem as if line 607 is a recursive call to the same function.

axosoft-ramint (commented):
On testing, the only issue I'm seeing is that owner/repo are not showing up on GitLab search results:

image

They do for GitLab PRs that are not search results.

Apart from that, some minor nitpicks that should be easy to change:

src/plus/integrations/integration.ts
```diff
@@ -556,7 +557,7 @@ export abstract class IntegrationBase<
 		const connected = this.maybeConnected ?? (await this.isConnected());
 		if (!connected) return undefined;
 
-		const pr = this.container.cache.getPullRequest(id, resource, this, () => ({
+		const pr = await this.container.cache.getPullRequest(id, resource, this, () => ({
```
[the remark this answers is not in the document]
sergeibbb: @axosoft-ramint 

I did it because I set a breakpoint at this place it was convenient to see a resolved promise.

src/plus/integrations/integration.ts
sergeibbb: fixed

sergeibbb:
@axosoft-ramint 

> I'm seeing is that owner/repo are not showing up on GitLab search results:

fixed

src/plus/launchpad/launchpadProvider.ts
sergeibbb: fixed

axosoft-ramint (approved):
I wont have the chance to test this morning, but if all my prior comments are addressed, feel free to merge.

PR #3797, opened 2024-11-26
Adds stashes into Commit view

eamodio:
I think this is ready for review/testing, as it works for any views that use "branch nodes", e.g. Commits, Branches, Worktrees, Repositories views. 

~~We will need to figure out how to get it to work with the Commit Graph without having to refresh all the commits (and stashes) based on the selected "Branches Visibility"~~

~~@axosoft-ramint maybe instead of filtering the stash commits, we just also return a set of "reachable" commits, and then be able to pass both sets into the Graph or something.~~

We will need to handle the Commit Graph differently.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1743,6 +1743,13 @@
 				"title": "Commits View",
 				"order": 110,
 				"properties": {
+					"gitlens.views.commits.showStashes": {
+						"type": "boolean",
+						"default": false,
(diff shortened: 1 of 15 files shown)


eamodio:
The settings are all off by default, because there could be perf implications for large repos. So the risks here are super low.
I think this just need a quick glance.

PR #3802, opened 2024-11-26
Adds ability to show the default remote branches in the Branches view

eamodio:
The setting is off by default.

Has some helpers that can be useful when adding the local/remote toggle to "recents" on Home


axosoft-ramint (approved):
LGTM, thanks!

PR #3804, opened 2024-11-26
(WIP) Refactoring out view nodes to better deal with id collisions and storage

eamodio:
Requires a lot more work, but once it lands we should be able to save & restore selected nodes when switching between "tabs" on the GitLens view


d13:
This was resolved with 91ace9acf, ed48f6300, a0d9b4dc4, 5e668aa6d, e192f31bf

PR #3822, opened 2024-12-03
Utilizes paging to get more Jira issues

axosoft-ramint:
Fixes current bug in which only 10 Jira issues maximum show up in Start Work (unable to create an issue due to a GitHub bug, will link the issue when I can create it).

There is an update to the shared provider library which increases the results per page from 10 to 100 - will look at bumping our library version as a separate task since there are some refactors needed to make it happy (and other risks that need to be tested).


eamodio (approved):
:shipit:

PR #3845, opened 2024-12-09
Restores optimistic behavior for Launchpad pin/snooze

### LINKED ISSUE #3791: Pin/snooze in Launchpad trigger a full refresh and are no longer optimistic
Clicking "pin/snooze" on Launchpad items would previously be optimistic, immediately updating the affected item before calling the backend. Now it waits for a full refresh before anything changes.

axosoft-ramint:
Closes #3791

At some point along the way, we seem to have changed the "update items" logic in Launchpad to always force a full refresh of everything. For actions like pin, snooze, open on provider, merge, etc. we already have the provider bust the relevant cache so that only items that have changed are refreshed. So I changed the default behavior to use caches and only in a few places now do we force the full refresh.

Furthermore, to make pin/snooze/unpin/unsnooze even more optimistic, I created an `optimisticallyUpdateItems` function that just refreshes the list without having to do all the rest of the work.

These actions should be much more responsive now.

### DIFF
--- a/src/plus/launchpad/launchpad.ts
+++ b/src/plus/launchpad/launchpad.ts
@@ -461,11 +461,15 @@ export class LaunchpadCommand extends QuickCommand<State> {
 			};
 		};
 
-		const getItems = (result: LaunchpadCategorizedResult, isSearching?: boolean) => {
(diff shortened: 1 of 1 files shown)


eamodio (approved):
:shipit:

PR #3861, opened 2024-12-12
Redesigns Home account footer into new header bar

eamodio:
Adds a popover for account information and integration connection and status


d13:
This was approved by me.

PR #3865, opened 2024-12-13
Refines issue model and adds getIssue to supported integrations

axosoft-ramint:
Refines the issue model to be more supportive of Jira issues now that we're working with them more:
 - Repository info on an issue would now be optional
 - Adds optional project property with information on project and resource

Improves typing on resource descriptors in preparation for encoding them with a branch's associated issues in the git config

Adds getIssue and getProviderIssue to the base integration model, and implements for GitHub, GitLab and Jira. Updates provider API functionality to accommodate both GitLab and Jira.


eamodio (approved):
LGTM

PR #3878, opened 2024-12-17
Reduces VSIX size ignoring docs media

nzaytsev:
# Description

# Checklist


.vscodeignore
```diff
@@ -7,8 +7,7 @@
 .yarn/**
 docs/**
 emoji/**
-images/docs/sponsors/**
-images/docs/gitlens-preview.gif
+images/docs
```
eamodio, line 10: We can't exclude all the docs images, because any images referenced in the README (which is the extension manifest page), walkthroughs, etc, need to be available locally.

.vscodeignore
```diff
@@ -46,3 +45,4 @@ tsconfig*.json
 tsconfig*.tsbuildinfo
 webpack.config*.mjs
 yarn.lock
+.wdio-vscode-service
```
eamodio, line 48: This doesn't hurt anything, but this already isn't getting bundled today.

eamodio (changes requested):
Can't merge as is -- need to go through the images and see if they are used or not and if not should be removed.

.vscodeignore
nzaytsev: > but this already isn't getting bundled today

I see that this folder is not git ignored, and I see it in the bundled zip
Screenshot 2024-12-18 at 10 22 00

> This doesn't hurt anything

Maybe it's truth in the environment that is isolated from tests running, but I see that it may affect the final size

Screenshot 2024-12-18 at 10 23 12

.vscodeignore
nzaytsev: >  any images referenced in the README

They are referenced to raw.githubusercontent.com, for example: https://github.com/gitkraken/vscode-gitlens/blob/0f6f6ac6fa05b6d6be1d225d6cc973dcfcd2a60e/README.md?plain=1#L14

If I open readme, I see that it's referenced to cloud storage, not to locally bundled images

Screenshot 2024-12-18 at 10 27 12

There is no place where the image from this dir is used locally

> walkthroughs, etc

walkthroughs assets are placed in another directory, I didn't touch it. But it can also be moved to cloud storage and removed from the bundle

.vscodeignore
eamodio: Not sure what is creating that folder -- we aren't using webdriver, is that just an old folder you have? When I said it won't be bundled, I meant via CI (your local folder can have all kinds of stuff that would get "accidentally" bundled)

.vscodeignore
eamodio: Ah -- I forgot that the VS Code marketplace changed to require all README links to be `https`. So it looks like you are correct those images don't need to be bundled.

As for the walkthroughs, if we use `https` links then the walkthroughs won't work offline, which is not ideal. So we should leave them alone.

eamodio (approved):

Issue #3901, opened 2024-12-20
Integrations: support self-hosted providers from gkdev


d13:
The goal is not to just convert over to GKdev like our other providers. We need to continue support of multiple instances of self-hosted and GKdev only supports one instance.

High level changes:

- ✅ Send users to GKdev to connect to self-hosted GitHub or GitLab
  - ✅ if the integration exists on GKdev show the enable/disable toggle on the remote
  - ❌ if the integration does not exist: show the enable toggle and send user to GKdev
  - ✅ Show the "Connect GitHub Enterprise" on the Home View
- ✅ Existing self-hosted integations (GitHub or GitLab) on GKdev, bring those into our system, but still support our custom self-hosted auth
- ✅ Once we have the info, we need to register a custom provider
- ➡️ A follow-up: #3923 

👍 The code where this happens in [remoteProviders.ts](https://github.com/gitkraken/vscode-gitlens/blob/47608852888944469e6bcede504228fc9b611de1/src/git/remotes/remoteProviders.ts#L76-L98)
👍 Essentially, we can just take the list of custom remotes and add another custom entry from the GKdev data, at runtime. We don't want to update their `settings.json` (though it might be an easy path for initial testing).

Potential gotchas:

- During development, if GKdev doesn't provide the correct data, let's just temporarily stub the data in while we wait for GKDS/GKdev to provide the info (aka URL)

### Testing

- ✅ issue linking works
- ✅ pull requests work

PR #3922, opened 2025-01-13
Integrations support self hosted providers from gkdev (#3901)

sergeibbb:
# Description

#3901

# Checklist

### DIFF
--- a/src/constants.integrations.ts
+++ b/src/constants.integrations.ts
@@ -7,6 +7,7 @@ export enum HostingIntegrationId {
 
 export enum SelfHostedIntegrationId {
 	GitHubEnterprise = 'github-enterprise',
+	CloudGitHubEnterprise = 'cloud-github-enterprise',
 	GitLabSelfHosted = 'gitlab-self-hosted',
 }
 
@@ -19,6 +20,7 @@ export type IntegrationId = HostingIntegrationId | IssueIntegrationId | SelfHost
 
 export const supportedOrderedCloudIssueIntegrationIds = [IssueIntegrationId.Jira];
 export const supportedOrderedCloudIntegrationIds = [
+	SelfHostedIntegrationId.CloudGitHubEnterprise,
 	HostingIntegrationId.GitHub,
 	HostingIntegrationId.GitLab,
 	IssueIntegrationId.Jira,
--- a/src/constants.integrations.ts
+++ b/src/constants.integrations.ts
@@ -48,6 +48,12 @@ export const supportedCloudIntegrationDescriptors: IntegrationDescriptor[] = [
 		supports: ['prs', 'issues'],
 	},
 	{
+		id: SelfHostedIntegrationId.CloudGitHubEnterprise,
+		name: 'GitHub Enterprise',
+		icon: 'gl-provider-github',
+		supports: ['prs', 'issues'],
+	},
+	{
 		id: HostingIntegrationId.GitLab,
 		name: 'GitLab',
 		icon: 'gl-provider-gitlab',
--- a/src/constants.integrations.ts
+++ b/src/constants.integrations.ts
@@ -20,8 +20,8 @@ export type IntegrationId = HostingIntegrationId | IssueIntegrationId | SelfHost
 
 export const supportedOrderedCloudIssueIntegrationIds = [IssueIntegrationId.Jira];
 export const supportedOrderedCloudIntegrationIds = [
-	SelfHostedIntegrationId.CloudGitHubEnterprise,
 	HostingIntegrationId.GitHub,
+	SelfHostedIntegrationId.CloudGitHubEnterprise,
 	HostingIntegrationId.GitLab,
 	IssueIntegrationId.Jira,
 ];
--- a/src/constants.storage.ts
+++ b/src/constants.storage.ts
@@ -78,6 +78,7 @@ export type GlobalStorage = {
 	'launchpadView:groups:expanded': StoredLaunchpadGroup[];
 	'graph:searchMode': StoredGraphSearchMode;
 	'views:scm:grouped:welcome:dismissed': boolean;
+	'integrations:configured': StoredIntegrationConfigurations;
 } & { [key in `plus:preview:${FeaturePreviews}:usages`]: StoredFeaturePreviewUsagePeriod[] } & {
 	[key in `confirm:ai:tos:${AIProviders}`]: boolean;
 } & {
@@ -87,6 +88,19 @@ export type GlobalStorage = {
 } & { [key in `jira:${string}:organizations`]: Stored<StoredJiraOrganization[] | undefined> } & {
 	[key in `jira:${string}:projects`]: Stored<StoredJiraProject[] | undefined>;
 };
+
+export type StoredIntegrationConfigurations = Record<
+	string,
+	StoredConfiguredProviderAuthenticationDescriptor[] | undefined
+>;
+
+export interface StoredConfiguredProviderAuthenticationDescriptor {
+	cloud: boolean;
+	integrationId: IntegrationId;
+	domain?: string;
+	expiresAt?: string;
+	scopes: string;
+}
 
 export type DeprecatedWorkspaceStorage = {
 	/** @deprecated use `confirm:ai:tos:${AIProviders}` */
--- a/src/constants.storage.ts
+++ b/src/constants.storage.ts
@@ -89,12 +89,9 @@ export type GlobalStorage = {
 	[key in `jira:${string}:projects`]: Stored<StoredJiraProject[] | undefined>;
 };
 
-export type StoredIntegrationConfigurations = Record<
-	string,
-	StoredConfiguredProviderAuthenticationDescriptor[] | undefined
->;
+export type StoredIntegrationConfigurations = Record<string, StoredConfiguredIntegrationDescriptor[] | undefined>;
 
-export interface StoredConfiguredProviderAuthenticationDescriptor {
+export interface StoredConfiguredIntegrationDescriptor {
 	cloud: boolean;
 	integrationId: IntegrationId;
 	domain?: string;
--- a/src/env/node/git/sub-providers/remotes.ts
+++ b/src/env/node/git/sub-providers/remotes.ts
@@ -38,8 +38,11 @@ export class RemotesGitSubProvider extends RemotesGitProviderBase implements Git
 		let remotesPromise = this.cache.remotes?.get(repoPath);
 		if (remotesPromise == null) {
 			async function load(this: RemotesGitSubProvider): Promise<GitRemote[]> {
+				const ci = await this.container.cloudIntegrations;
+				const connections = await ci?.getConnections();
 				const providers = loadRemoteProviders(
 					configuration.get('remotes', this.container.git.getRepository(repoPath!)?.folder?.uri ?? null),
+					connections,
 				);
 
 				try {
@@ -48,7 +51,7 @@ export class RemotesGitSubProvider extends RemotesGitProviderBase implements Git
 						this.container,
 						data,
 						repoPath!,
-						getRemoteProviderMatcher(this.container, providers),
+						await getRemoteProviderMatcher(this.container, providers),
 					);
 					return remotes;
 				} catch (ex) {
--- a/src/env/node/git/sub-providers/remotes.ts
+++ b/src/env/node/git/sub-providers/remotes.ts
@@ -38,11 +38,9 @@ export class RemotesGitSubProvider extends RemotesGitProviderBase implements Git
 		let remotesPromise = this.cache.remotes?.get(repoPath);
 		if (remotesPromise == null) {
 			async function load(this: RemotesGitSubProvider): Promise<GitRemote[]> {
-				const ci = await this.container.cloudIntegrations;
-				const connections = await ci?.getConnections();
 				const providers = loadRemoteProviders(
 					configuration.get('remotes', this.container.git.getRepository(repoPath!)?.folder?.uri ?? null),
-					connections,
+					this.container.integrations.getConfiguredIntegrationDescriptors(),
 				);
 
 				try {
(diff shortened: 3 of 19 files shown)


sergeibbb:
@eamodio,  @axosoft-ramint,  @d13,

To sum up what is happening here: new type of integration "cloud-github-enterprise" that works as enterprise but gets authorization as cloud.

src/constants.integrations.ts
```diff
@@ -19,6 +20,7 @@ export type IntegrationId = HostingIntegrationId | IssueIntegrationId | SelfHost
 
 export const supportedOrderedCloudIssueIntegrationIds = [IssueIntegrationId.Jira];
 export const supportedOrderedCloudIntegrationIds = [
+	SelfHostedIntegrationId.CloudGitHubEnterprise,
```
eamodio: Should be moved down below `HostingIntegrationId.GitHub` as this is a sorted array

src/constants.integrations.ts
```diff
@@ -37,13 +39,22 @@ export interface IntegrationDescriptor {
 	name: string;
 	icon: string;
 	supports: IntegrationFeatures[];
+	weight?: number;
```
eamodio, line 42: Not sure I understand the usage of `weight` here, since `supportedOrderedCloudIntegrationIds` is already an ordered array, is this needed?

src/git/parsers/remoteParser.ts
```diff
@@ -6,11 +6,13 @@ import type { getRemoteProviderMatcher } from '../remotes/remoteProviders';
...
 const remoteRegex = /^(.*)\t(.*)\s\((.*)\)$/gm;
 
+type PromiseType<T> = T extends Promise<infer U> ? U : T;
+
 export function parseGitRemotes(
 	container: Container,
 	data: string,
 	repoPath: string,
-	remoteProviderMatcher: ReturnType<typeof getRemoteProviderMatcher>,
+	remoteProviderMatcher: PromiseType<ReturnType<typeof getRemoteProviderMatcher>>,
```
eamodio: Use the built-in `Awaited` type here

src/git/parsers/remoteParser.ts
```diff
@@ -6,11 +6,13 @@ import type { getRemoteProviderMatcher } from '../remotes/remoteProviders';
 
 const remoteRegex = /^(.*)\t(.*)\s\((.*)\)$/gm;
 
+type PromiseType<T> = T extends Promise<infer U> ? U : T;
```
eamodio: Remove

src/webviews/home/homeWebview.ts
```diff
@@ -830,22 +830,36 @@ export class HomeWebviewProvider implements WebviewProvider<State, State, HomeWe
...
+				if (a.weight !== undefined && b.weight !== undefined) {
+					return -(a.weight - b.weight);
+				}
+				return (
 					supportedOrderedCloudIntegrationIds.indexOf(a.id) -
-					supportedOrderedCloudIntegrationIds.indexOf(b.id),
-			);
+					supportedOrderedCloudIntegrationIds.indexOf(b.id)
+				);
+			});
```
eamodio, line 862: Not sure I understand why this changed/is needed.

src/git/remotes/remoteProviders.ts
```diff
@@ -97,6 +103,29 @@ export function loadRemoteProviders(cfg: RemotesConfig[] | null | undefined): Re
...
+					creator: providerCreator,
+				};
+
+				const indexOfCustomDuplication: number = providers.findIndex(p => p.matcher === matcher);
+
+				if (indexOfCustomDuplication !== -1) {
+					providers[indexOfCustomDuplication] = provider;
+				} else {
+					providers.push(provider);
+				}
```
eamodio, line 124: Not sure we want to override the custom configuration, shouldn't we just skip adding this provider if there is a custom one configured?

eamodio (approved):
Overall it looks good - though I did have a few questions.

src/constants.integrations.ts
sergeibbb: ok. Thank you

src/git/parsers/remoteParser.ts
sergeibbb: ok

src/constants.integrations.ts
sergeibbb: @eamodio 
Because here they are sorted by ID: https://github.com/gitkraken/vscode-gitlens/blob/7befb4b2f8e32c2be745df585e1c29802d222f4c/src/webviews/home/homeWebview.ts#L885-L889

and "cloud-github-enterprise" goes before the "github". But we want GitHub stay before GitHub Enterprize in the list, so we need some way to swap them.

src/webviews/home/homeWebview.ts
sergeibbb: @eamodio 

The array used to be sorted by ID, but we need some way to put "github" before "cloud-github-enterprise"

src/git/remotes/remoteProviders.ts
sergeibbb: I'm not sure about that. I would leave it this way, because it shows better coverage of different cases.

Imagine we have a custom configured but not connected. In this case (if we keep the custom one), here we will send user to GKDev to connect enterprise, but then ignore what they have connected.

In what I've done here:
- if custom is connected (we have a local PAT), it works.
- if it's configured but not connected, we send user to GK Dev to connect and use cloud connection.


**Now (replacing the custom creator):**
if they have local PAT and don't have it connected on cloud:



if they **don't have local PA** and have it connected on cloud:




**Not replacing the custom:**
if they don't have local PA and **have it connected on cloud:**





cc @justinrobots

src/webviews/home/homeWebview.ts
sergeibbb: @eamodio 

Ah. Another way of doing it is to use `index` of how integrations are placed in the `supportedOrderedCloudIntegrationIds` array. So, instead adding a new "weight" field we would use their original order to sort the list of collected items.

src/env/node/git/sub-providers/remotes.ts
```diff
@@ -38,8 +38,11 @@ export class RemotesGitSubProvider extends RemotesGitProviderBase implements Git
 		let remotesPromise = this.cache.remotes?.get(repoPath);
 		if (remotesPromise == null) {
 			async function load(this: RemotesGitSubProvider): Promise<GitRemote[]> {
+				const ci = await this.container.cloudIntegrations;
+				const connections = await ci?.getConnections();
```
axosoft-ramint, line 42: This is an API call to our backend. Have we checked to make sure this isn't happening very frequently, especially when scaled across all GL clients that update to this version?

src/constants.integrations.ts
```diff
@@ -7,6 +7,7 @@ export enum HostingIntegrationId {
 
 export enum SelfHostedIntegrationId {
 	GitHubEnterprise = 'github-enterprise',
+	CloudGitHubEnterprise = 'cloud-github-enterprise',
```
axosoft-ramint, line 10: I'm definitely very late to this conversation but, would you mind clarifying why this needs to be a different key and not just the same "github-enterprise" we use for the non-cloud one (your connection is either one or the other, right? but both are GH enterprise, just like how we treat regular github integrations).

src/git/remotes/remoteProviders.ts
```diff
@@ -136,12 +165,14 @@ function getCustomProviderCreator(cfg: RemotesConfig) {
...
-export function getRemoteProviderMatcher(
+export async function getRemoteProviderMatcher(
 	container: Container,
 	providers?: RemoteProviders,
-): (url: string, domain: string, path: string) => RemoteProvider | undefined {
+): Promise<(url: string, domain: string, path: string) => RemoteProvider | undefined> {
 	if (providers == null) {
-		providers = loadRemoteProviders(configuration.get('remotes', null));
+		const ci = await container.cloudIntegrations;
+		const c = await ci?.getConnections();
```
axosoft-ramint, line 174: Same concern here re: API calls.

src/webviews/home/homeWebview.ts
eamodio: Yeah, it is already sorted by the index in the array, but the order in that array was wrong, which I fixed in my commit.

src/git/remotes/remoteProviders.ts
eamodio: OK, it has the potential to change things for existing users with custom configuration, but makes sense.

src/constants.integrations.ts
eamodio: That is sorting by the index in the array, not by `id`

src/plus/integrations/authentication/integrationAuthentication.ts
```diff
@@ -94,27 +100,58 @@ abstract class IntegrationAuthenticationProviderBase<ID extends IntegrationId =
...
 	}
 
 	protected async writeSecret(
 		key: IntegrationAuthenticationKeys,
 		session: ProviderAuthenticationSession | StoredSession,
 	) {
 		await this.container.storage.storeSecret(key, JSON.stringify(session));
+		await this.authenticationService.addConfigured({
+			integrationId: this.authProviderId,
+			domain: isSelfHostedIntegrationId(this.authProviderId) ? session.id : undefined,
```
eamodio, line 118: This scares me a since we could in the future change the session.id to be/mean something else and we'd implicitly break this.

I would suggest we add `domain` on to the session like we are doing with `cloud` to make it explicit.

src/plus/integrations/authentication/models.ts
```diff
@@ -14,6 +14,14 @@ export interface ProviderAuthenticationSession extends AuthenticationSession {
 	readonly expiresAt?: Date;
 }
 
+export interface ConfiguredProviderAuthenticationDescriptor {
```
eamodio, line 17: Naming nit
```suggestion
export interface ConfiguredIntegrationDescriptor {
```

src/constants.storage.ts
```diff
@@ -88,6 +89,19 @@ export type GlobalStorage = {
 	[key in `jira:${string}:projects`]: Stored<StoredJiraProject[] | undefined>;
 };
 
+export type StoredIntegrationConfigurations = Record<
+	string,
+	StoredConfiguredProviderAuthenticationDescriptor[] | undefined
+>;
+
+export interface StoredConfiguredProviderAuthenticationDescriptor {
```
eamodio, line 97: Naming nit
```suggestion
export interface StoredConfiguredIntegrationDescriptor {
```

src/plus/integrations/authentication/integrationAuthentication.ts
```diff
@@ -94,27 +100,58 @@ abstract class IntegrationAuthenticationProviderBase<ID extends IntegrationId =
 
 	protected abstract restoreSession(sessionId: string): Promise<ProviderAuthenticationSession | undefined>;
 
-	protected async deleteSecret(key: IntegrationAuthenticationKeys) {
+	protected async deleteSecret(key: IntegrationAuthenticationKeys, sessionId: string) {
 		await this.container.storage.deleteSecret(key);
+		await this.authenticationService.removeConfigured({
+			integrationId: this.authProviderId,
+			domain: isSelfHostedIntegrationId(this.authProviderId) ? sessionId : undefined,
+		});
```
eamodio, line 108: Nit for another day, but it feels like these operations should be combined into the AuthService rather than in the base class.

src/plus/integrations/integrationService.ts
```diff
@@ -857,12 +932,33 @@ export class IntegrationService implements Disposable {
...
+	private findCachedById(id: SupportedSelfHostedIntegrationIds): Integration | undefined {
+		const key = this.getCacheKey(id, '');
+		for (const [k, integration] of this._integrations) {
+			if (k.startsWith(key)) {
+				return integration;
+			}
+		}
+		return undefined;
+	}
+
```
eamodio, line 944: @axosoft-ramint doesn't this still need to be fixed up to look in the configured set?

axosoft-ramint: @eamodio I added the configured set lookup in the case statement for `CloudGitHubEnterprise` in the `get` function of this file, rather than here. But we could just as easily move that lookup into this function.

src/plus/integrations/authentication/integrationAuthentication.ts
axosoft-ramint: That's a good idea, though we would want to include a migration for existing sessions to include that like we did with `cloud`. @sergeibbb

src/webviews/home/homeWebview.ts
sergeibbb: I've reverted this.

src/plus/integrations/authentication/models.ts
sergeibbb: fixed

src/constants.storage.ts
sergeibbb: fixed

src/plus/integrations/integrationService.ts
```diff
@@ -437,6 +467,45 @@ export class IntegrationService implements Disposable {
...
+							if (domain == null) throw new Error(`Domain is required for '${id}' integration`);
+							integration = new (
+								await import(/* webpackChunkName: "integrations" */ './providers/github')
+							).GitHubEnterpriseIntegration(
+								this.container,
+								this.authenticationService,
+								this.getProvidersApi.bind(this),
+								domain,
+								id,
+							);
```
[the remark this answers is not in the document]
sergeibbb: closing with:

> if any existed regardless of their domain, we would have never reached that point

src/plus/integrations/authentication/integrationAuthentication.ts
axosoft-ramint: @eamodio I'm going to add a TODO here and write this up as post-release debt

PR #3981, opened 2025-01-24
Updates home overview - stale states and performance

d13:
- Splits home overview state by active and inactive branches
- Fixes unneeded overview refresh after opening the switch repo picker
- Fixes unneeded overview refresh when subscription data doesn't change
- Fixes repo file changes not reflecting while home that was made while home was hidden

### DIFF
--- a/src/webviews/home/protocol.ts
+++ b/src/webviews/home/protocol.ts
@@ -28,12 +28,11 @@ export interface State extends WebviewState {
 	integrations: IntegrationState[];
 	avatar?: string;
 	organizationsCount?: number;
-	walkthroughProgress: {
+	walkthroughProgress?: {
 		doneCount: number;
 		allCount: number;
 		progress: number;
 	};
-	showWalkthroughProgress?: boolean;
 	previewEnabled: boolean;
 	newInstall: boolean;
 }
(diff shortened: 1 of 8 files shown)


src/webviews/home/protocol.ts
```diff
@@ -189,32 +184,57 @@ export interface GetOverviewBranch {
...
+	provider?: {
+		name: string;
+		icon?: string;
+		url?: string;
+	};
 }
 
-export type GetOverviewResponse =
+export interface GetActiveOverviewRequest {
+	[key: string]: unknown;
```
eamodio, line 199: These need to be typed

src/webviews/home/protocol.ts
```diff
@@ -189,32 +184,57 @@ export interface GetOverviewBranch {
...
 	| undefined;
-export const GetOverview = new IpcRequest<GetOverviewRequest, GetOverviewResponse>(scope, 'overview');
+
+export const GetActiveOverview = new IpcRequest<GetActiveOverviewRequest, GetActiveOverviewResponse>(
+	scope,
+	'overview/active',
+);
+
+export interface GetInactiveOverviewRequest {
+	[key: string]: unknown;
```
eamodio, line 216: Needs to be typed

eamodio:
Ultimately it would be good to split the set of branches and the real data of each branch, so that we can query independently

src/webviews/home/protocol.ts
d13: We don't actually send anything, I think I just forgot to delete these types.

Issue #3976, opened 2025-01-23
Azure DevOps Integration Support - Set Up Azure DevOps Auth Provider and Basic Integration


axosoft-ramint:
- ✅ Add Azure as a cloud auth provider
- ✅ Set up basic Azure integration with connect/disconnect
- ✅ Show Azure as a provider in home integrations list

(parent #3902)

### Testing notes

Connection flow and displaying in list of integrations on Home:

<img width="723" alt="Image" src="https://github.com/user-attachments/assets/c47e5525-b08d-4d6b-ad8b-c721c5a41fd1" />

Displaying the rich integration status in remotes:

<img width="769" alt="Image" src="https://github.com/user-attachments/assets/2d141044-8a06-4125-8284-9b1e9bbade3a" />

During testing check that:
- connect-disconnect flow through remotes works
- connect-disconnect flow through Home works

sergeibbb:
Hi @axosoft-ramint! 
Could you please review the changes: https://github.com/gitkraken/vscode-gitlens/pull/3984

d13:
I verified this as part of https://github.com/gitkraken/vscode-gitlens/issues/4024

PR #3984, opened 2025-01-24
Rich azure integration basics (#3976)

sergeibbb:
# Description
Implements #3976

# Checklist


src/plus/integrations/integrationService.ts
```diff
@@ -668,10 +668,10 @@ export class IntegrationService implements Disposable {
 
 		switch (remote.provider.id) {
 			// TODO: Uncomment when we support these integrations
-			// case 'azure-devops':
-			// 	return get(HostingIntegrationId.AzureDevOps) as RT;
 			// case 'bitbucket':
 			// 	return get(HostingIntegrationId.Bitbucket) as RT;
+			case 'azure-devops':
+				return get(HostingIntegrationId.AzureDevOps) as RT;
```
axosoft-ramint, line 674: We probably need to check the remote's domain here and only return the azure devops id if it's non-enterprise. We don't support azure enterprise yet and should probably return `undefined` for that.

src/constants.integrations.ts
```diff
@@ -29,6 +29,7 @@ export const supportedOrderedCloudIntegrationIds = [
 	SelfHostedIntegrationId.CloudGitHubEnterprise,
 	HostingIntegrationId.GitLab,
 	SelfHostedIntegrationId.CloudGitLabSelfHosted,
+	HostingIntegrationId.AzureDevOps,
```
axosoft-ramint, line 32: If this were to merge today, and this array was updated, would it be safe? Or would anything start trying to connect to azure and fail?

axosoft-ramint (commented):
Just one comment re:

src/constants.integrations.ts
sergeibbb: @axosoft-ramint 
I don't think it would connect to azure, because:

1. Mostly doesn't even try to call to Azure until it appears in `supportedIntegrations` lists of Launchpad, Start Work, etc.
2. All methods of Azure provided are implemented to do nothing:
image

src/plus/integrations/integrationService.ts
sergeibbb: @axosoft-ramint 
Done.
Something like this?
image

axosoft-ramint: I haven't tested it yet, but if you've verified by cloning an azure repo and making sure it passes this check, I think we should be good to go.

sergeibbb: Yes, I cloned both: a repo under `dev.azure.com` and another one under `*.visualstudio.com`

PR #3985, opened 2025-01-25
Moves configured integrations and session storage to new service

axosoft-ramint:
Major refactor (and fix to Home View integration states) that does the following:
 - Creates a new configuredIntegrationService
 - Moves all management of stored integration configurations as well as stored integration authentication sessions to the new service
 - Renames functions across auth providers and the new service to be clearer, and simplifies logic on the integration authentication providers where possible
 - Adds domain property to integration auth sessions
 - Removes redundant code that isn't used, and makes a session descriptor required when interacting with an auth provider
- Updates home view to use the new service to get configured integrations

We will want this in before adding other cloud integrations into the mix, like Azure.

### DIFF
--- a/src/plus/integrations/authentication/integrationAuthenticationProvider.ts
+++ b/src/plus/integrations/authentication/integrationAuthenticationProvider.ts
@@ -1,32 +1,18 @@
-import type { CancellationToken, Disposable, Event, Uri } from 'vscode';
-import { authentication, EventEmitter, window } from 'vscode';
+import type { Disposable, Event } from 'vscode';
+import { authentication, EventEmitter } from 'vscode';
 import type { IntegrationId } from '../../../constants.integrations';
 import { HostingIntegrationId } from '../../../constants.integrations';
-import type { IntegrationAuthenticationKeys } from '../../../constants.storage';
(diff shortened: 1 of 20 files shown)


src/plus/integrations/authentication/integrationAuthenticationProvider.ts
```diff
@@ -387,20 +272,23 @@ export abstract class CloudIntegrationAuthenticationProvider<
 
 		if (!session) return undefined;
 
+		// TODO: Once we care about domains, we should try to match the domain here against ours, and if it fails, return undefined
```
sergeibbb, line 275: @axosoft-ramint 

quick question. Which domains are "ours"? those that are stored in configuredIntegrationService?

sergeibbb (approved):

src/plus/integrations/authentication/integrationAuthenticationProvider.ts
axosoft-ramint: "Ours" are the ones we have locally stored and configured, as you guessed. While the other domain being compared here is the one returned by our api for the provider.

Issue #3977, opened 2025-01-23
Azure DevOps Integration Support - Add Home PR/Issue Support


axosoft-ramint:
Add support for seeing associated Azure PRs/issues for branches/commits in Home view, 

- ✅ autolinks on commits issues and pull requests: #3996
- ✅ you can see associated issues and prs on Home: #3999
  - ✅ pull requests
  - ✅ autolinks to PRs and issues of a branch (autodone by the previous PR)
  - ➡️ issues: It's possible to associate a branch to some issue, e.g. from GitHub, but the rest of it is waiting for #3979 
- 💡 you can see associated issues and prs on commit graph
- ❓ etc.

(parent #3902)

#### Issues and PRs on a commit

An open issue

<img width="600" alt="Image" src="https://github.com/user-attachments/assets/9d33d27b-7742-48d4-8459-db4752b4e521" />

A closed issue

<img width="600" alt="Image" src="https://github.com/user-attachments/assets/a6070433-fc5a-4ab0-a99d-5d245a0cbc50" />

An open PR

<img width="600" alt="Image" src="https://github.com/user-attachments/assets/7d712a91-c6f7-460a-a76d-08b0d22a3a89" />

A closed PR

<img width="600" alt="Image" src="https://github.com/user-attachments/assets/f1e611eb-77d5-47d6-860f-51413aa9f7a1" />

#### PRs and autolinks on Home

A PR on Home:
<img width="600" src="https://github.com/user-attachments/assets/7f07c441-da9f-4785-a720-78c0ec9d8c4b">

Enriched autolinks of PRs and issues on Home:
<img width="600" src="https://github.com/user-attachments/assets/db6592ef-b5c3-4621-91e6-3f4a0c551250">

#### PR and issue links go to their webpages
- Make sure they don't go to API
- Check links of PRs and issues separately (there are different pieces of code for that)

sergeibbb:
Hi @axosoft-ramint 

My work-in-progress commit of accessing Azure issue or PR is here: b1b10e79ff5fe11512a8410e8fdd3d597ae44909 
If you have a chance to make a quick look at that and share any thoughts, you're very welcome.

I've noticed that `States` are customizable and we need to rely on `stateCategory`. So, now I'm going to retrieve and cache state-map for the current project.

PR #3996, opened 2025-01-30
Enriches Azure autolinks on commits: issues and pull requests

sergeibbb:
# Description

#3977 

#### Issues and PRs on a commit

An open issue

Image

A closed issue

Image

An open PR

Image

A closed PR

Image

# Checklist

### DIFF
--- a/src/autolinks/autolinks.ts
+++ b/src/autolinks/autolinks.ts
@@ -6,7 +6,9 @@ import { IssueIntegrationId } from '../constants.integrations';
 import type { Container } from '../container';
 import type { GitRemote } from '../git/models/remote';
 import { getIssueOrPullRequestHtmlIcon, getIssueOrPullRequestMarkdownIcon } from '../git/utils/-webview/icons';
-import type { HostingIntegration, IssueIntegration } from '../plus/integrations/integration';
+import type { HostingIntegration, Integration, IssueIntegration } from '../plus/integrations/integration';
+import { IntegrationBase } from '../plus/integrations/integration';
+import { remoteProviderIdToIntegrationId } from '../plus/integrations/integrationService';
 import { configuration } from '../system/-webview/configuration';
 import { fromNow } from '../system/date';
 import { debug } from '../system/decorators/log';
@@ -214,9 +216,26 @@ export class Autolinks implements Disposable {
 
 		const enrichedAutolinks = new Map<string, EnrichedAutolink>();
 		for (const [id, link] of messageOrAutolinks) {
-			let linkIntegration = link.provider
-				? await this.container.integrations.get(link.provider.id as IntegrationId)
-				: undefined;
+			let integrationId: IntegrationId | undefined;
+			let linkIntegration: Integration | undefined;
+			if (link.provider != null) {
+				// Try to make a smart choice
+				integrationId =
+					link.provider instanceof IntegrationBase
+						? link.provider.id
+						: remoteProviderIdToIntegrationId(link.provider.id);
+				if (integrationId == null) {
+					// Fall back to the old logic assuming that integration id might be saved as provider id.
+					// TODO: it should be removed when we put providers and integrations in order. Conversation: https://github.com/gitkraken/vscode-gitlens/pull/3996#discussion_r1936422826
+					integrationId = link.provider.id as IntegrationId;
+				}
+				try {
+					linkIntegration = await this.container.integrations.get(integrationId);
+				} catch (e) {
+					Logger.error(e, `Failed to get integration for ${link.provider.id}`);
+					linkIntegration = undefined;
+				}
+			}
 			if (linkIntegration != null) {
 				const connected = linkIntegration.maybeConnected ?? (await linkIntegration.isConnected());
 				if (!connected || !(await linkIntegration.access())) {
@@ -226,7 +245,7 @@ export class Autolinks implements Disposable {
 			const issueOrPullRequestPromise =
 				remote?.provider != null &&
 				integration != null &&
-				link.provider?.id === integration.id &&
+				integrationId === integration.id &&
 				link.provider?.domain === integration.domain
 					? integration.getIssueOrPullRequest(
 							link.descriptor ?? remote.provider.repoDesc,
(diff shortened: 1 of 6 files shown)


sergeibbb:
@axosoft-ramint 

Enriched autolinks work. I'm switching to Home. Do you think we can merge what's done without the rest of #3977?
Here are my current changes #3996

src/autolinks/autolinks.ts
```diff
@@ -214,9 +214,9 @@ export class Autolinks implements Disposable {
 
 		const enrichedAutolinks = new Map<string, EnrichedAutolink>();
 		for (const [id, link] of messageOrAutolinks) {
-			let linkIntegration = link.provider
-				? await this.container.integrations.get(link.provider.id as IntegrationId)
-				: undefined;
+			const integrationId = link.provider ? remoteProviderIdToIntegrationId(link.provider.id) : undefined;
```
sergeibbb, line 217: Here with this new converting function we'll need to make sure that other types of enriched autolinks are not broken. What do we have? Jira, GitHub Cloud and Ent, GitLab Cloud and Ent. Anything else to check?

axosoft-ramint: In some cases like in Jira, we pass the entire integration in as a provider reference, so `remoteProviderIdToIntegrationId` will fail here. I think in the short term here, we can first try treating `link.provider.id` as an integration id and try to get the integration from that, and put that in a try/catch so if it fails, then we can try the solution here (try treating it as a remote provider id and convert first before getting the integration).

But in the long term, we need to distinguish, on pull requests, issues, and autolinks, the difference between a remote provider reference and an integration reference. They should be treated separately and put into different properties. If we cannot accomplish that here, we should treat it as a follow-up task.

src/plus/integrations/integrationService.ts
```diff
@@ -1029,13 +1029,8 @@ export class IntegrationService implements Disposable {
 	}
 }
 
-export function remoteProviderIdToIntegrationId(
```
axosoft-ramint, line 1032: We should avoid setting the input typing to `unknown` here. I know we messed up by treating integrations as remote provider references in several places, but it would be better to work toward addressing that directly rather than loosening the typing here.

src/autolinks/autolinks.ts
axosoft-ramint: My short term idea would be something like:

```suggestion
			const integrationId: IntegrationId | undefined = link.provider instanceof Integration ? link.provider.id : isRemoteProviderReference(link.provider) ? remoteProviderIdToIntegrationId(link.provider.id) : undefined;
			let linkIntegration = integrationId != null ? await this.container.integrations.get(integrationId) : undefined;
```

But in long term, we should make sure we never put an Integration into `provider` property, reserve that for remote providers, and instead have a separate `integration` property or something that is an integration reference.

src/plus/integrations/providers/azure/models.ts
```diff
@@ -0,0 +1,144 @@
...
+	owner: string,
+	projectName: string,
+	repoName: string,
+	pullRequestId: number,
+): string {
+	return `${baseUrl}/${owner}/${projectName}/_git/${repoName}/pullrequest/${pullRequestId}`;
+}
+
+export interface AzurePullRequest {
+	repository: unknown;
```
axosoft-ramint, line 107: This needs a type. See https://learn.microsoft.com/en-us/rest/api/azure/devops/git/pull-requests/get-pull-request?view=azure-devops-rest-7.1#gitpullrequest for all pull request properties and subtypes.

src/plus/integrations/providers/azure/models.ts
```diff
@@ -0,0 +1,144 @@
...
+	mergeId: string;
+	lastMergeSourceCommit: {
+		commitId: string;
+		url: string;
+	};
+	lastMergeTargetCommit: {
+		commitId: string;
+		url: string;
+	};
+	reviewers: unknown[];
```
axosoft-ramint, line 128: Please add and define the appropriate subtypes

src/plus/integrations/providers/azure/models.ts
```diff
@@ -0,0 +1,144 @@
...
+	baseUrl: string,
+	owner: string,
+	projectName: string,
+	repoName: string,
+	pullRequestId: number,
+): string {
+	return `${baseUrl}/${owner}/${projectName}/_git/${repoName}/pullrequest/${pullRequestId}`;
+}
+
+export interface AzurePullRequest {
```
axosoft-ramint, line 106: Missing properties here:
`autoCompleteSetBy`
`closedBy`
`commits`
`completionOptions`
`completionQueueTime`
`forkSource`
`hasMultipleMergeBases`
`labels`
`lastMergeCommit`
`mergeFailureMessage`
`mergeFailureType`
`mergeOptions`
`mergeStatus`
`workItemRefs`

You can find their descriptions and types here: https://learn.microsoft.com/en-us/rest/api/azure/devops/git/pull-requests/get-pull-request?view=azure-devops-rest-7.1#gitpullrequest

axosoft-ramint: Haven't checked but please make sure we have as complete as possible information on Azure issue (work item) typings as well.

axosoft-ramint (changes requested):
Will likely break autolinks for several integrations including Jira, but not really your fault here - we just did a very poor job putting both integrations and remote providers into the same `provider` property just because they both match the very loose `providerreference` interface.

We should improve on that. I offer a short-term idea and long-term suggestion below, up to you which you think you can get to for this release.

src/plus/integrations/integrationService.ts
sergeibbb: @axosoft-ramint Allow it to be a`string`?
If I require it to be `RemoteProviderId` it will cause type casting in `autolinks.ts` where we don't really know an actual type.

src/autolinks/autolinks.ts
sergeibbb: updated

src/plus/integrations/integrationService.ts
sergeibbb: updated

src/plus/integrations/providers/azure/models.ts
sergeibbb: fixed

src/plus/integrations/providers/azure/models.ts
sergeibbb: fixed

src/plus/integrations/providers/azure/models.ts
sergeibbb: fixed

axosoft-ramint (approved):

PR #4012, opened 2025-02-03
Updates to new Org Members route

d13:


src/plus/integrations/authentication/configuredIntegrationService.ts
```diff
@@ -27,6 +29,11 @@ export type ConfiguredIntegrationType = 'cloud' | 'local';
 export class ConfiguredIntegrationService {
 	private _configured?: Map<IntegrationId, ConfiguredIntegrationDescriptor[]>;
 
+	private readonly _onDidChangeConfiguredIntegrations = new EventEmitter<void>();
+	get onDidChangeConfiguredIntegrations(): Event<void> {
+		return this._onDidChangeConfiguredIntegrations.event;
+	}
```
axosoft-ramint, line 35: This should probably suffice for listeners that care about any change, but in the near future I'll probably update this to emit the changed integration ids and how they changed (removed, added, etc.).

src/plus/gk/models/organization.ts
```diff
@@ -29,7 +29,18 @@ export enum OrganizationType {
...
 export interface OrganizationMember {
+	readonly id: string;
+	readonly email: string;
+	readonly name: string;
+	readonly username: string;
+	readonly role: OrganizationRole;
+	readonly status: OrganizationMemberStatus;
+}
+
+export interface OrganizationMemberOld {
```
axosoft-ramint, line 43: If we're not using this anywhere else, maybe we just chuck it? Or at least comment that it's deprecated and should be removed soon?

axosoft-ramint (approved):
LGTM, just minor nitpicks. Make sure we don't merge this though until that route is available in all environments, including prod

src/plus/integrations/authentication/configuredIntegrationService.ts
d13: Bah, this was supposed to be stashed. Removing

src/plus/gk/models/organization.ts
d13: Missed it, thanks.

PR #4034, opened 2025-02-10
Adds async loaded product config & more

eamodio:
Adds async loaded product config
Refactors promo handling via product config
Moves all required content into promo model
Adds percentile support to promos


axosoft-ramint (approved):
LGTM. Once the percentile value is added into global telemetry, I think we're good to go.

eamodio:
Added in latest commit

d13 (approved):

Issue #4039, opened 2025-02-11
Launchpad view can cause an unwanted "add remote" prompt on load


axosoft-ramint:
If you are on a branch that has a PR where you don't have the remote associated with its head ref, Launchpad view can show an "Unable to open details for PR #${pullRequest.id} because of a missing remote." prompt and ask you to add a remote.

This occurs because in Launchpad view, we always fully expand the "current branch" PR node, and expanding the PR node means finding its remote/repo info. Usually the GitHub PR extension hides this issue because it retrieves the virtual repo.

To reproduce:
1. Have GitHub PR extension disabled.
2. Be on a branch with a PR where you don't have the remote of the headref.
3. Load GitLens with Launchpad view open and visible.

We should suppress this and not expand the PR node when the remote isn't available.

PR #4081, opened 2025-02-20
4039 Launchpad view can cause an unwanted "add remote" prompt on load

axosoft-ramint:
Closes [#4039](https://github.com/gitkraken/vscode-gitlens/issues/4039)

Two fixes, each one commit:
- Fixes false "current branch" detection using the base repository ("current branch" and "remote" must use the head repository, but "open repo" can still use either)
- Instead of prompting, returns a command message node which can be clicked to add the missing remote.

### DIFF
--- a/src/views/viewCommands.ts
+++ b/src/views/viewCommands.ts
@@ -22,7 +22,9 @@ import * as StashActions from '../git/actions/stash';
 import * as TagActions from '../git/actions/tag';
 import * as WorktreeActions from '../git/actions/worktree';
 import { GitUri } from '../git/gitUri';
+import type { PullRequest } from '../git/models/pullRequest';
 import { RemoteResourceType } from '../git/models/remoteResource';
+import type { Repository } from '../git/models/repository';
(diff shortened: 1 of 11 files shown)


src/views/viewCommands.ts
```diff
@@ -481,6 +485,16 @@ export class ViewCommands implements Disposable {
 		return RemoteActions.add(getNodeRepoPath(node));
 	}
 
+	@log()
+	private async addPullRequestRemote(node: ViewNode, pr: PullRequest, repo: Repository) {
+		const identity = getRepositoryIdentityForPullRequest(pr);
+		if (identity.remote?.url == null) return;
+		await repo.git
+			.remotes()
+			.addRemoteWithResult?.(identity.provider.repoDomain, identity.remote.url, { fetch: true });
```
eamodio, line 494:
```suggestion
			.addRemote?.(identity.provider.repoDomain, identity.remote.url, { fetch: true });
```

Since we aren't returning the remote we don't need to wait for it

eamodio (approved):

Issue #4046, opened 2025-02-12
Bitbucket Cloud: Add Launchpad Support


sergeibbb:
- Add support for seeing Bitbucket PRs in Launchpad.

(parent #3916, related #3995, a follow-up #4128)

# Testing notes

- [ ] Bitbucket pull requests are loaded and categorized
- [ ] Globe icon shows "Open on Bitbucket" and opens the right page
    <img width="400" alt="Image" src="https://github.com/user-attachments/assets/9d9fbabc-4e5c-4cf0-9eea-0040edcf2550" />
- [ ] Items are categorized and the current branch is detected
  - <img width="400" alt="Image" src="https://github.com/user-attachments/assets/b3f098e6-3831-4d79-8ed9-bf34818d8842" />
  - Review requested:
      <img width="300" alt="Image" src="https://github.com/user-attachments/assets/153753c1-0a83-48c4-a360-3762ec2ce5ae"><img width="300" alt="Image" src="https://github.com/user-attachments/assets/8a0b223b-ca34-4e9b-8490-a5fc4d971232">
  - Approved:
      <img width="300" alt="Image" src="https://github.com/user-attachments/assets/4a9afb06-96b1-4bdc-afb6-f697d17000e0"><img width="300" alt="Image" src="https://github.com/user-attachments/assets/3df7ffc0-f7bf-4c43-817e-57379c8f836a">
  - Changes requested:
      <img width="300" alt="Image" src="https://github.com/user-attachments/assets/f4181e6c-305e-4286-b9e0-93864d98f8c4"><img width="300" alt="Image" src="https://github.com/user-attachments/assets/e0571eda-e9d0-4f4c-bf65-2df3d38f3ff8">

- [ ] Item actions work:
  - [ ] Open in Worktree
  - [ ] Switch to Branch
  - [ ] Open in Commit Graph
  - [ ] Merge
- [ ] Pin and snooze work:
  - Pinning, snoozing, unpinning, unsnoozing in GitLens gets reflected on GKDev
  - Pinning, snoozing, unpinning, unsnoozing on GKDev gets reflected in GitLens
  - It's reflected in Launchpad (especially after a restart), 
  - On Home make sure snoozed stuff doesn't enter the summary:

https://github.com/user-attachments/assets/17e3e88a-8e2f-486e-b020-6c65057a1bf4

PR #4099, opened 2025-02-24
Bitbucket: Launchpad

sergeibbb:
# Description
solves #4046

# Checklist

The follow-up for improvement of the shared library: #4128

### DIFF
--- a/src/plus/integrations/providers/bitbucket.ts
+++ b/src/plus/integrations/providers/bitbucket.ts
@@ -7,17 +7,18 @@ import type { IssueOrPullRequest, IssueOrPullRequestType } from '../../../git/mo
 import type { PullRequest, PullRequestMergeMethod, PullRequestState } from '../../../git/models/pullRequest';
 import type { RepositoryMetadata } from '../../../git/models/repositoryMetadata';
 import type { IntegrationAuthenticationProviderDescriptor } from '../authentication/integrationAuthenticationProvider';
-import type { ResourceDescriptor } from '../integration';
+import type { ProviderAuthenticationSession } from '../authentication/models';
 import { HostingIntegration } from '../integration';
-import { providersMetadata } from './models';
+import type {
+	BitbucketRemoteRepositoryDescriptor,
+	BitbucketRepositoryDescriptor,
+	BitbucketWorkspaceDescriptor,
+} from './bitbucket/models';
+import type { ProviderPullRequest } from './models';
+import { fromProviderPullRequest, providersMetadata } from './models';
 
 const metadata = providersMetadata[HostingIntegrationId.Bitbucket];
 const authProvider = Object.freeze({ id: metadata.id, scopes: metadata.scopes });
-
-interface BitbucketRepositoryDescriptor extends ResourceDescriptor {
-	owner: string;
-	name: string;
-}
 
 export class BitbucketIntegration extends HostingIntegration<
 	HostingIntegrationId.Bitbucket,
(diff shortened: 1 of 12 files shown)


src/plus/integrations/providers/bitbucket.ts
```diff
@@ -136,11 +153,148 @@ export class BitbucketIntegration extends HostingIntegration<
...
+					repo.owner,
+					repo.name,
+					this.apiBaseUrl,
+				),
+			),
+		);
+		return prsResult
+			.map(r => getSettledValue(r))
+			.filter(r => r != null)
+			.flat();
```
axosoft-ramint, line 297: Why not just use `getPullRequestsForUserAndWorkspace` from the shared provider library here, mapping over your workspaces (which is one call per workspace and includes paging) and then converting from `ProviderPullRequest` to `PullRequest` with the transformation function instead of doing it ourselves on a per-repo basis?

sergeibbb: @axosoft-ramint 

Because they use the [built-in Bitbucket API call](https://developer.atlassian.com/cloud/bitbucket/rest/api-group-workspaces/#api-workspaces-workspace-pullrequests-selected-user-get) that  it returns only PRs where I'm the author, not a reviewer.
image


Another reason is that they don't return the information about reviewers and participants:
image

sergeibbb: @axosoft-ramint 

I've updated it here: 81ddf674e25f971572d7fb1066de7635e7b62189

- It calls for authored per workspace
- and for reviewing it calls for active repositories only

src/plus/integrations/providers/bitbucket.ts
```diff
@@ -136,11 +139,104 @@ export class BitbucketIntegration extends HostingIntegration<
...
-		_session: AuthenticationSession,
-		_repos?: BitbucketRepositoryDescriptor[],
+		session: ProviderAuthenticationSession,
+		requestedRepositories?: BitbucketRepositoryDescriptor[],
 	): Promise<PullRequest[] | undefined> {
-		return Promise.resolve(undefined);
+		if (requestedRepositories != null) {
+			// TODO: implement repos version
+			return undefined;
+		}
```
axosoft-ramint, line 201: The "requested repositories" would always be your open bitbucket repositories - they can be calculated here rather than passed in from above.

src/plus/integrations/providers/bitbucket.ts
```diff
@@ -136,11 +139,104 @@ export class BitbucketIntegration extends HostingIntegration<
...
+		if (user?.username == null) return undefined;
+
+		const workspaces = await this.getProviderResourcesForUser(session);
+		if (workspaces == null || workspaces.length === 0) return undefined;
+
+		const api = await this.container.bitbucket;
+		if (!api) return undefined;
+
+		const authoredPrs = workspaces.map(ws =>
+			api.getPullRequestsForWorkspaceAuthoredByUser(this, session.accessToken, user.id, ws.slug, this.apiBaseUrl),
```
axosoft-ramint, line 218: We should not need to implement this locally if the shared library already has a function for it. If we made improvements on the shared library version, we should PR to there rather than re-implement here.

I thought the plan we had settled on was this:
1. Use the shared library function to get PRs authored by user for each workspace.
2. Use the repo-specific query to get the user's assigned PRs (that was previously in this PR but appears to have been removed?) only for open bitbucket repositories in the current workspace.

sergeibbb: @axosoft-ramint 

> We should not need to implement this locally if the shared library already has a function for it. If we made improvements on the shared library version, we should PR to there rather than re-implement here.

Sounds like another subtask?

src/plus/integrations/providers/bitbucket.ts
sergeibbb: @axosoft-ramint 
Do you mean that if they are passed, then I don't extract them from the context another time and just take the values from the argument?

axosoft-ramint: I mean we don't need `requestedRepositories` as an input here. This function should just use open repositories for the repo-specific query, and all your workspaces for the workspace-specific query.

src/plus/integrations/providers/bitbucket.ts
axosoft-ramint: I would expect that in this PR, we use a combination of `getPullRequestsForUserAndWorkspace` from the shared library and the repo-specific query you had from before (the former using our workspaces cache and the latter using open repositories) and return a deduplicated list of PRs from the combined outputs of the two.

sergeibbb: @axosoft-ramint 

The shared library does not provide info about reviewers, therefore it needs to be extended. But I'm afraid of ruining the world by extending the library, so it needs to be reviewed and tested well. The whole process of accepting and testing changes may take a lot of time. That's why I thought that this PR could go with our implementation, so we achieve our goal and provide the feature. And when the main job is done we can move our update of the `getPullRequestsForUserAndWorkspace` to the shared library: in another subtask.

src/plus/integrations/providers/bitbucket.ts
sergeibbb: Ah. Just to ignore its content regardless if it's undefined or defined?

src/plus/integrations/providers/bitbucket.ts
axosoft-ramint: Ah, I see. In that case, we can keep the local implementation for now and work on fixing the shared library one ASAP as follow up. But why aren't we using the repo-specific query that you had previously implemented, for open repositories here and combining it with the workspace one?

axosoft-ramint (approved):
LGTM. Please make sure to follow up with shared library team on improvements to bitbucket queries.

src/plus/integrations/providers/bitbucket.ts
axosoft-ramint: My mistake again here - we can keep this as is, just rename `requestedRepositories` to `repos` to align with the other providers.

sergeibbb:
@axosoft-ramint 

> Please make sure to follow up with shared library team on improvements to bitbucket queries.

Here it is: #4128

src/plus/integrations/providers/bitbucket.ts
sergeibbb: @axosoft-ramint 

https://github.com/gitkraken/vscode-gitlens/commit/b0ef7487e597adec9c70114df251cffdb4e5ba12 ?

PR #4125, opened 2025-03-04
Adds squash merge detection and merge target actions

### LINKED ISSUE #4124: Add squash merge detection and merged target actions
### Description

In home view, detect when a branch's commits have been squash-merged or other cases besides standard merge into its merge target.

On the hover in merged cases, offer the ability to:
- Push the merge target to its upstream when it is behind locally
- Delete the merged branch/worktree.

axosoft-ramint:
Closes #4124

Note: there are known issues with the "delete worktree" case where when the worktree is deleted before deleting the branch, the "delete branch" portion of the commands fail to proceed because the commands are using a worktree path that no longer exists. These should be addressed in follow-up.

### DIFF
--- a/src/webviews/home/homeWebview.ts
+++ b/src/webviews/home/homeWebview.ts
@@ -1174,18 +1174,24 @@ export class HomeWebviewProvider implements WebviewProvider<State, State, HomeWe
 
 	private async deleteBranchOrWorktree(ref: BranchRef, mergeTarget?: BranchRef) {
 		const repo = this._repositoryBranches.get(ref.repoPath);
-		const branch = repo?.branches.find(b => b.id === ref.branchId);
+		let branch = repo?.branches.find(b => b.id === ref.branchId);
+		if (branch == null) {
+			branch = await repo?.repo.git.branches().getBranch(ref.branchId);
+		}
+
 		const worktree = branch ? repo?.worktreesByBranch.get(branch.id) : undefined;
 		if (branch == null) return;
+
 		if (branch.current && mergeTarget != null && (!worktree || worktree.isDefault)) {
 			const mergeTargetLocalBranchName = getBranchNameWithoutRemote(mergeTarget.branchName);
 			const confirm = await window.showWarningMessage(
-				`The merge target branch ${mergeTargetLocalBranchName} will be checked out before deleting the current branch ${branch.name}.`,
+				`Before deleting the current branch '${branch.name}', you will be switched to '${mergeTargetLocalBranchName}'.`,
 				{ modal: true },
-				{ title: 'Proceed' },
+				{ title: 'Continue' },
 			);
 
-			if (confirm == null || confirm.title !== 'Proceed') return;
+			if (confirm == null || confirm.title !== 'Continue') return;
+
 			await this.container.git.checkout(ref.repoPath, mergeTargetLocalBranchName);
 
 			void executeGitCommand({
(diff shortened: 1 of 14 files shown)


src/webviews/home/homeWebview.ts
```diff
@@ -1167,6 +1172,87 @@ export class HomeWebviewProvider implements WebviewProvider<State, State, HomeWe
...
+		const repo = this._repositoryBranches.get(ref.repoPath);
+		const branch = repo?.branches.find(b => b.id === ref.branchId);
+		const worktree = branch ? repo?.worktreesByBranch.get(branch.id) : undefined;
+		if (branch == null) return;
+		if (branch.current && mergeTarget != null && (!worktree || worktree.isDefault)) {
+			const mergeTargetLocalBranchName = getBranchNameWithoutRemote(mergeTarget.branchName);
+			const confirm = await window.showWarningMessage(
+				`The merge target branch ${mergeTargetLocalBranchName} will be checked out before deleting the current branch ${branch.name}.`,
+				{ modal: true },
+				{ title: 'Proceed' },
```
eamodio, line 1185: Use "Continue" rather than "Proceed"

Other updated language suggestions
```suggestion
				`Before deleting the current branch '${branch.name}', '${mergeTargetLocalBranchName}' will be checked out.`,
				{ modal: true },
				{ title: 'Continue' },
```

or 

```suggestion
				`Before deleting the current branch '${branch.name}', you will be switched to '${mergeTargetLocalBranchName}'.`,
				{ modal: true },
				{ title: 'Continue' },
```

I think I prefer the latter.

src/webviews/home/homeWebview.ts
```diff
@@ -1167,6 +1172,87 @@ export class HomeWebviewProvider implements WebviewProvider<State, State, HomeWe
...
+			});
+		} else if (repo != null && branch != null && worktree != null && !worktree.isDefault) {
+			const commonRepo = await repo.repo.getCommonRepository();
+			const defaultWorktree = [...repo.worktreesByBranch.values()].find(w => w.isDefault);
+			if (defaultWorktree == null || commonRepo == null) return;
+
+			const confirm = await window.showWarningMessage(
+				`You will be returned to the default worktree before deleting the worktree for ${branch.name}.`,
+				{ modal: true },
+				{ title: 'Proceed' },
```
eamodio, line 1207: Use "Continue"

Maybe more like:
```suggestion
				`Before deleting the worktree for '${branch.name}', you will be switched to the default worktree.`,
				{ modal: true },
				{ title: 'Continue' },
```

eamodio (approved):
Looks good! Approved with minor tweaks

src/webviews/home/homeWebview.ts
```diff
@@ -1167,6 +1172,87 @@ export class HomeWebviewProvider implements WebviewProvider<State, State, HomeWe
 		});
 	}
 
+	private async deleteBranchOrWorktree(ref: BranchRef, mergeTarget?: BranchRef) {
```
eamodio, line 1175: Is there a confirmation for the delete happening or just the "Proceed" warnings below? If there is no other confirmations, then I think the wording need to be changed from what is below or suggested.

src/webviews/home/homeWebview.ts
axosoft-ramint: I chose the latter.

PR #4126, opened 2025-03-04
Converts the graph view to web-components

d13:
Minimizes the use of React down to just the shared Graph component.

You will need to add `"gitlens.graph.experimental.renderer.enabled": true` in your settings.json to this new version.


eamodio (approved):
Looks good! Some issues with multiselect, but we'll fix that up after the release

Issue #4107, opened 2025-02-26
Bitbucket Server


sergeibbb:
### Description

- parent #3916
- probably related:
  - #3901 (#3932)
  - #3923

### Testing notes:

- [ ] Remotes view:
    <img width="400" src="https://github.com/user-attachments/assets/8ad7be19-a9ab-4a67-8b9d-f4fbfaf6be13">
- [ ] Enriched autolinks on hover of commits

PR #4146, opened 2025-03-12
Bitbucket Data Center cloud integration

sergeibbb:
# Description

solves #4107 

# Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
   - shows enriched links to PRs and issues [#4045](https://github.com/gitkraken/vscode-gitlens/issues/4045)
   - shows Bitbucket PRs in Launchpad [#4046](https://github.com/gitkraken/vscode-gitlens/issues/4046)
   - supports Bitbucket issues in Start Work and lets associate issues with branches [#4047](https://github.com/gitkraken/vscode-gitlens/issues/4047)
+  - supports Bitbucket Data Center in remotes, in autolinks and on the Launchpad [#4107](https://github.com/gitkraken/vscode-gitlens/issues/4107)
 - Adds ability to control how worktrees are displayed in the views
   - Adds a `gitlens.views.worktrees.worktrees.viewAs` setting to specify whether to show worktrees by name, path, or relative path
   - Adds a `gitlens.views.worktrees.branches.layout` setting to specify whether to show branch worktrees as a list or tree, similar to branches
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,11 +11,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 - Adds AI model status and model switcher to the _Home_ view ([#4064](https://github.com/gitkraken/vscode-gitlens/issues/4064))
 - Adds Anthropic Claude 3.7 Sonnet model for GitLens' AI features ([#4101](https://github.com/gitkraken/vscode-gitlens/issues/4101))
 - Adds Google Gemini 2.0 Flash-Lite model for GitLens' AI features ([#4104](https://github.com/gitkraken/vscode-gitlens/issues/4104))
-- Adds integration with Bitbucket Cloud ([#3916](https://github.com/gitkraken/vscode-gitlens/issues/3916))
+- Adds integration with Bitbucket Cloud and Data Center ([#3916](https://github.com/gitkraken/vscode-gitlens/issues/3916))
   - shows enriched links to PRs and issues [#4045](https://github.com/gitkraken/vscode-gitlens/issues/4045)
-  - shows Bitbucket PRs in Launchpad [#4046](https://github.com/gitkraken/vscode-gitlens/issues/4046)
-  - supports Bitbucket issues in Start Work and lets associate issues with branches [#4047](https://github.com/gitkraken/vscode-gitlens/issues/4047)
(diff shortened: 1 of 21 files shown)


src/plus/integrations/providers/bitbucket-server.ts
```diff
@@ -0,0 +1,256 @@
...
+		super(container, authenticationService, getProvidersApi);
+	}
+
+	get domain(): string {
+		return this._domain;
+	}
+
+	protected get apiBaseUrl(): string {
+		const protocol = this._session?.protocol ?? 'https:';
+		return `${protocol}//${this.domain}/rest/api/1.0`;
```
sergeibbb, line 47: Hi @ericf-axosoft, @axosoft-ramint 

URL of our test server is on HTTP, that's why I needed to make an update to gather the protocol from Cloud. These changes are here: a7a792a2b1722e79fab91a5fe55c7b7d2def5169

sergeibbb:
Hi @ericf-axosoft. I've unblocked it

sergeibbb:
Hi @eamodio. I've unblocked it

src/git/remotes/remoteProviders.ts
```diff
@@ -76,6 +76,12 @@ const builtInProviders: RemoteProviders = [
 	},
 ];
 
+const cloudRemotesMap: { [key: string]: typeof GitHubRemote | typeof GitLabRemote | typeof BitbucketServerRemote } = {
```
d13, line 79: Nit for stricter typing
```suggestion
const cloudRemotesMap: Record<CloudSelfHostedIntegrationId, typeof GitHubRemote | typeof GitLabRemote | typeof BitbucketServerRemote> = {
```

sergeibbb: @d13 

Fixed: 9d18d8bdb9efd6f652816489783f73c901e83530

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
   - shows enriched links to PRs and issues [#4045](https://github.com/gitkraken/vscode-gitlens/issues/4045)
   - shows Bitbucket PRs in Launchpad [#4046](https://github.com/gitkraken/vscode-gitlens/issues/4046)
   - supports Bitbucket issues in Start Work and lets associate issues with branches [#4047](https://github.com/gitkraken/vscode-gitlens/issues/4047)
+  - supports Bitbucket Data Center in remotes, in autolinks and on the Launchpad [#4107](https://github.com/gitkraken/vscode-gitlens/issues/4107)
```
axosoft-ramint, line 18: I would update the top-level line item to include Data Center

src/plus/integrations/providers/providersApi.ts
```diff
@@ -564,6 +582,26 @@ export class ProvidersApi {
 		}
 	}
 
+	async getBitbucketServerPullRequestsForCurrentUser(
```
axosoft-ramint, line 585: I'm guessing the reason we are defining this separately rather than making it an overload of `GetPullRequestsForUserFn` is because we do not need a user identifier like "username" or "user id", correct?

src/plus/integrations/providers/bitbucket.ts
```diff
@@ -275,13 +275,14 @@ export class BitbucketIntegration extends HostingIntegration<
 		return issueResult;
 	}
 
+	private readonly storagePrefix = 'bitbucket';
```
axosoft-ramint, line 278: I'm guessing this `storagePrefix` was made when we were setting up the bitbucket server model as an extension or override on the base bitbucket model, but since it's an entirely separate model now, do we need this anymore?

src/plus/integrations/providers/bitbucket-server/models.ts
```diff
@@ -0,0 +1,231 @@
...
+	return reviews.reduce(
+		(prev: GitPullRequestReviewState, review) =>
+			reviewDecisionWeightByReviewState[review.state] > reviewDecisionWeightByReviewState[prev]
+				? review.state
+				: prev,
+		GitPullRequestReviewState.Approved,
+	);
+};
+
+export const normalizeBitbucketServerPullRequest = (pr: BitbucketServerPullRequest): ProviderPullRequest => {
```
axosoft-ramint, line 139: I think in at least one other place, we convert to our local `PullRequest` model first, then use `toProviderPullRequest`. Any reason we cannot or shoud not do that here?

axosoft-ramint (commented):
Mostly questions as I scanned through

src/plus/integrations/providers/providersApi.ts
sergeibbb: Right: no userId, no repos, no workspace.

src/plus/integrations/providers/bitbucket.ts
sergeibbb: @axosoft-ramint 
> do we need this anymore?

It lets us to reduce amount of copying the string value and forgetting to change it. E.g. if somewhere in DataCenter code I write `'bitbucket...` instead of `'bitbucket-server...` the type system does not help me because both prefix are valid. And by using a value baked into a constant I reduce probability of such mistake.

src/plus/integrations/providers/bitbucket-server/models.ts
sergeibbb: @axosoft-ramint 

> Any reason we cannot or shoud not do that here?

My opinion. The best solution is to make `normalizeBitbucketServerPullRequest` public in the library. Then I could just re-use that method here.

axosoft-ramint: Got it. We can go with this normalizer then and update when the library exposes theirs.

PR #4238, opened 2025-04-16
Improves branch name autolink matching logic

### LINKED ISSUE #3894: Improve issue matching with branch names
The branch for this issue is `release-notes-16-1` and GitLens thinks these two issues (`#16` and `#1`) are associated but they are not. I suspect the matching algorithm needs to be refined to avoid mismatches like this.
 
Image

axosoft-ramint:
Refactors the branch name autolink matching logic to improve accuracy and reduce false positives.

- Introduces new regex matching rules/heuristics for branch name matching based on prefix and issue number patterns:

1. If a ref is prefixed, match to a 2+ digit number with the prefix followed by a connector `(/|-|_|.)` or end-of-string and take the first match encountered..
2. If ref is not prefixed, match to a 2+ digit number in the pattern of `(feature|feat|fix|bug|bugfix|hotfix|issue|ticket)(/|-|_|#|/#|-#|_#)<num>` followed by a connector `(/|-|_|.)` or end-of-string and take the first match.
3. If ref is not prefixed and no matches from rule 2, match to a 3+ digit number that is either the full branch name, or is preceded or followed by at least two non-slash (`/`), non-numeric characters. Note: this is to rule out numbers that are full sections like `<num>/...`, `.../<num>/...`, or `.../<num>` and to rule out date-formats like `<num.<num2>` or `<num>-<num2>`.

- Filters remote autolinks for branch names to only include non-dynamic autolinks of type 'branch'.
- Fixes issue where integration autolinks were incorrectly collected for branch autolinking.
- Sorts refsets so that issue integrations are checked first for matches.
- Limits branch autolink matches to 1 maximum per branch name.

Closes #3894

### DIFF
--- a/src/autolinks/__tests__/autolinks.test.ts
+++ b/src/autolinks/__tests__/autolinks.test.ts
@@ -24,24 +24,34 @@ function assertAutolinks(actual: Map<string, Autolink>, expected: Array<string>)
 
 suite('Autolinks Test Suite', () => {
 	test('Branch name autolinks', () => {
-		assertAutolinks(getBranchAutolinks('123', mockRefSets()), ['test/123']);
-		assertAutolinks(getBranchAutolinks('feature/123', mockRefSets()), ['test/123']);
+		// Matches under rule 1 (prefixed 2+ digit number followed by end of string)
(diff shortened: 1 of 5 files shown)


d13 (approved):
LGTM

PR #4245, opened 2025-04-22
Improves handling of 'delete worktree' action on merge target when the default worktree is already open

### LINKED ISSUE #4232: Attemping to delete the current worktree from Home fails if the default worktree is already open
### Description

Clicking on the below button, fails if there already is an open VS Code window with the default worktree

Image

The VS Code window will be shown, but nothing will happen.

axosoft-ramint:
https://github.com/gitkraken/vscode-gitlens/issues/4232

When using the "delete worktree" action on the merge target hover, closes the current window and hands the rest of the flow over to the existing window with the default worktree, whenever it exists already.

To make this happen:
1. Fixes a bug in `findCommonRepository` where we are passing a '.git' folder path into the lower-level rev-parse git ops, which ultimately results in an error from git (to repro in any terminal, navigate to the .git directory of a repo and try `git rev-parse --show-toplevel`. You will see a `fatal: this operation must be run in a work tree` error.)
2. Pending deep links are moved to secret storage so that GitLens in other windows can detect when they are stored externally.
3. A listener is added to the deep link service to respond to pending deep links in storage whenever they match a repo open in the current window, and for the time being, only for pending deep links with the "delete worktree" action, to minimize possible regression risk.
4. A delayed command is triggered to close the window that initiated the handoff.

This carries a great deal of regression risk with deep links, so we should check for regressions where we can.

Closes #4232

### DIFF
--- a/src/constants.commands.ts
+++ b/src/constants.commands.ts
@@ -147,6 +147,7 @@ export type CoreCommands =
 	| 'workbench.action.closeActiveEditor'
 	| 'workbench.action.closeAllEditors'
 	| 'workbench.action.closePanel'
+	| 'workbench.action.closeWindow'
 	| 'workbench.action.focusRightGroup'
 	| 'workbench.action.nextEditor'
 	| 'workbench.action.newGroupRight'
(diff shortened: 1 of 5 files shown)


d13 (approved):
LGTM

PR #4272, opened 2025-04-30
Uses PAT on shared provider library requests from Azure

### LINKED ISSUE #4271: Azure calls can fail or return incorrect info for some accounts
### Description

We use our token with bearer auth on most of our calls to shared provider library for Azure. This can cause issues. For example, with my API tokens (from gkdev), asking for Azure orgs without using the `isPAT: true` option and without converting my token to a PAT, results in my projects being returned instead of my orgs. Other calls result in 401s from Azure.

GKDev converts the token to a PAT and uses `isPAT: true` on all Azure calls except for `getCurrentUser`, which doesn't support using a PAT. 

We should match this as long as it doesn't break any calls for existing tokens that worked before using bearer auth.

### GitLens Version

17.0.3

### VS Code Version

_No response_

### Git Version

_No response_

_No response_

axosoft-ramint:
Fixes #4271

Forces the use of a PAT (and converts the token to PAT) on almost all shared library calls from Azure, except for `getCurrentUser` which does not support it.

Note that we may need to update local queries as well using the same logic.

### DIFF
--- a/src/plus/integrations/providers/azure/azure.ts
+++ b/src/plus/integrations/providers/azure/azure.ts
@@ -24,6 +24,7 @@ import { Logger } from '../../../../system/logger';
 import type { LogScope } from '../../../../system/logger.scope';
 import { getLogScope } from '../../../../system/logger.scope';
 import { maybeStopWatch } from '../../../../system/stopwatch';
+import { base64 } from '../../../../system/string';
 import type {
 	AzureProjectDescriptor,
 	AzurePullRequest,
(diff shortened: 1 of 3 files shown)


axosoft-ramint:
@sergeibbb Since your integration currently works without these changes, please verify that they do not break any of the existing Azure DevOps functionality. You will probably need to clear your stored Azure account, organizations, projects, etc. to test it properly.

eamodio (approved):
I pushed a commit to avoid `btoa` and uses our base64 function

sergeibbb:
@axosoft-ramint 

ok. I'll test it tomorrow

axosoft-ramint:
Going to merge this and test in pre-release.

sergeibbb:
@axosoft-ramint 

I confirm that on my old Azure integration it keeps receiving my organizations with the new code. I've tested retrieving data for Launchpad, get-pr-for-commit, get-account-for-commit.

PR #4277, opened 2025-05-02
Adds user setting to disable AI features

d13:
To disable AI features, add `gitlens.ai.enabled: false` to your settings


contributions.json
```diff
@@ -25,11 +25,11 @@
 		},
 		"gitlens.ai.explainCommit": {
 			"label": "Explain Commit...",
-			"commandPalette": "gitlens:enabled && !gitlens:readonly && !gitlens:untrusted && gitlens:gk:organization:ai:enabled",
+			"commandPalette": "gitlens:enabled && !gitlens:readonly && !gitlens:untrusted && gitlens:gk:organization:ai:enabled && config.gitlens.ai.enabled",
```
eamodio, line 28: I think we should add a post-release follow up task to unify these into a single context key -- at least for the majority of cases where we don't care about the difference

src/quickpicks/aiModelPicker.ts
```diff
@@ -33,6 +34,16 @@ export async function showAIProviderPicker(
...
+	if (!configuration.get('ai.enabled')) {
+		await window.showQuickPick([{ label: 'OK' }], {
+			title: 'AI is Disabled',
+			placeHolder: 'GitLens AI features have been disabled via settings',
+			canPickMany: false,
+		});
+
+		return undefined;
+	}
+
```
eamodio, line 46: Please move this into a helper shared with the below. And it would be create to have an entry to re-enable the setting.

src/webviews/apps/plus/shared/components/integrations-chip.ts
```diff
@@ -377,7 +385,8 @@ export class GlIntegrationsChip extends LitElement {
 						</span>`
 				: html`<span class="integration__content">
 						<span class="integration_details"
-							>GitLens AI features have been disabled by your GitKraken admin</span
+							>GitLens AI features have been
+							disabled${!this.aiSettingEnabled ? ' via settings' : ' by your GitKraken admin'}</span
```
eamodio, line 389: Similar to above, would be great to have a button to re-enable.

eamodio (approved):
Added some comments but we can merge and do them in a follow up

contributions.json
d13: https://github.com/gitkraken/vscode-gitlens/issues/4283

src/quickpicks/aiModelPicker.ts
d13: Done

src/webviews/apps/plus/shared/components/integrations-chip.ts
d13: Done

Issue #4297, opened 2025-05-15
Sync does not update sessions


sergeibbb:
## Steps to reproduce

1. have two Azure accounts.
2. open a repo of a 1st Azure account in VSCode.
3. open GK.dev. Connect to the 2nd Azure account. Sync GitLens with GK.dev
4. Make sure you cannot get the information about the repository pull-requests.
5. open GK.dev. Disconnect the 2nd Azure account. Connect it to the 1st account.
6. Sync GitLens with GK.dev
7. try to reach out the information about the repository pull-requests.

**Expected result**
GitLens is logged in to the 1st Azure account. It can get the information about thre repository pull-requests.

**Actual result**
GitLens keeps being logged in to the 2nd Azure account. It fails getting the information about thre repository pull-requests.

I think that the problem is that `syncCloudConnection` does not refresh token if it thinks that the current session is valid. I believe it can be reproduced with other types of integrations too.

## Workaround
When there is an old session that you cannot sync with GK.dev the following helps:

1. Disconnect the integration on GK.dev
2. Sync
3. Connect the integration on GK.dev
4. Sync

d13:
This was verified with the 17.1.1 patch

Issue #4192, opened 2025-03-31
Get "associated PR for commit" working on all providers that don't have it


sergeibbb:
### Roadmap
- getPullRequestForCommit
  - [x] azure
  - [x] bitbucket
  - [x] Show message about enabling an app for Bitbucket and sending to docs
  - [x] bitbucket DC
- getAccountForCommit
  - [x] azure
  - [x] bitbucket
  - [x] bitbucket DC

Related PRs:
- gitkraken/gitlens-docs#109
- #4243

### Description

Associated PRs don't show up on commits because getProviderPullRequestForCommit isn't implemented (for BitBucket, Azure, possibly others)

![Image](https://github.com/user-attachments/assets/cefa06e9-c45e-4ce6-b976-ec4d1438e3fe)

For Bitbucket.org we need to warn user that an action is required in order to get PR info:

<img width="400" src="https://github.com/user-attachments/assets/185122bb-3e5d-44b6-8475-6a5f3b69d382">

# Verification steps
- connect Azure integration
- go to GitLens branches or remotes views
- open branches one by one until it hits `getAccountForCommit`
  - check that it receives the account
  - check that it does not live author avatar without any image (it could happen when if we use Azure's avatar URLs which reject anonymous access)
- open commits one by one
  - check that it hits `getPullRequestForCommit`
  - repeat until some commits give a PR result
  - and others don't give any result
- repeat these steps for Bitbucket and Bitbucket DC

d13:
@sergeibbb I’m not able to see Pull Requests for an Azure repo. I’m getting `Your 'azureDevOps' credentials are either invalid or expired` even though I just auth’ed. 


```
AzureDevOpsApi.getPullRequestForCommit  AuthenticationError: Your 'azureDevOps' credentials are either invalid or expired
```

![Image](https://github.com/user-attachments/assets/6b632af4-301b-4b49-a877-c95fcfbd530f)

sergeibbb:
@d13 

Does launchpad show you Azure PRs?

d13:
@sergeibbb no

sergeibbb:
@d13 

I've looked at it. I suppose that the problem is not related to this ticket.

Steps to reproduce:
1. have two Azure accounts.
2. open a repo of a 1st Azure account in VSCode.
3. open GK.dev. Connect to the 2nd Azure account. Sync GitLens with GK.dev
4. Make sure you cannot get the information about the repository pull-requests.
[...]

PR #4324, opened 2025-05-21
Sync cloud integration on Authorization problems

sergeibbb:
# Description
Here we sync the cloud connected integration with GKDev when we face an authorization problem hoping that during the sync they respond us with the actual token.

If we see that we keep facing the same error we don't try to sync again. Instead we disconnect the integration soon letting user know that they need to take an action to reconnect.

Related to:
- #4297
- #4192

## Follow-ups:
- #4496

# Verification steps

- [ ] Simple stop syncing
  2. Connect GitHub-com integration on GKDev
  3. [Reject the OAuth applicaiton](https://github.com/settings/applications) on GitHub's side, by keeping it on GKDev
  4. Open launchpad and refresh its content
  5. Check that GitLens stops trying to refresh the invalid token (stops auto-syncing)
- [ ] Keeps being stopped after manual sync when token has not been refreshed
  6. Continue the previous scenario
  7. Manually sync the integrations, make sure GitHub is still highlighted as connected
  8. Open launchpad and refresh its content
  9. Check that GitLens does not try sync GitHub integration
- [ ] Manual sync recovers the integration when it is fixed on GKDev
  9. Continue the previous scenario
  10. Go to GKDev and re-connect GitHub integration
  11. Got to GitLens and sync integrations manually
  12. Open launchpad and refresh its content
  13. Check that GitHub integration started working
- [ ] Tries to sync once after it was blocked and then recovered
  14. Continue the previous scenario
  15. Go to [GitHub settings](https://github.com/settings/applications) and reject the application again
  16. Go to GKDev and re-connect: so GKDev has a new token and GitLens yet has the old token.
  17. Open launchpad and refresh its content twice
  18. Check that the GitLens has refreshed the token and shows the Launchpad content
- [ ] Lack of permissions: it does not sync continuously when some requests succeed and others fail.
  - ❓ I wanted to connect a self-hosted integrations (e.g. GitHub Enterprise) with lack of permissions . But I failed. With whatever limitations of the token, I was able to get public PRs. I tested on GitHub Enterprise, GitLab Server, Bitbucket Server. If you find the exact way to create such token, then please let me know.
  - Another way to check it is to make a change in code to alwasy fail PRs and check that it does not try to sync again and again (that it stops syncing)
- [ ] We do not message user due to errors in Launchpad (if the user errors too many times, and folks whose integration was having issues with Launchpad, due to too many PRs, large/complex repos, etc.)

# Checklist

### DIFF
--- a/src/plus/integrations/models/integration.ts
+++ b/src/plus/integrations/models/integration.ts
@@ -278,7 +278,8 @@ export abstract class IntegrationBase<
 		}
 
 		switch (state) {
-			case 'connected':
+			case 'connected': {
+				const oldSession = this._session;
 				if (forceSync) {
 					// Reset our stored session so that we get a new one from the cloud
 					const authProvider = await this.authenticationService.get(this.authProvider.id);
@@ -301,8 +302,14 @@ export abstract class IntegrationBase<
 
 				// sync option, rather than createIfNeeded, makes sure we don't call connectCloudIntegrations and open a gkdev window
 				// if there was no session or some problem fetching/refreshing the existing session from the cloud api
-				await this.ensureSession({ sync: forceSync });
+				const newSession = await this.ensureSession({ sync: forceSync });
+
+				if (oldSession && newSession && newSession.accessToken !== oldSession.accessToken) {
+					this.resetRequestExceptionCount('all');
+				}
+
 				break;
+			}
 			case 'disconnected':
 				await this.disconnect({ silent: true });
 				break;


src/plus/integrations/models/integration.ts
```diff
@@ -270,7 +282,17 @@ export abstract class IntegrationBase<
...
 		Logger.error(ex, scope);
 
-		if (ex instanceof AuthenticationError || ex instanceof RequestClientError) {
+		if (ex instanceof AuthenticationError && this._session?.cloud) {
+			if (this.syncCountDueToRequestException < IntegrationBase.syncDueToRequestExceptionLimit) {
+				this.syncCountDueToRequestException++;
+				this._session = {
+					...this._session,
+					expiresAt: new Date(Date.now() - 1),
+				};
```
ramin-t: I have a hypothetical scenario I want to check with you here.

Let's say I have a GitHub token which has the required scopes to get my current account but not the scopes to get PRs.

So, every time the Launchpad automatically fetches PRs (let's say, every 5 minutes), it first gets the current account, which succeeds (and also resets the sync count) and then it tries to get PRs and fails because of the scope issue, throws an AuthenticationError, and we force it to sync.

This would cause us to sync every time the Launchpad fetches PRs which would hit our backend pretty hard and probably is not the intention, but..is it possible? If we can confirm that I am worrying for nothing, I think we are ok to merge.

sergeibbb: @axosoft-ramint 

That makes sense. It's hard to reproduce, probably should be reproduced on a self-hosted server where the integration is connected using PAT.

Anyway, I've implemented the fix here: 4e1b4116b1178db0a5b5bf0d82fe84cfae8ebbb1 when a success in one use-case doesn't cancel failure in another case, so we avoid continuous syncing. 

Also I've noticed that when GKDev keeps returning expired key we keep refreshing it in `refreshSessionIfExpired()`, so I added another fix: 3c064009c3264291831034820dd927a28df574c6

sergeibbb: @axosoft-ramint 

better solution for unsuccessful syncing of expired tokens: b178a037686bbfac46ac4d75874fc74a9df1dc91

src/plus/integrations/models/gitHostIntegration.ts
```diff
@@ -663,10 +683,13 @@ export abstract class GitHostIntegration<
 				cancellation,
 				silent,
 			);
+			this.resetRequestExceptionCount('searchMyPullRequests');
 			return { value: pullRequests, duration: Date.now() - start };
 		} catch (ex) {
-			Logger.error(ex, scope);
-			return { error: ex, duration: Date.now() - start };
+			return this.handleProviderException('searchMyPullRequests', ex, scope, {
```
ramin-t: Sorry I'm catching this a bit late, but: we moved `searchMyPullRequests` off of `handleProviderException` on purpose, specifically because it results in an error notification in VS Code if the user errors too many times, and folks whose integration was having issues with Launchpad (due to too many PRs, large/complex repos, etc.) were seeing the notification all the time since Launchpad query automatically runs in an interval.

We need to make sure we do not send a message in this case.

axosoft-ramint (changes requested):

src/plus/integrations/models/integration.ts
```diff
@@ -265,12 +309,27 @@ export abstract class IntegrationBase<
...
+		scope: LogScope | undefined,
+		defaultValue: T,
+	): T {
 		if (ex instanceof CancellationError) return defaultValue;
 
 		Logger.error(ex, scope);
 
-		if (ex instanceof AuthenticationError || ex instanceof RequestClientError) {
+		if (ex instanceof AuthenticationError && this._session?.cloud) {
+			if (!this.hasSessionSyncRequests()) {
```
ramin-t, line 329: Just want to note here that once we have a "session sync request" for a particular case like "getIssuesForProject", we will not pass this check again until either:
1. that specific function has had a chance to run successfully, or
2. the integration disconnects
So for a three hour session, if the token lasts an hour and we run into this once for some specific query, we will go back to the default non-sync failures for other queries if that specific query is not used again after syncing the first time.

Not a big deal I suppose, and not sure there is a perfect solution here - just making sure we are aware of the limitations even though this is better than the previous version,

src/plus/integrations/models/gitHostIntegration.ts
sergeibbb: @axosoft-ramint 

> RE: we moved searchMyPullRequests off of handleProviderException on purpose, specifically because it results in an error notification in VS Code if the user errors too many times, and folks whose integration was having issues with Launchpad (due to too many PRs, large/complex repos, etc.)

What do you think about adding `silent` flag: b4db45d16cdbceb108169c107ebc1e4a03915c75 ?
(It's done after a little refactoring: 07116b42462e1cbc0cba1a4217cd475b49347ea9)

src/plus/integrations/models/integration.ts
sergeibbb: @axosoft-ramint 

> we will not pass this check again until either:

What do you think about this: add09ccd55a6e097eec0ce569c657a746d1753de

It should work like this:
- we detect an auth error on `getIssuesForProject`
- we expire it locally in our memory to force it resync
- Then one of the following happens:
  - if it's not actually expired GK.dev returns us the same token, so we don't reset the flags
  - if GK.dev refreshes token and returns a new one, we reset all the flags, to be able to see how it works.

I haven't tested it thoroughly yet, but if you approve the approach I go ahead with it.

sergeibbb:
Hi @axosoft-ramint 
all recent questions on this has been attended.

src/plus/integrations/models/integration.ts
ramin-t: This approach sounds ok. Once you've tested it, if it seems like it works, let me know.

PR #4338, opened 2025-05-29
Supports AI Controls for GK Organizations

### LINKED ISSUE #4300: Support AI Controls for GK Organizations
### Description

Support AI controls for GK Organizations that want to specify the AI providers their users have access to, including:
- Self-hosted AI control for the whole org (specify a URL path)
- Provide their own key for the whole org to use for a provider (or self-hosted AI)
- Block specific AI providers so no one in the org can use them (not specific models)

Refer to the [Product Brief](https://docs.google.com/document/d/1kpM6CbT1EYi4nR8v886f08By-DbYEMM4VBCes8Gs0lc/edit?usp=sharing) for details regarding this change.

[API Spec](https://gitkraken.atlassian.net/wiki/spaces/ENGR/pages/982450180/Custom+AI+Providers+in+organizations)

### Client Requirements as of May 15:
- For each provider, when a key is provided, all traffic to that provider should use that key
- For each provider, when a URL is provided, all traffic to that provider should use that URL
- ~~When a provider is disabled, it should be disabled (non selectable)~~
  - ~~Ideally this should be accompanied with a message explaining the admin disabled it at an org level~~
  - "disabled" means missing, e.g. not in the UI at all -- because the user can't do anything to re-enable it. So if OpenAI is disabled, OpenAI will not be shown in the picker at all.
- If the data is invalid for any reason, then the client will enter an error state. In this state, the existence of a problem will be indicated in the AI preferences, and the client will not allow any AI features whatsoever whenever that org is selected, until the client receives good data.
  - Example: we add a new AI provider, which the customer chooses. One user fails to update their client in a timely manner, and their client doesn’t recognize the provider. In this case the user won’t be able to use AI features until they update.

### Verification steps
Every time organization settings are changed, you should reload VSCode window.

- [ ] `aiEnabled` -> `false`
  - [ ] All AI features are disabled
  - [ ] "Switch AI provider" menu doesn't exist
- [ ] `aiEnabled` -> `true` (\<- here and below) `enforceAiProviders` -> `false`
  - [ ] AI features are enabled, for the first usage it opens the "Select AI model" menu from VSCode provider
  - [ ] "Switch AI provider" menu lets user to select any supported provider and their models.
  - [ ] Nothing from `aiProviders` configuration is applied. Meaning that user has to provide their own URLs and API keys.
- [ ] `enforceAiProviders` -> `true`
  - [ ] "Switch AI provider" menu lets user to select only configured (and enabled) providers
  - [ ] AI features are enabled, for the first usage it opens the "Select AI provider" menu where only configured providers
- [ ] When `enforceAiProviders == true` and there are zero enabled providers, it's treated as `aiEnabled == false`
- [ ] Switching `enforceAiProviders` from `false` to `true` OR disabling a provider that earlier has been enabled causes "Switch AI provider" menu to pop up on AI actions.
  1. Select an enabled provider. Check that you can use AI features, such as "Explain commig" or "Generate commit message"
  2. Disable the selected provider in Org configurations.
  3. Start using an AI feature.
  - [ ] Expected result: during using the feature it asks you to select another Provider.
- [ ] Configured URL is used
  1. `enforceAiProviders` -> false
  2. Configure your Ollama, Azure or OpenAI Compatible provider with your URL. E.g. `http://localhost:11434`
  3. Check that AI features work
  4. `enforceAiProviders` -> `true`. Configure another URL for your provider. I suggest you use a broken url, e.g. `http://localhost:99999`.
  - [ ] Expected result: AI features stopped working because the AI server cannot be reached.
  5. `enforceAiProviders` -> back to `false`
  - [ ] Expected result: AI features started working again with your server.
  6. You can also repeat steps above using working URL for org settings and a broken one for your settings
- [ ] Configured API key is used
  1. `enforceAiProviders` -> false
  2. Configure your OpenAI or any other server your API key.
  3. Check that AI features work
  4. `enforceAiProviders` -> `true`. Configure another API key for your provider. I suggest you use a broken API key, e.g. `aaaaaaaa`.
  - [ ] Expected result: AI features stopped working because the AI server rejects requests.
  5. `enforceAiProviders` -> back to `false`
  - [ ] Expected result: AI features started working again with your server.
  6. You can also repeat steps above using working URL for org settings and a broken one for your settings
 - [ ] Cached in local storage
   1. Run GitLens and load the configuration where AI providers are limited
   2. Go offline. Reload window.
   - [ ] Expected result: AI features work with the same limitations.
- [ ] Malformed settings
   - [ ] No `enforceAiProviders` flag in org configuration means the feature is not implemented yet, so treat it as `false`
   - [ ] malformed structure of `aiProviders` object probably means that the feature implementation has been changed at the last moment. I suggest we treat it as "not implemented yet", so turn the `enforceAiProviders` to `false`.
   - We probably want to change it in future.

sergeibbb:
# Description

Support AI Controls for GK Organizations. Fixes #4300 

1. Supports GK Organization AI config:
    1. Disables non-configured AI providers
    2. Uses configured URLs and API keys
2. Caches GK Organization settings, so it can be used after window reload
3. When all Organizations are disabled it's treated as AI feature is disabled

# Checklist

### DIFF
--- a/src/plus/gk/organizationService.ts
+++ b/src/plus/gk/organizationService.ts
@@ -153,11 +153,14 @@ export class OrganizationService implements Disposable {
 			}
 		}
 
+		const enforceAiProviders = settings?.enforceAiProviders ?? false;
+		const disabledByEnforcing = enforceAiProviders && !Object.values(aiProviders).some(p => p.enabled);
+
 		void setContext(
 			'gitlens:gk:organization:ai:enabled',
-			settings?.aiSettings.enabled ?? settings?.aiEnabled ?? true,
+			(!disabledByEnforcing && settings?.aiSettings.enabled) ?? settings?.aiEnabled ?? true,
 		);
-		void setContext('gitlens:gk:organization:ai:enforceProviders', settings?.enforceAiProviders ?? false);
+		void setContext('gitlens:gk:organization:ai:enforceProviders', enforceAiProviders);
 		void setContext('gitlens:gk:organization:ai:providers', aiProviders);
 		void setContext('gitlens:gk:organization:drafts:byob', settings?.draftsSettings.bucket != null);
 		void setContext('gitlens:gk:organization:drafts:enabled', settings?.draftsSettings.enabled ?? true);


src/plus/ai/aiProviderService.ts
```diff
@@ -325,12 +325,18 @@ export class AIProviderService implements Disposable {
...
-			const vsCodeModels = await this.getModels('vscode');
-			if (vsCodeModels.length !== 0) {
-				chosenProviderId = 'vscode';
-			} else if ((await this.container.subscription.getSubscription()).account?.verified) {
+			if (isProviderEnabledByOrg('vscode', orgAiConf)) {
+				const vsCodeModels = await this.getModels('vscode');
+				if (vsCodeModels.length !== 0) {
+					chosenProviderId = 'vscode';
+				}
+			} else if (
```
ramin-t: This nested `if` statement changes the previous logic such that if we do not find a vscode model, it does not move to the `else` block and therefore does not choose the GK model. I suggest moving the existing vscode models check into the `if` statement to make sure that the original flow is preserved.

i.e.
```suggestion
                        const vsCodeModels = await this.getModels('vscode');
			if (isProviderEnabledByOrg('vscode', orgAiConf) && vsCodeModels.length !== 0) {
					chosenProviderId = 'vscode';
			} else if (
```

axosoft-ramint (changes requested):
Overall looks good, just one issue I noticed. This will need heavy testing once merged.

src/plus/ai/aiProviderService.ts
sergeibbb: @axosoft-ramint 
Agree. Fixed.

PR #4380, opened 2025-06-09
Handles cancellation of AI commands

### LINKED ISSUE #4354: Improve Explain error messaging
When an "Explain (Branch|Commit|Stash|WIP)" command is ran, the wrong toast messaging can occur. 

Example, if the operation is cancelled due to gating, the toast message will show "No changes to explain".

axosoft-ramint:
Fixes #4354 

- Modifies return output of `sendRequest` and callers up the chain to return (and handle) `'cancelled'` when a cancellation occurs.
- Returns without showing an error message on command-level logic when receiving `'cancelled'` as output.

### DIFF
--- a/src/env/node/git/commitMessageProvider.ts
+++ b/src/env/node/git/commitMessageProvider.ts
@@ -62,7 +62,10 @@ class AICommitMessageProvider implements CommitMessageProvider, Disposable {
 				},
 			);
 
+			if (result === 'cancelled') return;
+
 			if (result == null) return;
+
 			return `${currentMessage ? `${currentMessage}\n\n` : ''}${result.parsed.summary}${
 				result.parsed.body ? `\n\n${result.parsed.body}` : ''
 			}`;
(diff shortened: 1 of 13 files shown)


src/env/node/git/commitMessageProvider.ts
```diff
@@ -62,7 +62,10 @@ class AICommitMessageProvider implements CommitMessageProvider, Disposable {
 				},
 			);
 
+			if (result === 'cancelled') return;
+
 			if (result == null) return;
+
```
d13: Nit that's in a few files, otherwise looks good.
```suggestion
			if (result == null || result === 'cancelled') return;

```

ramin-t: Fixed in commit: https://github.com/gitkraken/vscode-gitlens/pull/4380/commits/ffb03f585cd4d0d851dd2670cd36f0986fd8ecb6

Issue #4387, opened 2025-06-11
Clicking the repo icon to collapse the repo name on the current work item section of home doesn't work


axosoft-ramint:
### Description
it's a follow-up of #4332

![Image](https://github.com/user-attachments/assets/10c74b8c-89f6-4bf3-bae1-1ec8136bdec6)

### GitLens Version

2025.6.1105 (pre-release)

### VS Code Version

Version: 1.100.3 (user setup)
Commit: 258e40fedc6cb8edf399a463ce3a9d32e7e1f6f3
Date: 2025-06-02T13:30:54.273Z
Electron: 34.5.1
ElectronBuildId: 11369351
Chromium: 132.0.6834.210
Node.js: 20.19.0
V8: 13.2.152.41-electron.0
OS: Windows_NT x64 10.0.26100

sergeibbb:
@d13 
Assigning to you because you [said](https://gitkraken-hq.slack.com/archives/C06QFFUSVUZ/p1749677356132669?thread_ts=1749671016.133499&cid=C06QFFUSVUZ) that you're going to take it

d13:
Resolved with f31b51ac877dfeb9e975575bfbf212ffb0f82895

ramin-t:
Verified

PR #4411, opened 2025-06-18
Enhances remote provider connection flow

sergeibbb:
# Description

This is a follow-up on #4387

Originates from [this conversation](https://gitkraken-hq.slack.com/archives/C06QFFUSVUZ/p1750157934439699?thread_ts=1750096212.425849):

> And there is another thing that I think is worth fixing: handling of the "connect" click should be different. I'm attaching the example how it's handled in Remotes View and in Home. I think on Home it should work the same way as on Remotes View.

How it should work (without a web page, using existing connection):

https://github.com/user-attachments/assets/7fb47859-2025-4782-90cc-716606009293

How it works (it passes user through a web page) and what this pull request changes:

image

# Verification steps

- [ ] basic change: open "remotes", disconnect, open "Home", connect: it should not go through the browser.
- [ ] right remote is selected
  - switch to a branch, that has an upstream from `origin`
    image

  - on remotes: disconnect, if there is a default remote then "Unset as default"
  - open "Home", connect.
  - [ ] on "remotes" check that the `origin` is selected as default
  - edit the branch, change its upstream to another remote
    image
  - disconnect and "Unset the default" again
  - open "Home", connect
  - [ ] on "remotes" check that the `orther` is set as default
- [ ] prioritized remote is selected
  - switch to a branch without an upstream
    image
  - Have a remote called as "upstream"
  - disconnect and "Unset the default"
  - open "Home", connect
  - [ ] on "remotes" check that no "upstream" is set as default
  - [ ] same steps but no "upstream" remote: "origin" is set as default

# Checklist

### DIFF
--- a/src/git/models/repositoryShape.ts
+++ b/src/git/models/repositoryShape.ts
@@ -14,5 +14,6 @@ export interface RepositoryShape {
 		integration?: { id: SupportedCloudIntegrationIds; connected: boolean };
 		supportedFeatures: RemoteProviderSupportedFeatures;
 		url?: string;
+		bestRemoteName: string;
 	};
 }
--- a/src/webviews/apps/shared/components/repo-button-group.ts
+++ b/src/webviews/apps/shared/components/repo-button-group.ts
@@ -2,7 +2,7 @@ import { css, html, nothing } from 'lit';
 import { customElement, property } from 'lit/decorators.js';
 import { ifDefined } from 'lit/directives/if-defined.js';
 import { when } from 'lit/directives/when.js';
-import type { ConnectCloudIntegrationsCommandArgs } from '../../../../commands/cloudIntegrations';
+import type { ConnectRemoteProviderCommandArgs } from '../../../../commands/remoteProviders';
 import type { Source } from '../../../../constants.telemetry';
 import type { RepositoryShape } from '../../../../git/models/repositoryShape';
 import { createCommandLink } from '../../../../system/commands';
@@ -198,9 +198,9 @@ export class GlRepoButtonGroup extends GlElement {
 							return html`
 								<code-icon style="margin-top: -3px" icon="plug" aria-hidden="true"></code-icon>
 								<a
-									href=${createCommandLink<ConnectCloudIntegrationsCommandArgs>(
-										'gitlens.plus.cloudIntegrations.connect',
-										{ integrationIds: [provider.integration!.id], source: this.source },
+									href=${createCommandLink<ConnectRemoteProviderCommandArgs>(
+										'gitlens.connectRemoteProvider',
+										{ repoPath: repo.path, remote: provider.bestRemoteName },
 									)}
 								>
 									Connect to ${repo.provider!.name}
@@ -227,10 +227,10 @@ export class GlRepoButtonGroup extends GlElement {
 			<gl-button
 				part="connect-icon"
 				appearance="toolbar"
-				href=${createCommandLink<ConnectCloudIntegrationsCommandArgs>(
-					'gitlens.plus.cloudIntegrations.connect',
-					{ integrationIds: [provider.integration.id], source: this.source },
-				)}
+				href=${createCommandLink<ConnectRemoteProviderCommandArgs>('gitlens.connectRemoteProvider', {
+					repoPath: repo.path,
+					remote: provider.bestRemoteName,
+				})}
 			>
 				<code-icon icon="plug" style="color: var(--titlebar-fg)"></code-icon>
 				<span slot="tooltip">
--- a/src/git/utils/-webview/repository.utils.ts
+++ b/src/git/utils/-webview/repository.utils.ts
@@ -94,6 +94,7 @@ export async function toRepositoryShapeWithProvider(
 				: undefined,
 			supportedFeatures: remote.provider.supportedFeatures,
 			url: await remote.provider.url({ type: RemoteResourceType.Repo }),
+			bestRemoteName: remote.name,
 		};
 		if (provider.integration?.id == null) {
 			provider.integration = undefined;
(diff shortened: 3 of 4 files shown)


sergeibbb:
@axosoft-ramint What do we do with this? Reject, accept, transform to different requirements?

cc @eamodio

src/git/models/repositoryShape.ts
```diff
@@ -14,5 +14,7 @@ export interface RepositoryShape {
 		integration?: { id: SupportedCloudIntegrationIds; connected: boolean };
 		supportedFeatures: RemoteProviderSupportedFeatures;
 		url?: string;
+		remotePath?: string;
+		remoteName?: string;
```
ramin-t: I see `remotePath` being added here but I don't see where it is being used?

Maybe call it `bestRemoteName` since in the only two places we use this, it's the so-called "best remote" of the repo?

src/webviews/apps/shared/components/repo-button-group.ts
```diff
@@ -198,9 +198,9 @@ export class GlRepoButtonGroup extends GlElement {
 							return html`
 								<code-icon style="margin-top: -3px" icon="plug" aria-hidden="true"></code-icon>
 								<a
-									href=${createCommandLink<ConnectCloudIntegrationsCommandArgs>(
-										'gitlens.plus.cloudIntegrations.connect',
-										{ integrationIds: [provider.integration!.id], source: this.source },
+									href=${createCommandLink<ConnectRemoteProviderCommandArgs>(
+										'gitlens.connectRemoteProvider',
+										{ remote: provider.remoteName!, repoPath: repo.path },
```
ramin-t: Is this guaranteed to be the remote that the current branch is located on since this is where the link is located?

src/git/models/repositoryShape.ts
sergeibbb: Fixed

src/webviews/apps/shared/components/repo-button-group.ts
sergeibbb: @axosoft-ramint 

This is a good question. For picking the right integration the remote is unnecessary, because the only thing that matters is the integration type.

The problem here is setting the default remote when the integration is connected.

The old flow, that goes through the web site (ConnectCloudIntegrationsCommand), does not set the default remote when the integration is connected.
In the new flow the default remote is set to the value of the parameter that might be wrong.

So, I suggest to make the `remote` field optional and if it's not set, then do not set the default remote on connecting an integration. By doing that we:
1. keep the old behavior (not setting the remote at all)
2. won't fail with setting a wrong remote as default.

What do you think?

sergeibbb: @axosoft-ramint 

This is a good question. For picking the right integration the remote is unnecessary, because the only thing that matters is the integration type.

The problem here is setting the default remote when the integration is connected.

The old flow, that goes through the web site (ConnectCloudIntegrationsCommand), does not set the default remote when the integration is connected.
In the new flow the default remote is set to the value of the parameter that might be wrong.

So, I suggest to make the `remote` field optional and if it's not set, then do not set the default remote on connecting an integration. By doing that we:
1. keep the old behavior (not setting the remote at all)
2. won't fail with setting a wrong remote as default.

What do you think?

sergeibbb:
Hi @axosoft-ramint 

I've removed passing the remote. Now it's defined by some logic that already exists.
Also I've added "Verification steps" that describes how it selects the default origin and how it can be tested.

image

src/commands/remoteProviders.ts
```diff
@@ -13,7 +13,7 @@ import type { CommandContext } from './commandContext';
 import { isCommandContextViewNodeHasRemote } from './commandContext.utils';
 
 export interface ConnectRemoteProviderCommandArgs {
-	remote: string;
+	remote?: string;
```
ramin-t: It does feel a bit weird that we could call "ConnectRemoteProvider" as a command without a remote supplied, but that's probably just semantics at this point.

LGTM will verify once it's in, thanks for updating it!

axosoft-ramint (approved):

src/commands/remoteProviders.ts
sergeibbb: @axosoft-ramint 

> It does feel a bit weird that we could call "ConnectRemoteProvider" as a command without a remote supplied

Yes. It's called "connect provider" but connects an integration. **The question** is what if I have 2 remotes: one for GitHub and another one for Azure and both are disconnected. But at some point on the previous version with `ConnectCloudIntegrationsCommand` it should have been defined somehow too.

In `ConnectRemoteProviderCommand` it can be called without a defined `remote` and there is a code substitute a value for `remote`:

image

With old `ConnectCloudIntegrationsCommand` we already had a provider and took an integration-id from it:
image

But how that provider has been picked? I'm going to check out the origin of the `provider` value there and maybe it will be possible to take a remote for it in that place. Or, maybe the `ConnectRemoteProviderCommand` could accept `provider` argument when `remote` is unknown, because having `provider` is enough for getting its integration.

ramin-t:
This one disappeared from my feed/radar somehow, but I noticed today my review was requested again on this, so I assigned myself just to make it more visible to me

sergeibbb:
@axosoft-ramint I didn't requested the review recently. I kept it on me because I was review the optional `remote` case carefully and clarify whether it's right or wrong, but it has hanged on me.

src/commands/remoteProviders.ts
sergeibbb: @axosoft-ramint 
I've figured this out.

sergeibbb:
@axosoft-ramint 

I've figured this out: I do not need to make the `remote` argument optional anymore.

PR #4428, opened 2025-06-25
Adds support for mcp server installation

### LINKED ISSUE #4357: Install MCP server
### Description

- Download the mcp server from our release server into the vscode extension storage URI
  - We should be downloading the platform specific version
- Once downloaded, run `gk install` to install the mcp server onto the machine
- Once that completes run `gk mcp install`
  - See https://github.com/gitkraken/gkcli/pull/237

axosoft-ramint:
Closes #4357

### DIFF
--- a/contributions.json
+++ b/contributions.json
@@ -311,6 +311,10 @@
 			"icon": "$(sparkle)",
 			"commandPalette": "gitlens:enabled && !gitlens:untrusted && gitlens:gk:organization:ai:enabled"
 		},
+		"gitlens.ai.mcp.install": {
+			"label": "Install MCP Server",
+			"commandPalette": "gitlens:enabled && !gitlens:untrusted && gitlens:gk:organization:ai:enabled"
+		},
 		"gitlens.ai.rebaseOntoCommit:graph": {
 			"label": "AI Rebase Current Branch onto Commit (Preview)...",
 			"icon": "$(sparkle)",
--- a/contributions.json
+++ b/contributions.json
@@ -312,7 +312,7 @@
 			"commandPalette": "gitlens:enabled && !gitlens:untrusted && gitlens:gk:organization:ai:enabled"
 		},
 		"gitlens.ai.mcp.install": {
-			"label": "Install MCP Server",
+			"label": "Install GitKraken MCP Server",
 			"commandPalette": "gitlens:enabled && !gitlens:untrusted && gitlens:gk:organization:ai:enabled"
 		},
 		"gitlens.ai.rebaseOntoCommit:graph": {
--- a/docs/telemetry-events.md
+++ b/docs/telemetry-events.md
@@ -570,6 +570,40 @@ void
 }
 ```
 
+### cli/install/failed
+
+> Sent when a CLI install attempt fails
+
+```typescript
+{
+  'attempts': number,
+  'autoInstall': boolean,
+  'error.message': string
+}
+```
+
+### cli/install/started
+
+> Sent when a CLI install attempt is started
+
+```typescript
+{
+  'attempts': number,
+  'autoInstall': boolean
+}
+```
+
+### cli/install/succeeded
+
+> Sent when a CLI install attempt succeeds
+
+```typescript
+{
+  'attempts': number,
+  'autoInstall': boolean
+}
+```
+
 ### cloudIntegrations/connected
 
 > Sent when connected to one or more cloud-based integrations from gkdev
@@ -2724,6 +2758,35 @@ void
   'items.timings.enrichedItems': number,
   'items.timings.prs': number
 }
+```
+
+### mcp/setup/completed
+
+> Sent when GitKraken MCP setup is completed
+
+```typescript
+{
+  'requiresUserCompletion': boolean
+}
+```
+
+### mcp/setup/failed
+
+> Sent when GitKraken MCP setup fails
+
+```typescript
+{
+  'error.message': string,
+  'reason': string
+}
+```
+
+### mcp/setup/started
+
+> Sent when GitKraken MCP setup is started
+
+```typescript
+void
 ```
 
 ### openReviewMode
--- a/docs/telemetry-events.md
+++ b/docs/telemetry-events.md
@@ -578,7 +578,8 @@ void
 {
   'attempts': number,
   'autoInstall': boolean,
-  'error.message': string
+  'error.message': string,
+  'source': 'account' | 'subscription' | 'graph' | 'composer' | 'patchDetails' | 'settings' | 'timeline' | 'home' | 'ai' | 'ai:markdown-preview' | 'ai:markdown-editor' | 'ai:picker' | 'associateIssueWithBranch' | 'cloud-patches' | 'code-suggest' | 'commandPalette' | 'deeplink' | 'editor:hover' | 'feature-badge' | 'feature-gate' | 'inspect' | 'inspect-overview' | 'integrations' | 'launchpad' | 'launchpad-indicator' | 'launchpad-view' | 'merge-target' | 'notification' | 'prompt' | 'quick-wizard' | 'rebaseEditor' | 'remoteProvider' | 'scm' | 'scm-input' | 'startWork' | 'trial-indicator' | 'view' | 'walkthrough' | 'whatsnew' | 'worktrees'
 }
 ```
 
@@ -589,7 +590,8 @@ void
 ```typescript
 {
   'attempts': number,
-  'autoInstall': boolean
+  'autoInstall': boolean,
+  'source': 'account' | 'subscription' | 'graph' | 'composer' | 'patchDetails' | 'settings' | 'timeline' | 'home' | 'ai' | 'ai:markdown-preview' | 'ai:markdown-editor' | 'ai:picker' | 'associateIssueWithBranch' | 'cloud-patches' | 'code-suggest' | 'commandPalette' | 'deeplink' | 'editor:hover' | 'feature-badge' | 'feature-gate' | 'inspect' | 'inspect-overview' | 'integrations' | 'launchpad' | 'launchpad-indicator' | 'launchpad-view' | 'merge-target' | 'notification' | 'prompt' | 'quick-wizard' | 'rebaseEditor' | 'remoteProvider' | 'scm' | 'scm-input' | 'startWork' | 'trial-indicator' | 'view' | 'walkthrough' | 'whatsnew' | 'worktrees'
 }
 ```
 
@@ -600,7 +602,9 @@ void
 ```typescript
 {
   'attempts': number,
-  'autoInstall': boolean
+  'autoInstall': boolean,
+  'source': 'account' | 'subscription' | 'graph' | 'composer' | 'patchDetails' | 'settings' | 'timeline' | 'home' | 'ai' | 'ai:markdown-preview' | 'ai:markdown-editor' | 'ai:picker' | 'associateIssueWithBranch' | 'cloud-patches' | 'code-suggest' | 'commandPalette' | 'deeplink' | 'editor:hover' | 'feature-badge' | 'feature-gate' | 'inspect' | 'inspect-overview' | 'integrations' | 'launchpad' | 'launchpad-indicator' | 'launchpad-view' | 'merge-target' | 'notification' | 'prompt' | 'quick-wizard' | 'rebaseEditor' | 'remoteProvider' | 'scm' | 'scm-input' | 'startWork' | 'trial-indicator' | 'view' | 'walkthrough' | 'whatsnew' | 'worktrees',
+  'version': string
 }
 ```
 
@@ -829,7 +833,7 @@ or
 ```typescript
 {
   [`context.${string}`]: string | number | boolean,
-  'context.webview.host': 'editor' | 'view',
+  'context.webview.host': 'view' | 'editor',
   'context.webview.id': string,
   'context.webview.instanceId': string,
   'context.webview.type': string
@@ -852,7 +856,7 @@ or
   'context.repository.id': string,
   'context.repository.provider.id': string,
   'context.repository.scheme': string,
-  'context.webview.host': 'editor' | 'view',
+  'context.webview.host': 'view' | 'editor',
   'context.webview.id': string,
   'context.webview.instanceId': string,
   'context.webview.type': string,
@@ -871,7 +875,7 @@ or
   'context.pinned': boolean,
   'context.type': 'stash' | 'commit',
   'context.uncommitted': boolean,
-  'context.webview.host': 'editor' | 'view',
+  'context.webview.host': 'view' | 'editor',
   'context.webview.id': string,
   'context.webview.instanceId': string,
   'context.webview.type': string,
@@ -884,7 +888,7 @@ or
 
 ```typescript
 {
-  'context.webview.host': 'editor' | 'view',
+  'context.webview.host': 'view' | 'editor',
   'context.webview.id': string,
   'context.webview.instanceId': string,
   'context.webview.type': string,
@@ -917,7 +921,7 @@ or
   'context.repository.id': string,
   'context.repository.provider.id': string,
   'context.repository.scheme': string,
-  'context.webview.host': 'editor' | 'view',
+  'context.webview.host': 'view' | 'editor',
   'context.webview.id': string,
   'context.webview.instanceId': string,
   'context.webview.type': string,
@@ -944,7 +948,7 @@ or
   'context.pinned': boolean,
   'context.type': 'stash' | 'commit',
   'context.uncommitted': boolean,
-  'context.webview.host': 'editor' | 'view',
+  'context.webview.host': 'view' | 'editor',
   'context.webview.id': string,
   'context.webview.instanceId': string,
   'context.webview.type': string,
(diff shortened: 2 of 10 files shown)


ramin-t:
Draft until issues with the installer itself are addressed by the team that maintains it.

src/env/node/gk/cli/integration.ts
```diff
@@ -18,9 +29,18 @@ export class GkCliIntegrationProvider implements Disposable {
...
-		this._disposable = configuration.onDidChange(e => this.onConfigurationChanged(e));
+		this._disposable = Disposable.from(
+			configuration.onDidChange(e => this.onConfigurationChanged(e)),
+			this.container.subscription.onDidChange(this.onSubscriptionChanged, this),
+			...this.registerCommands(),
+		);
 
 		this.onConfigurationChanged();
+
+		const mcpInstallStatus = this.container.storage.get('ai:mcp:install');
```
eamodio: Nit, but this should really be `cli:install` and `cliInstallStatus`

src/constants.storage.ts
```diff
@@ -63,6 +63,8 @@ export type DeprecatedGlobalStorage = {
 };
 
 export type GlobalStorage = {
+	'ai:mcp:install': string;
```
eamodio: This should be a union type rather than just a string

contributions.json
```diff
@@ -311,6 +311,10 @@
 			"icon": "$(sparkle)",
 			"commandPalette": "gitlens:enabled && !gitlens:untrusted && gitlens:gk:organization:ai:enabled"
 		},
+		"gitlens.ai.mcp.install": {
+			"label": "Install MCP Server",
```
eamodio:
```suggestion
			"label": "Install GitKraken MCP Server",
```

eamodio (changes requested):
I feel like we are conflating the CLI install with MCP "setup" and they are different/distinct tasks that should be handled as such.

The auto-install is only of the CLI, not MCP, and we should attempt to "auto-install" the CLI until successful (or some number of successive failures)

src/env/node/gk/cli/integration.ts
```diff
@@ -56,4 +76,356 @@ export class GkCliIntegrationProvider implements Disposable {
...
+					break;
+				case 'Visual Studio Code - Insiders':
+					appName = 'vscode-insiders';
+					break;
+				case 'Cursor':
+					appName = 'cursor';
+					break;
+				case 'Windsurf':
+					appName = 'windsurf';
+					break;
```
d13: We reference in "VS Code - Exploration" (MS-provided adhoc build) and "VSCodium" (currently at 1.103.2) as well in other areas. Should we add these as well? 


```typescript
		case 'Visual Studio Code - Exploration':
			app = 'code-exploration';
			break;
		case 'VSCodium':
			app = 'codium';
			break;
```

ramin-t: Those aren't supported by the "mcp install" command for the time being and should hit the default case and fail for now.

If we want them supported we should ask the CLI team to add them.

ramin-t:
> I feel like we are conflating the CLI install with MCP "setup" and they are different/distinct tasks that should be handled as such.
> 
> The auto-install is only of the CLI, not MCP, and we should attempt to "auto-install" the CLI until successful (or some number of successive failures)

@eamodio This process is not installing a full-fledged GK CLI. It is installing something called "gk core", which is a lightweight version of the CLI designed to basically just support the MCP and a few other things. Should I still go ahead with calling all of this logic "install CLI"?

docs/telemetry-events.md
```diff
@@ -570,6 +570,40 @@ void
...
+```
+
+### cli/install/succeeded
+
+> Sent when a CLI install attempt succeeds
+
+```typescript
+{
+  'attempts': number,
+  'autoInstall': boolean
```
eamodio: Do we get a version number or anything after the install? Because that would be great to have.

docs/telemetry-events.md
```diff
@@ -2726,6 +2760,35 @@ void
...
+```
+
+### mcp/setup/started
+
+> Sent when GitKraken MCP setup is started
+
+```typescript
+void
+```
+
```
eamodio: We should make sure we have "source" plumbed in here

eamodio: Also the cli version would be great too

contributions.json
```diff
@@ -311,6 +311,10 @@
 			"icon": "$(sparkle)",
 			"commandPalette": "gitlens:enabled && !gitlens:untrusted && gitlens:gk:organization:ai:enabled"
 		},
+		"gitlens.ai.mcp.install": {
+			"label": "Install GitKraken MCP Server",
+			"commandPalette": "gitlens:enabled && !gitlens:untrusted && gitlens:gk:organization:ai:enabled"
+		},
```
eamodio, line 317: Eventually we will want versions of this command to other usages (e.g. home, etc) so we can understand the usage and make sure `source` is correct

src/env/node/gk/cli/integration.ts
eamodio: I checked in a new `getHostAppName` helper on `main` that we should use here. And IMO we should send whatever it is to the CLI and let them say its not supported, so that way the CLI can update and we dont need to for additional support.

src/env/node/gk/cli/integration.ts
```diff
@@ -56,4 +87,554 @@ export class GkCliIntegrationProvider implements Disposable {
...
+				} catch {
+					cliProxyFileExists = false;
+				}
+			}
+			if (cliInstall?.status !== 'completed' || cliPath == null || !cliProxyFileExists) {
+				try {
+					await window.withProgress(
+						{
+							location: ProgressLocation.Notification,
+							title: 'Setting up MCP integration...',
```
eamodio:
```suggestion
							title: 'Setting up the GitKraken MCP...',
```

src/env/node/gk/cli/integration.ts
```diff
@@ -56,4 +87,554 @@ export class GkCliIntegrationProvider implements Disposable {
...
+					);
+				} catch (ex) {
+					let failureReason = 'unknown error';
+					if (ex instanceof CLIInstallError) {
+						switch (ex.reason) {
+							case CLIInstallErrorReason.WebEnvironmentUnsupported:
+								void window.showErrorMessage(
+									'MCP installation is not supported in the web environment.',
+								);
+								failureReason = 'web environment unsupported';
```
eamodio: This is fine as a fallback, but we shouldn't allow the command on the Web and kick out at the start.

contributions.json
```diff
@@ -311,6 +311,10 @@
 			"icon": "$(sparkle)",
 			"commandPalette": "gitlens:enabled && !gitlens:untrusted && gitlens:gk:organization:ai:enabled"
 		},
+		"gitlens.ai.mcp.install": {
+			"label": "Install GitKraken MCP Server",
+			"commandPalette": "gitlens:enabled && !gitlens:untrusted && gitlens:gk:organization:ai:enabled"
```
eamodio, line 316: Should add `isWeb` (I think that is the context key) here to not show this on the web

src/env/node/gk/cli/integration.ts
```diff
@@ -56,4 +87,554 @@ export class GkCliIntegrationProvider implements Disposable {
...
+			if (cliPath != null) {
+				try {
+					await workspace.fs.stat(
+						Uri.joinPath(Uri.file(cliPath), getPlatform() === 'windows' ? 'gk.exe' : 'gk'),
+					);
+				} catch {
+					cliProxyFileExists = false;
+				}
+			}
+			if (cliInstall?.status !== 'completed' || cliPath == null || !cliProxyFileExists) {
```
eamodio: I'd suggest delegating all of this to `this.installCLI()` as this will need to check these same things. And have that method return the path/install data this method needs.

src/env/node/gk/cli/integration.ts
```diff
@@ -56,4 +87,554 @@ export class GkCliIntegrationProvider implements Disposable {
...
+						this.container.telemetry.sendEvent('mcp/setup/failed', {
+							reason: failureReason,
+							'error.message': ex instanceof Error ? ex.message : 'Unknown error during installation',
+						});
+					}
+				}
+			}
+
+			cliInstall = this.container.storage.get('gk:cli:install');
+			cliPath = this.container.storage.get('gk:cli:path');
```
eamodio: SHould be returned by `this.installCLI`

src/env/node/gk/cli/integration.ts
```diff
@@ -56,4 +87,554 @@ export class GkCliIntegrationProvider implements Disposable {
...
+						reason: 'unknown error',
+						'error.message': 'Unknown error during installation',
+					});
+				}
+				return;
+			}
+
+			if (appName !== 'cursor' && appName !== 'vscode' && appName !== 'vscode-insiders') {
+				const confirmation = await window.showInformationMessage(
+					`MCP configured successfully. Click 'Finish' to add it to your MCP server list and complete the installation.`,
```
eamodio:
```suggestion
					`MCP installed successfully. Click 'Finish' to add it to your MCP server list and complete the setup.`,
```

src/env/node/gk/cli/integration.ts
```diff
@@ -56,4 +87,554 @@ export class GkCliIntegrationProvider implements Disposable {
...
+						this.container.telemetry.sendEvent('mcp/setup/failed', {
+							reason: 'user cancelled',
+						});
+					}
+					return;
+				}
+			}
+
+			let output = await this.runCLICommand(
+				['mcp', 'install', appName, '--source=gitlens', `--scheme=${env.uriScheme}`],
```
eamodio, line 240: I don't know if `env.uriScheme` is enough when using remote hosts, we'd need to do some testing.

src/env/node/gk/cli/integration.ts
```diff
@@ -56,4 +88,577 @@ export class GkCliIntegrationProvider implements Disposable {
...
+				},
+			);
+
+			output = output.trim();
+			if (output === 'GitKraken MCP Server Successfully Installed!') {
+				return;
+			}
+
+			// Check if the output is a valid url. If so, run it
+			try {
```
eamodio, line 282:
```suggestion
				void window.showErrorMessage(`Failed to setup the GitKraken MCP: error getting install URL.`);
```

As for "error getting install URL" feels overly technical to send to the user, what can they do about it? But I don't have a better suggestion right now

src/env/node/gk/cli/integration.ts
```diff
@@ -56,4 +88,577 @@ export class GkCliIntegrationProvider implements Disposable {
...
+				new URL(output);
+			} catch {
+				if (this.container.telemetry.enabled) {
+					this.container.telemetry.sendEvent('mcp/setup/failed', {
+						reason: 'unexpected output from mcp install command',
+						'error.message': `Unexpected output from mcp install command: ${output}`,
+						source: commandSource,
+						'cli.version': cliVersion,
```
eamodio, line 290: This it just `true` right? Because we are only here if we need to use a URL?

src/env/node/gk/cli/integration.ts
```diff
@@ -56,4 +88,577 @@ export class GkCliIntegrationProvider implements Disposable {
...
+						this.container.telemetry.sendEvent('mcp/setup/failed', {
+							reason: 'user cancelled',
+							source: commandSource,
+							'cli.version': cliVersion,
+						});
+					}
+					return;
+				}
+			}
+
```
eamodio, line 238: We are missing the completion telemetry here.

src/env/node/gk/cli/integration.ts
```diff
@@ -56,4 +88,577 @@ export class GkCliIntegrationProvider implements Disposable {
...
+				void window.showErrorMessage(`Failed to install MCP integration: error getting install URL`);
+				return;
+			}
+
+			await openUrl(output);
+			if (this.container.telemetry.enabled) {
+				this.container.telemetry.sendEvent('mcp/setup/completed', {
+					requiresUserCompletion:
+						appName === 'cursor' || appName === 'vscode' || appName === 'vscode-insiders',
+					source: commandSource,
```
eamodio, line 302:
```suggestion
				`Failed to setup the GitKraken MCP: ${ex instanceof Error ? ex.message : String(ex)}`,
```

src/env/node/gk/cli/integration.ts
```diff
@@ -56,4 +88,577 @@ export class GkCliIntegrationProvider implements Disposable {
...
+					}
+
+					if (this.container.telemetry.enabled) {
+						this.container.telemetry.sendEvent('mcp/setup/failed', {
+							reason: failureReason,
+							'error.message': ex instanceof Error ? ex.message : 'Unknown error during installation',
+							source: commandSource,
+						});
+					}
+				}
```
eamodio, line 203:
```suggestion
				void window.showErrorMessage('Failed to install the GitKraken MCP: Unknown error.');
```
.

src/env/node/gk/cli/integration.ts
```diff
@@ -56,4 +88,577 @@ export class GkCliIntegrationProvider implements Disposable {
...
+								failureReason = 'unsupported platform';
+								break;
+							case CLIInstallErrorReason.ProxyUrlFetch:
+							case CLIInstallErrorReason.ProxyUrlFormat:
+							case CLIInstallErrorReason.ProxyFetch:
+							case CLIInstallErrorReason.ProxyDownload:
+							case CLIInstallErrorReason.ProxyExtract:
+							case CLIInstallErrorReason.CoreDirectory:
+							case CLIInstallErrorReason.CoreInstall:
+								void window.showErrorMessage('Failed to install MCP server locally.');
```
eamodio:
```suggestion
									`Failed to install the GitKraken MCP: ${ex instanceof Error ? ex.message : 'Unknown error}.`,
```

src/env/node/gk/cli/integration.ts
```diff
@@ -56,4 +88,577 @@ export class GkCliIntegrationProvider implements Disposable {
...
+								);
+								failureReason = 'web environment unsupported';
+								break;
+							case CLIInstallErrorReason.UnsupportedPlatform:
+								void window.showErrorMessage('MCP installation is not supported on this platform.');
+								failureReason = 'unsupported platform';
+								break;
+							case CLIInstallErrorReason.ProxyUrlFetch:
+							case CLIInstallErrorReason.ProxyUrlFormat:
+							case CLIInstallErrorReason.ProxyFetch:
```
eamodio:
```suggestion
								void window.showErrorMessage('Failed to install the GitKraken MCP.');
```

src/env/node/gk/cli/integration.ts
```diff
@@ -56,4 +88,577 @@ export class GkCliIntegrationProvider implements Disposable {
...
+							cliVersion = await this.installCLI();
+						},
+					);
+				} catch (ex) {
+					let failureReason = 'unknown error';
+					if (ex instanceof CLIInstallError) {
+						switch (ex.reason) {
+							case CLIInstallErrorReason.WebEnvironmentUnsupported:
+								void window.showErrorMessage(
+									'MCP installation is not supported in the web environment.',
```
eamodio:
```suggestion
								void window.showErrorMessage('GitKraken MCP installation is not supported on this platform.');
```

src/env/node/gk/cli/integration.ts
```diff
@@ -56,4 +88,577 @@ export class GkCliIntegrationProvider implements Disposable {
...
+							location: ProgressLocation.Notification,
+							title: 'Setting up MCP integration...',
+							cancellable: false,
+						},
+						async () => {
+							cliVersion = await this.installCLI();
+						},
+					);
+				} catch (ex) {
+					let failureReason = 'unknown error';
```
eamodio:
```suggestion
									'GitKraken MCP installation is not supported on the web.',
```

src/env/node/gk/cli/integration.ts
```diff
@@ -56,4 +88,577 @@ export class GkCliIntegrationProvider implements Disposable {
...
+					);
+				}
+
+				response = await fetch(downloadUrl);
+				if (!response.ok) {
+					throw new CLIInstallError(
+						CLIInstallErrorReason.ProxyFetch,
+						undefined,
+						`${response.status} ${response.statusText}`,
+					);
```
eamodio, line 481: We don't want to throw if this already exists right?

src/env/node/gk/cli/integration.ts
```diff
@@ -56,4 +88,577 @@ export class GkCliIntegrationProvider implements Disposable {
...
+				}
+
+				const cliProxyZipFileDownloadData = await response.arrayBuffer();
+				if (cliProxyZipFileDownloadData.byteLength === 0) {
+					throw new CLIInstallError(
+						CLIInstallErrorReason.ProxyDownload,
+						undefined,
+						'Downloaded proxy archive data is empty',
```
eamodio, line 489: What if the file already exists? We should overwrite it or nuke it first.

src/env/node/gk/cli/integration.ts
```diff
@@ -56,4 +88,577 @@ export class GkCliIntegrationProvider implements Disposable {
...
+
+				// Write the installer to the extension storage
+				try {
+					await workspace.fs.writeFile(cliProxyZipFilePath, new Uint8Array(cliProxyZipFileDownloadData));
+				} catch (ex) {
+					throw new CLIInstallError(
+						CLIInstallErrorReason.ProxyDownload,
+						ex instanceof Error ? ex : undefined,
+						'Failed to write proxy archive to global storage',
+					);
```
eamodio, line 515: Feels fragile -- maybe we should use an unzip lib?

src/env/node/gk/cli/integration.ts
ramin-t: @eamodio I am now using that helper. I still need to transform the app name from that helper because, for example, we use `code` and the command is expecting `vscode`, etc.

I also always continue to the command if the app name exists, and if they show an error in the output I send failure telemetry and kick out early. Check the latest commit.

src/env/node/gk/cli/integration.ts
eamodio: But we can deal with that later

eamodio: Will need to test this on linux

src/env/node/gk/cli/integration.ts
```diff
@@ -56,4 +89,586 @@ export class GkCliIntegrationProvider implements Disposable {
...
+					'error.message': ex instanceof Error ? ex.message : 'Unknown error',
+					source: source,
+				});
+			}
+			if (!autoInstall) {
+				throw ex;
+			}
+		}
+
+		return cliVersion;
```
eamodio: Why do we care about the status if the path exists?

eamodio (changes requested):

src/env/node/gk/cli/integration.ts
ramin-t: I can't understand your comment based on the code you're highlighting.

eamodio: GitHub is having serious issues as my comments keep moving their anchor. 

This was about the requiresUserConfirmation setter

src/env/node/gk/cli/integration.ts
ramin-t: I can't understand your suggestion here based on the line you're highlighting. was probably on an outdated version

src/env/node/gk/cli/integration.ts
ramin-t: I can't understand this comment based on the line it's attached to

contributions.json
ramin-t: It's not `isWeb`, or at least that's not working as intended because using `&& !isWeb` in the "when" clause prevents the command from appearing even when I'm not on web.

ramin-t:
@eamodio I tried addressing as many of your comments as I could, but most of them don't make sense because the code they are attached to got shifted. Would you mind resubmitting any that I did not address, with the latest on the branch?

d13:
Changes are included in https://github.com/gitkraken/vscode-gitlens/pull/4570

PR #4473, opened 2025-07-03
Adds new banner component

axosoft-ramint:


src/commands/resets.ts
```diff
@@ -71,6 +78,7 @@ export class ResetCommand extends GlCommandBase {
 				detail: 'Clears stored data associated with the current workspace',
 				item: 'workspace',
 			},
+			...(this.container.debugging ? [{ label: 'Promo Opt-Ins...', detail: 'Clears any locally stored promo opt-ins', item: 'promoOptIns' }]  satisfies ResetQuickPickItem[] : []),
```
eamodio: I'd suggest putting this into the `DEBUG` section below

ramin-t: Good catch, not sure how I missed that this section exists already. Thanks! Done in commit: https://github.com/gitkraken/vscode-gitlens/pull/4473/commits/6e1ec7eb745b05d529f60cd92a6b7f3c86b5033b

Issue #4449, opened 2025-06-26
Capture sentiment feedback about AI feature effectiveness


justinrobots:
### Description

- Allow users to submit a 👍 or 👎 rating on all AI features that generate a summary (explain, commit composer, etc.)
- Consider providing pre-defined follow-up responses like "Output was inaccurate", etc.
- Consider allowing users to add write-in responses

Send this data via OTel (see how GKD is doing this)

[Details in slack](https://gitkraken-hq.slack.com/archives/C06QFFUSVUZ/p1750958550816149)

ramin-t:
Cannot verify in today's pre-release if telemetry is being sent successfully or not, but can verify that the controls exist and toolbar ones go away when telemetry is off.

dzp5103:
N

PR #4479, opened 2025-07-08
Adds AI feedback support for changelog markdown editors

sergeibbb:
# Description
This is a follow-up of #4449

Enables the use of helpful/unhelpful AI feedback buttons in untitled markdown editors generated by the changelog command.

Tracks feedback context for these documents, updates relevant context keys.

# Checklist


sergeibbb:
Hi @eamodio 

[This PR](https://github.com/gitkraken/vscode-gitlens/pull/4479) is the implementation for Changelogs in a similar manner how it's been done for markdown previews.

Please let me know if you think that something needs to be improved here.

Also it has [a follow-up](https://github.com/gitkraken/vscode-gitlens/pull/4480) where I suggest to wrap static values to a disposable provider.

src/commands/generateChangelog.ts
```diff
@@ -98,12 +132,22 @@ export async function generateChangelogAndOpenMarkdownDocument(
 	if (result === 'cancelled') return;
 
 	const { range, changes: { length: count } = [] } = await changes.value;
+	const feedbackContext = result && getAIResultContext(result);
```
ramin-t, line 135: Is this always guaranteed to hit the case `uri?.scheme === Schemes.GitLensAIMarkdown`? Otherwise on a quick read it looks like it would result in a `getChangelogFeedbackContext` which wouldn't exist yet (I must be missing something).

sergeibbb: @axosoft-ramint 
It is rather guaranteed that changelog's scheme is not `GitLensAIMarkdown`. It's `"untitled"`.


- `getAIResultContext` is an util function creates an object assigned to `feedbackContext`.
- `const document = await workspace.openTextDocument` creates a document with `untitled` scheme.
- then `setChangelogFeedbackContext` stores the context.

ramin-t: Thanks

PR #4522, opened 2025-08-06
Adds AI powered rebase and explain feature of a branch

### LINKED ISSUE #4443: Add an option to explain commits from all changes on a branch
### Description

There should be an option to explain commits for all changes on a branch that is accessible from multiple places:
- in the context menu when right-clicking a branch ref in the graph
- as a command palette command
- from branches in other GitLens views

Add the command:
- `"Explain Unpushed Changed"`
- Modify `"Explain Branch Changes"` - by trying to find user-configured merge base.

image

### Follow-ups

Originally it was a ticket for "Recompose", but then "Explain" appeared from nowhere and eventually, "Recompose" has been moved out to wait for readiness of the CommitComposer. The follow up PR for recompose is here: https://github.com/gitkraken/vscode-gitlens/pull/4644 the original description of the current ticket has been moved there. Probably, an issue needs to be created as well.

### Verification steps

- [ ] test how each new menu item works
- [ ] test that new menu items are visible when and only when there are commits (unpushed or unmerged): both in graph and in Branch View.
- [ ] test how "Explain Branch Changes" works because it's base branch has been modified in `getBranchMergeTargetName `
- [ ] test how other features that depend on `getBranchMergeTargetName `
  - [ ] Get Branch Contributions Overview
  - [ ] Comparison default

sergeibbb:
# Description
Resolves #4443

image

# Checklist


src/constants.commands.ts
```diff
@@ -31,6 +31,8 @@ export type GlCommandsDeprecated =
 	| 'gitlens.showFileHistoryInView';
 
 type InternalGraphWebviewCommands =
+	| 'gitlens.ai.aiRebaseBranch:graph'
+	| 'gitlens.ai.aiRebaseUnpushed:graph'
```
ramin-t: Why are these listed under InternalGraphWebviewCommands? In contributions it looks like they are fully exposed graph commands?

src/constants.ts
```diff
@@ -47,6 +47,22 @@ export const enum CharCode {
...
+ * `gk-merge-target-user` — merge target branch explicitly defined by user,
+ *    if it's defined we use this value instead of `gk-merge-target`, but we keep storing `gk-merge-target` value that was determined automatically.
+ *
+ *  `gk-merge-base` means the branch that the current branch originates from, e.g. what was the base in the moment of creation.
+ *   This value is used for: ... (TODO describe use cases).
+ *
+ * `vscode-merge-base` — value determined by VS Code that is used to determine the merge base for the current branch.
+ *   once `gk-merge-base` is determined, we stop using `vscode-merge-base`
+ *
+ */
```
ramin-t, line 65: Thanks for explaining these here. These keys needed to be documented somewhere so I'm glad we have that here.

src/git/utils/-webview/branch.utils.ts
```diff
@@ -200,3 +201,75 @@ export function getStarredBranchIds(container: Container): Set<string> {
...
+	let mostRecentMergeBase = mergeBases[0];
+	for (let i = 1; i < mergeBases.length; i++) {
+		const isCurrentMoreRecent = await svc.commits.isAncestorOf(mostRecentMergeBase?.commit, mergeBases[i].commit);
+		if (isCurrentMoreRecent) {
+			mostRecentMergeBase = mergeBases[i];
+		}
+	}
+
+	return mostRecentMergeBase;
+}
```
ramin-t: Can I get a more detailed rundown of what this function is doing? Also, if it already exists in graph.ts, can we just have that import from here so we aren't duplicating logic?

sergeibbb: @axosoft-ramint 
I've added to the explanation and removed the obsolete sentence about the duplication:

```
/**
 * Selects the most recent merge base from multiple target branches.
 *
 * It gets the merge base for each target, then uses isAncestorOf() to find which one is newest.
 */
 ```

The story was that I've implemented originally in graph, then created the dupe here and marked it in the comment. But eventually I refactored to re-use this function everywhere so it's stopped being duplicated, but the explanation phrase has left.

File: src/constants.commands.ts (hunk 1 above)
Comment (sergeibbb) [reply src/constants.commands.ts]: fixed

File: src/git/utils/-webview/branch.utils.ts (hunk 3 above)
Comment (sergeibbb) [reply src/git/utils/-webview/branch.utils.ts]: @axosoft-ramint 

I've updated the code according to your notes above. Also I've refreshed the branch and fixed the conflicts.

The question now is the following: should I actually call "Rebase" action, or is it better to try to launch the Commit Composer?

cc @justinrobots

**Comment by sergeibbb** [comment]
@axosoft-ramint , @ericf-axosoft 

I've moved "recompose" related part of requirements and the code to another PR draft: #4644

All other review notes has been attended.

cc @justinrobots

File: src/env/node/git/sub-providers/graph.ts [hunk 4]
```diff
@@ -347,6 +347,13 @@ export class GraphGitSubProvider implements GitGraphSubProvider {
 
 							branch = branchMap.get(tip);
 							branchId = branch?.id ?? getBranchId(repoPath, false, tip);
+
+							// Check if branch has commits that can be recomposed and get merge base
+							const mergeBaseResult =
+								branch && (await getBranchMergeBaseAndCommonCommit(this.container, branch));
+							const isRecomposable = Boolean(mergeBaseResult && mergeBaseResult.commit !== branch?.sha);
+							const mergeBase = isRecomposable ? mergeBaseResult : undefined;
```
ramin-t: It seems these calculations are expensive in some cases, especially when placed in the main graph loop.

Without these operations (commented out and left out of context), I can load the graph for my test repo (not a very large one) in about 5 seconds:

Image

Same repo, same graph, with these operations, takes nearly 2 minutes to load:

Image

axosoft-ramint (changes requested):
The context calculations are tanking graph performance. We need to find a lighter way to get that context included.

src/env/node/git/sub-providers/graph.ts
ramin-t: Let me know if you need any details about my repo/branches/commits that would help with reproducing this.

sergeibbb: @axosoft-ramint  I've moved more changes out to rebase related branch. So, the problem can be handled in a future PR. Now this PR only contains Explain related changes that are not bring new calculations to rendering.

PR #4565, opened 2025-09-03
Unifies autolink enrichment keys for integrations

sergeibbb:
# Description

Unifies autolink enrichment keys for integrations in order to avoid referencing the exact integration type outside of integratin class.

Updates the autolink enrichment workflow to consistently use a pair of values (id and key) when resolving issues or pull requests across integrations. This removes provider-specific logic for constructing enrichment keys and centralizes the approach, simplifying maintenance and reducing errors when linking to external systems.

Renames and refactors relevant methods in integration classes to reflect this unified signature. Enhances consistency and clarity in how autolinked items are identified and retrieved across supported providers.

# Checklist

### DIFF
--- a/src/plus/integrations/models/integration.ts
+++ b/src/plus/integrations/models/integration.ts
@@ -499,9 +499,9 @@ export abstract class IntegrationBase<
 	): Promise<IssueShape[] | undefined>;
 
 	@debug()
-	async getIssueOrPullRequest(
+	async getLinkedIssueOrPullRequest(
 		resource: T,
-		id: string,
+		link: { id: string; key: string },
 		options?: { expiryOverride?: boolean | number; type?: IssueOrPullRequestType },
 	): Promise<IssueOrPullRequest | undefined> {
 		const scope = getLogScope();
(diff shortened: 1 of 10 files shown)


sergeibbb:
@axosoft-ramint @eamodio 

I'm making this change because Linear is another integration that needs a full issue key rather than just `<num>`, that would require me to mention `Linear` integration in the `autolinksProvider.ts` as well as `Jira` was mentioned there. By this improvement we remove switching between from the `autlinksProvider` and move the decision down to the integrations because they know better what they want to use to retrieve data for the link.

src/plus/integrations/models/integration.ts
```diff
@@ -512,17 +512,17 @@ export abstract class IntegrationBase<
 		await this.refreshSessionIfExpired(scope);
 
 		const issueOrPR = this.container.cache.getIssueOrPullRequest(
```
ramin-t, line 514: Should we update the cache function's name and parameter as well since this has been updated everywhere else?

sergeibbb: @axosoft-ramint 
I'm not sure about that. Integration provider is smart while cache is stupid.

Integration provider is smart enough to understand why it have to pick among two incoming strings (id and key). So, when a developer creates, e.g. Gerrit integration it knows that it's a `Gerrit. getLinkedIssueOrPullRequest()` and picks a better value for retrieving the linked object by API.

While cache is just a key-value storage that stores issues or PR: Hey cache, take a key and give me whatever you have.

sergeibbb: @axosoft-ramint 
well, now I'm thinking that maybe it's worth to add _type_ (`type?: IssueOrPullRequestType`) to the cache key, because it plays role in the retrieving algorithm (in the provider-specific part of the method).

what do you think?

PR #4570, opened 2025-09-05
Register MCP using the extension

d13:


src/env/node/gk/mcp/integration.ts
```diff
@@ -0,0 +1,132 @@
...
+		const appName = toMcpInstallProvider(await getHostAppName());
+		if (appName == null) {
+			return undefined;
+		}
+
+		let output = await runCLICommand(['mcp', 'config', appName, '--source=gitlens', `--scheme=${env.uriScheme}`], {
+			cwd: cliPath,
+		});
+		output = output.replace(CLIProxyMCPConfigOutputs.checkingForUpdates, '').trim();
+		console.log(output);
```
ramin-t: Looks like maybe a leftover console log?

src/env/node/gk/mcp/integration.ts
```diff
@@ -0,0 +1,132 @@
...
+		this.checkStorage();
+	}
+
+	private checkStorage(e?: StorageChangeEvent): void {
+		if (e != null && !(e.keys as string[]).includes('gk:cli:install')) return;
+		this._onDidChangeMcpServerDefinitions.fire();
+	}
+
+	private async provideMcpServerDefinitions(): Promise<McpServerDefinition[]> {
+		const config = await this.getMcpConfigurationFromCLI();
```
ramin-t, line 34: This looks like it depends on the CLI/proxy to have been installed first, which currently only happens if the user runs the "install gitkraken mcp server" command since the auto-install part has been commented out. Do we want to uncomment it for this PR since it feels like a dependency (cc @eamodio )? And are there timing concerns between that happening on the CLI provider's constructor and this?

src/env/node/gk/mcp/integration.ts
```diff
@@ -0,0 +1,132 @@
...
+	// 	const appName = toMcpInstallProvider(await getHostAppName());
+	// 	const args = ['mcp', `--host=${appName}`, '--source=gitlens', `--scheme=${env.uriScheme}`];
+	// 	return {
+	// 		name: 'GitKraken MCP Server',
+	// 		type: 'stdio',
+	// 		command: command.fsPath,
+	// 		args: args,
+	// 		version: cliInstall.version,
+	// 	};
+	// }
```
ramin-t: Looks like leftover code - do you want it left in?

src/env/node/gk/mcp/integration.ts
d13: We've re-enabled the auto-install. For now, while in pre-release, auto-install will be behind an `gitlens.gitkraken.cli.autoInstall.enabled` setting which will require a window reload.

axosoft-ramint (approved):
LGTM

PR #4582, opened 2025-09-11
Stores hunks and safety state on webview host and uses for composer operations

ramin-t:
This is to prevent entire diffs being passed back from the app via IPC for commands like auto-compose, generate commit message, finish and commit, etc.

One thing I noticed is that the diff contents have artifacts from serialization and this can break the safety checks and the diff hashes, causing false positive errors on safety state. Also, it seems expensive to pass massive payloads like this over IPC for these ops anyway.

We still need to pass the hunks with their diffs through to the app for rendering content, but that should be the only purpose of the hunk contents in app state.

Lots of regression risk here i.e. with undo/reset history, reload, all AI ops, finish & commit, etc. so lots of testing required.

### DIFF
--- a/src/plus/integrations/integrationService.ts
+++ b/src/plus/integrations/integrationService.ts
@@ -174,6 +174,9 @@ export class IntegrationService implements Disposable {
 			if (cloudIntegrationTypes.length > 0) {
 				query += `&provider=${cloudIntegrationTypes.join(',')}`;
 			}
+			if (cloudIntegrationTypes.length > 1) {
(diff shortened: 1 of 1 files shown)


ramin-t:
@d13 @eamodio This is ready for review - I'm marking as draft just so we don't accidentally merge it with the current release.

It should be in a follow-up patch because there's high regression risk and it needs extensive testing.

ramin-t:
Wrong branch targets. Reopening as a different PR

PR #4609, opened 2025-09-23
Fixes some AI cancellation cases being treated as errors

ramin-t:
Repro steps:

1. Open commit composer
2. Try to generate commit message with no provider selected and github copilot disabled as an extension (so no vscode ai provider available) and not logged into GK (so no GK provider available as fallback)
3. Cancel or escape at any point in the flow (choosing a provider, choosing a model, sign in prompt for GK provider)

Expect for composer to treat it as a cancellation. See an error instead.

Note: these changes affect logic being used in many places because they touch core AI provider service paths, so lots of possible regressions to test:
1. AI provider selection in other places should work as before
2. AI commit message generation in other places as well as other AI ops should be spot-checked

### DIFF
--- a/src/plus/ai/aiProviderService.ts
+++ b/src/plus/ai/aiProviderService.ts
@@ -25,6 +25,7 @@ import {
 	AINoRequestDataError,
 	AuthenticationRequiredError,
 	CancellationError,
+	isCancellationError,
 } from '../../errors';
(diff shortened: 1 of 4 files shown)


d13 (approved):
LGTM

eamodio (approved):

PR #4653, opened 2025-10-03
Adds support for untracked files in composer

### LINKED ISSUE #4636: Add support for untracked files being included in Commit Composer
When you add a file, it is untracked by default. But this is quite common, and people should be able to commit these with composer when using "include unstaged changes".

We should add support for including untracked changes in the diffs and including them in commits when that option is on (or when no staged changes exist).

ramin-t:
Closes #4636

### DIFF
--- a/src/webviews/plus/composer/composerWebview.ts
+++ b/src/webviews/plus/composer/composerWebview.ts
@@ -322,9 +322,7 @@ export class ComposerWebviewProvider implements WebviewProvider<State, State, Co
 		]);
 
 		if (untrackedPaths?.length) {
-			try {
-				await repo.git.staging?.unstageFiles(untrackedPaths);
-			} catch {}
+			await repo.git.staging?.unstageFiles(untrackedPaths).catch();
 		}
 
 		const diffs = getSettledValue(diffsResult)!;


src/webviews/plus/composer/composerWebview.ts
```diff
@@ -299,13 +303,30 @@ export class ComposerWebviewProvider implements WebviewProvider<State, State, Co
...
 			// Handle baseCommit - could be string (old format) or ComposerBaseCommit (new format)
 			getWorkingTreeDiffs(repo),
 			repo.git.commits.getCommit('HEAD'),
 			repo.git.branches.getBranch(),
 		]);
 
+		if (untrackedPaths?.length) {
+			try {
+				await repo.git.staging?.unstageFiles(untrackedPaths);
+			} catch {}
```
d13: Nit: I think this should work the same

```suggestion
			await repo.git.staging?.unstageFiles(untrackedPaths).catch();
```

ramin-t: Is it the try/catch enclosure we want to remove? I just don't want the temporary staging and unstaging of the untracked files to throw anything that would break the flow since it is only being done to get them included in the diffs.

ramin-t: Oh I see what you did - never mind, good point, thanks!

ramin-t: Done in commit: [4e04a8b](https://github.com/gitkraken/vscode-gitlens/pull/4653/commits/4e04a8b183b7774db966429e7ad4f3b37ecb2eab)

PR #4656, opened 2025-10-03
Adds support for composing without a base commit

### LINKED ISSUE #4637: Add support for using commit composer without a base commit
### Description

We've seen a lot of failures where people are intending to use the commit composer with no base commit. This can occur if for example, you use it on a fresh repo with only changes in the workdir (no commits yet, no branches yet).

We should support this use case since it's common enough, especially for people who want to try out the composer in a "sandbox" repo.

ramin-t:
Closes #4637

Strategy here is to create an empty base commit without touching the working tree by using temporary index and commit-tree, then pointing HEAD to it, and then there is a commit that we can stash on temporarily. We then proceed to create the composer commits as patches and commit them with a similar strategy, except they are on another temporary index off the root sha. Then when we reset, our stash and empty commit are orphaned but still accessible. We pop the stash as normal and then let garbage collection eliminate the temporary empty commit we created.

### DIFF
--- a/src/env/node/git/git.ts
+++ b/src/env/node/git/git.ts
@@ -34,6 +34,7 @@ import {
 } from '../../../git/errors';
 import type { GitDir } from '../../../git/gitProvider';
 import type { GitDiffFilter } from '../../../git/models/diff';
+import { rootSha } from '../../../git/models/revision';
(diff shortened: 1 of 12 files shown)


d13 (approved):
LGTM

PR #4687, opened 2025-10-14
Improves MCP checks and adds offline check

### LINKED ISSUE #4672: Investigate isWeb check on MCP install flow and make sure it works as intented
There was at least one case where the MCP banner showed up in a web environment where it should not have - so we need to make sure isWeb is properly flagging all web environments correctly.

### LINKED ISSUE #4673: Kick out earlier on MCP install/configuration flows when offline
Currently, if you are offline, the fetch calls fail during CLI install and it is treated as an error, and uses up your auto-install attempts.

Being offline should kick the process out earlier and not use your auto-install attempts, so that you can auto-install when back online.

### LINKED ISSUE #4674: Reset MCP auto-install attempts on the next version
We will introduce a fix in a patch for unzipping during the CLI install part of the MCP flow. Folks who previously ran into issues with the unzip part may have used all their auto-install attempts already, and would not benefit from this fix. So we should do a one-time reset of auto-install attempts for anyone who has not successfully installed the CLI when this change goes live.

Related issue: https://github.com/gitkraken/vscode-gitlens/issues/4666

ramin-t:
Closes #4672
Closes #4673
Closes #4674

- Adds an isOffline check which does a one-time check on extension load and stores it for use. Uses this check to kick out of several MCP/CLI flows early
- Adds an isWeb check to the upgrade notification for MCP that prevents showing it in web environments
- Clears MCP autoinstall attempts when upgrading to allow folks who previously ran into the unzip issue to use the new unzip in the install process

### DIFF
--- a/src/container.ts
+++ b/src/container.ts
@@ -190,6 +190,7 @@ export class Container {
 		this._context = context;
 		this._prerelease = prerelease;
 		this._version = version;
+		this._previousVersion = previousVersion;
 		this.ensureModeApplied();
 
 		this._disposables = [
@@ -745,6 +746,11 @@ export class Container {
 	private readonly _version: string;
 	get version(): string {
 		return this._version;
+	}
+
+	private readonly _previousVersion: string | undefined;
+	get previousVersion(): string | undefined {
+		return this._previousVersion;
 	}
 
 	private readonly _views: Views;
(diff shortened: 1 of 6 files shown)


d13 (approved):
LGTM!

PR #4713, opened 2025-10-21
Adds initial support for recomposing commits from a branch

### LINKED ISSUE #4598: Recompose an entire branch in Composer
### Description

Allow users to recompose all commits on a branch in Composer.

- [x] Add paths to recompose all commits on a branch wherever we surface branches
  - [x] branch refs in the graph
  - [x] GL branches view
  - [x] Command palette option for recomposing a branch
- [x] When recomposing a branch, open Commit Composer with existing commits listed
- [x]  #4599

### LINKED ISSUE #4599: Allow existing commits to be auto-recomposed with AI
When recomposing existing commits in Composer, allow users to auto-recompose all of the commits.

ramin-t:
Closes #4598
Closes #4599

New command is `Recompose Commits (Preview)` and is accessible from branches in views and graph, and from Command Palette.

### DIFF
--- a/src/webviews/plus/composer/composerWebview.ts
+++ b/src/webviews/plus/composer/composerWebview.ts
@@ -504,74 +504,94 @@ export class ComposerWebviewProvider implements WebviewProvider<State, State, Co
 			};
 		}
 
-		// Get the merge target for the branch
-		const baseBranchNameResult = await getBranchMergeTargetName(this.container, branch);
+		// Get the merge target for the branch with recursive resolution
 		let mergeTargetName: string | undefined;
-		if (!baseBranchNameResult.paused) {
-			mergeTargetName = baseBranchNameResult.value;
+		let currentMergeTargetBranchName = branchName;
+		let currentMergeTargetBranch = branch;
+		const visitedBranches = new Set<string>();
+		let attempts = 0;
+		const maxAttempts = 10;
+
+		while (attempts < maxAttempts) {
+			attempts++;
+
+			// Prevent infinite loops by tracking visited branches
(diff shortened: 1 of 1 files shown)


ramin-t:
Note: we have a bug with storing merge bases for branches at the moment (we are choosing the upstream rather than the upstream's merge base) which causes the composer to find no diff when you make a local branch from a remote one. I will address this separately and rebase this on top of it once it is addressed.

ramin-t:
> Note: we have a bug with storing merge bases for branches at the moment (we are choosing the upstream rather than the upstream's merge base) which causes the composer to find no diff when you make a local branch from a remote one. I will address this separately and rebase this on top of it once it is addressed.

It was not a bug. It looked for the upstream only when the local wasn't available. But that logic wasn't working well in several cases (it would return zero commit responses), so I just used recursive merge target look-up until we find one with commits.

src/webviews/plus/composer/utils/composer.utils.ts
```diff
@@ -393,5 +574,178 @@ export function validateResultingDiff(
...
+	let match;
+	while ((match = coAuthorRegex.exec(commit.message)) !== null) {
+		const [, name, email] = match;
+		if (name) {
+			coAuthors.push({ name: name.trim(), email: email?.trim(), date: commit.date });
+		}
+	}
+
+	return coAuthors;
+}
```
d13, line 660: This will be great to add in other commit authorship/avatar content throughout GitLens.

ramin-t: Will make a note to pull it into a shared util file later. Thanks!

PR #4746, opened 2025-11-12
Uninstalls duplicate MCP

### LINKED ISSUE #4626: Uninstall duplicate GK MCP servers
### Description

In cases where GitLens is unable to bundle the GitKraken MCP (GK MCP) (usually when an IDE does not support VS Code's bundled MCP API) a more manual setup process is required for the GK MCP to function. Since the bundled and manually setup GK MCP servers are different to an IDE, this can result in duplicate GK MCP servers being setup at once if an IDE starts supporting VS Code's bundled MCP API.

This issue is to explore cleaning up the manually setup versions of the GK MCP server when the bundled version also exists.

### GK CLI version
UPD. 2025-11-21: CLI is fixed on core@3.1.48, installer@3.1.48.

### Steps to reproduce

- Run VSCode
- Go to the folder where GK installer is installed
- Run `./gk mcp install vscode-insiders` or `./gk mcp install vscode`
    - If you launch VSCode from a custom profile folder then run `./gk mcp install vscode-insiders --source=gitlens`. Then click "install" button on the "MCP Server: GitKraken" tab in VS Code.
    - If you launch in standard folder in debug sub-profile, make sure you use main `mcp.json` to list MCP servers:
        Image

- Install GitLens pre-release.
- Using VSCode palette run the command "MCP: List Servers"

#### Expected result
- previously installed (independent) GitKraken MCP is not in the list
- you see "GitKraken (bundled with GitLens)" MCP server _only._

#### Actual result
you see both MCP servers:
- "GitKraken (bundled with GitLens)"
- "GitKraken" - global in VSCode

sergeibbb:
# Description
Fixes #4626

# Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,8 @@ gitlens-*.vsix
 product.json
 tsconfig*.tsbuildinfo
 localdev*.instructions.md
+*localdev.instructions.md
+localdev*.prompt.md
+*localdev.prompt.md
+localdev*.chatmode.md
+*localdev.chatmode.md
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -73,6 +73,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 
 ### Fixed
 
+- Fixes duplication of GitKraken MCP [#4626](https://github.com/gitkraken/vscode-gitlens/issues/4626)
 - Fixes an issue where the _Home_ view would not update when switching repositories ([#4717](https://github.com/gitkraken/vscode-gitlens/issues/4717))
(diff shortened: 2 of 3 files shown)


sergeibbb:
@eamodio I have one doubt here.

I suppose, the result of the fix should give us the following:

- User comes to us and says "I have duplication of MCP"
- We say: "Try one of the following" (both work):
1. Reload your IDE window
2. Call `GitLens: Reinstall GitKraken MCP` in the command palette.

However, now we might have problem with the option #2.

That's because, I call deduping after the successful config call, here:
image

But it's never called twice until the version of the GK CLI is updated:
image
image


Therefore, the question is the following:
1. Is it OK that the second recipe won't always work (only once per launch)
2. Or is it better to move deduping command to another place? e.g. from `onRegistrationCompleted` to `getMcpConfigurationFromCLI` ?

cc @axosoft-ramint

PR #4759, opened 2025-11-25
Improves commit composer UX

### LINKED ISSUE #4540: Commit Composer UX Improvements
### Description

Minor refinements to make in Commit Composer

- [x] Auto-compose summary content should have a border around it
- [x] Commit message and description should have different font sizes and colors (description should be dimmer, and perhaps smaller)
- [x] Need a “maximize” button for the entire view
- [x] Should we add “generate message” and “regenerate message” text to the button? It’s just a ✨ right now.
- [x] The custom instructions font size should probably be a little larger, especially since the color contrast is not as high on the grey background.
- [x] When resetting changes, we should re-select the first draft commit
- [x] Base commit shouldn’t get a pointer cursor when hovered since there is no interaction
- [x] Commit message input doesn’t lose blue focus border when focus is blurred (clicking outside the input, etc.)
- [x] make Commit message and File Header sticky when scrolling diffs (stick to the top of the view)

ramin-t:
Closes #4540

Note: most of the "viral" changes (that span across many files) are to support the "maximize" function of the composer tab, or to refactor the message of a composer commit to an object that includes whether it was AI generated or not (so we can label the button Generate/Regenerate Commit Message correctly in the message box).

### DIFF
--- a/src/webviews/apps/plus/composer/components/app.ts
+++ b/src/webviews/apps/plus/composer/components/app.ts
@@ -58,6 +58,7 @@ interface ComposerDataSnapshot {
 	selectedHunkIds: Set<string>;
 	hasUsedAutoCompose: boolean;
 	recompose: { enabled: boolean; branchName?: string; locked: boolean; commitShas?: string[] } | null;
+	compositionSummarySelected: boolean;
 }
 
 interface ComposerHistory {
@@ -620,6 +621,7 @@ export class ComposerApp extends LitElement {
 			selectedHunkIds: new Set([...this.selectedHunkIds]),
 			hasUsedAutoCompose: this.state?.hasUsedAutoCompose ?? false,
 			recompose: this.state?.recompose ? JSON.parse(JSON.stringify(this.state.recompose)) : null,
+			compositionSummarySelected: this.compositionSummarySelected,
 		};
 	}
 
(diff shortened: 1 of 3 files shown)


ramin-t:
Secondary note: the commits for this PR were generated by Commit Composer

src/webviews/apps/plus/composer/components/commit-message.ts
```diff
@@ -55,20 +66,32 @@ export class CommitMessage extends LitElement {
...
 			}
 
 			.commit-message__input {
 				box-sizing: content-box;
-				width: calc(100% - 1rem);
-				border: 1px solid var(--vscode-input-border);
-				background: var(--vscode-input-background);
+				width: calc(100% - 2rem);
+				border: 1px solid var(--vscode-input-border, #858585);
+				background: var(--vscode-input-background, #3c3c3c);
```
d13, line 94: This is going to apply the `#858585` most of time in dark themes because most dark themes only use the border color for the high-contrast versions.

The opposite for `background`, high-contrast themes typically remove the background color, so the `#3c3c3c` will add one back, reducing contrast.

ramin-t: Thanks! What would you suggest here?

d13: After chatting, this is good to go

PR #4875, opened 2026-01-13
Adds core infrastructure for Git commit signing

eamodio:
Implements the implementation plan for Phase 1 of Git commit signing support, establishing the foundational architecture for cryptographic signing within the extension.

- **Core Infrastructure**: Adds the ability to read and write signing configurations directly from Git global/local config, supporting GPG, SSH, and X.509 formats.
- **Signature Parsing**: Updates the commits provider to lazily fetch and parse signatures from git logs, verifying status (good, bad, expired, etc.) and ownership.
- **Commit Creation**: Enhances the patch provider to support signing when creating commits and unreachable commits, respecting the user's auto-sign configuration.
- **User Interface**: Introduces a new status bar controller to display the current signing mode and a setup command to help users configure signing keys.
- **Integration**: Hooks signing capabilities into the Composer feature, allowing generated commits to be signed.
- **Telemetry & Error Handling**: Implements specific error handling for common signing issues (missing keys, passphrase failures) and adds telemetry events to track adoption and stability.

### DIFF
--- a/src/git/gitProvider.ts
+++ b/src/git/gitProvider.ts
@@ -391,8 +391,8 @@ export interface GitCommitsSubProvider {
 		rev: string,
 		cancellation?: CancellationToken,
 	): Promise<GitCommitReachability | undefined>;
-
 	getCommitSignature?(repoPath: string, sha: string): Promise<CommitSignature | undefined>;
+	isCommitSigned?(repoPath: string, sha: string): Promise<boolean>;
 }
 
 export interface GitOperationsSubProvider {
--- a/src/env/node/git/sub-providers/commits.ts
+++ b/src/env/node/git/sub-providers/commits.ts
@@ -1309,7 +1309,12 @@ export class CommitsGitSubProvider implements GitCommitsSubProvider {
 
 		try {
 			const result = await this.git.exec(
-				{ cwd: repoPath, errors: 'ignore', configs: gitConfigsLog },
+				{
+					cwd: repoPath,
+					errors: 'ignore',
+					caching: { cache: this.cache.gitResults, options: { accessTTL: 60 * 1000 } },
+					configs: gitConfigsLog,
+				},
 				'log',
 				`--format=${signatureFormat}`,
 				'-1',
@@ -1322,6 +1327,28 @@ export class CommitsGitSubProvider implements GitCommitsSubProvider {
 		} catch (ex) {
 			Logger.error(ex, scope);
 			return undefined;
+		}
+	}
+
+	@log()
+	async isCommitSigned(repoPath: string, sha: string): Promise<boolean> {
+		const scope = getLogScope();
+
+		try {
+			const result = await this.git.exec(
+				{
+					cwd: repoPath,
+					errors: 'ignore',
+					caching: { cache: this.cache.gitResults, options: { accessTTL: 60 * 1000 } },
+				},
+				'cat-file',
+				'commit',
+				sha,
+			);
+			return /^gpgsig(-sha256)? /m.test(result.stdout);
+		} catch (ex) {
+			Logger.error(ex, scope);
+			return false;
 		}
 	}
 }
--- a/src/git/models/commit.ts
+++ b/src/git/models/commit.ts
@@ -59,6 +59,7 @@ export class GitCommit implements GitRevisionReference {
 	private _stashUntrackedFilesLoaded = false;
 	private _recomputeStats = false;
 	private _signature: CommitSignature | undefined | null;
+	private _signed: boolean | undefined;
 
 	readonly lines: GitCommitLine[];
 	readonly ref: string;
@@ -599,6 +600,15 @@ export class GitCommit implements GitRevisionReference {
 			null;
 
 		return this._signature ?? undefined;
+	}
+
+	async isSigned(): Promise<boolean> {
+		if (this.isUncommitted) return false;
+		if (this._signed != null) return this._signed;
+
+		this._signed =
+			(await this.container.git.getRepositoryService(this.repoPath).commits.isCommitSigned?.(this.sha)) ?? false;
+		return this._signed;
 	}
 
 	async getCommitForFile(file: string | GitFile, staged?: boolean): Promise<GitCommit | undefined> {
--- a/src/hovers/hovers.ts
+++ b/src/hovers/hovers.ts
@@ -251,7 +251,7 @@ export async function detailsMessage(
 		presenceResult,
 		previousLineComparisonUrisResult,
 		_fullDetailsResult,
-		signatureResult,
+		signedResult,
 	] = await Promise.allSettled([
 		enhancedAutolinks
 			? pauseOnCancelOrTimeoutMapTuplePromise(
@@ -278,9 +278,7 @@ export async function detailsMessage(
 			? commit.getPreviousComparisonUrisForRange(editorLineToDiffRange(editorLine), uri.sha)
 			: undefined,
 		commit.message == null ? commit.ensureFullDetails() : undefined,
-		showSignature
-			? pauseOnCancelOrTimeout(commit.getSignature(), options?.cancellation, options?.timeout)
-			: undefined,
+		showSignature ? commit.isSigned() : undefined,
 	]);
 
 	if (options?.cancellation?.isCancellationRequested) return undefined;
@@ -289,7 +287,7 @@ export async function detailsMessage(
 	const pr = getSettledValue(prResult);
 	const presence = getSettledValue(presenceResult);
 	const previousLineComparisonUris = getSettledValue(previousLineComparisonUrisResult);
-	const signature = getSettledValue(signatureResult);
+	const signed = getSettledValue(signedResult);
 
 	const details = await CommitFormatter.fromTemplateAsync(
 		options.format,
@@ -308,7 +306,7 @@ export async function detailsMessage(
 			previousLineComparisonUris: previousLineComparisonUris,
 			outputFormat: 'markdown',
 			remotes: remotes,
-			signed: signature?.value != null,
+			signed: signed,
 		},
 	);
 
--- a/src/views/nodes/commitNode.ts
+++ b/src/views/nodes/commitNode.ts
@@ -275,7 +275,7 @@ export class CommitNode extends ViewRefNode<'commit', ViewsWithCommits | FileHis
 				allowFilteredFiles: this._options.allowFilteredFiles,
 				include: { stats: true },
 			}),
-			showSignature ? pauseOnCancelOrTimeout(this.commit.getSignature(), cancellation) : undefined,
+			showSignature ? pauseOnCancelOrTimeout(this.commit.isSigned(), cancellation) : undefined,
 		]);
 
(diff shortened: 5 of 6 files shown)


ianhattendorf:
I've pushed up several commits. Still outstanding (highest priority at the top, roughly):

- [x] Signatures are being verified more often than they should be, leading to some delays in the UI
- [x] Need to display an icon on inline blame hovers
- [x] Test X.509 support
- [ ] Should users be able to set up signing without an active repo? (Currently the config sub-provider requires a repoPath)

image
image
image
image

src/env/node/git/sub-providers/config.ts
```diff
@@ -147,4 +158,98 @@ export class ConfigGitSubProvider implements GitConfigSubProvider {
...
+	@log()
+	async getSigningConfig(repoPath: string): Promise<SigningConfig> {
+		const [enabled, format, signingKey, gpgProgram, sshProgram, allowedSignersFile] = await Promise.all([
+			this.getConfig(repoPath, 'commit.gpgsign', { type: 'bool' }),
+			this.getConfig(repoPath, 'gpg.format'),
+			this.getConfig(repoPath, 'user.signingkey'),
+			this.getConfig(repoPath, 'gpg.program'),
+			this.getConfig(repoPath, 'gpg.ssh.program'),
+			this.getConfig(repoPath, 'gpg.ssh.allowedSignersFile'),
+		]);
```
ianhattendorf: I don't like this (kicking off 6 execs). Need to think a bit on the best approach here (`--list`, `--regexp`, etc.).

src/webviews/apps/shared/components/commit/signature-badge.ts
```diff
@@ -0,0 +1,239 @@
...
+			color: var(--vscode-testing-iconPassed, #73c991);
+		}
+
+		.badge--unknown {
+			color: var(--vscode-editorWarning-foreground, #cca700);
+		}
+
+		.badge--untrusted {
+			color: var(--vscode-testing-iconFailed, #f14c4c);
+		}
```
ianhattendorf, line 34: I'm not familiar with VS Code theming: what variables/fallbacks should I use here?

src/commands/signing/setup.ts
```diff
@@ -0,0 +1,166 @@
...
+
+		// Get signing key
+		const signingKey = await window.showInputBox({
+			title: 'Commit Signing Setup',
+			prompt: `Enter your ${format.value.toUpperCase()} signing key ${format.value === 'ssh' ? '(file path)' : '(key ID)'}`,
+			placeHolder: format.value === 'ssh' ? '~/.ssh/id_ed25519.pub' : 'Your key ID',
+			ignoreFocusOut: true,
+		});
+
+		if (!signingKey) return;
```
ianhattendorf, line 144: Ideally we would use the placeholder value when pressing enter when `format.value === 'ssh'` and nothing has been typed by the user (since it's typically a valid key).

package.json
```diff
@@ -5664,6 +5664,34 @@
...
+				"order": 1400,
+				"properties": {
+					"gitlens.signing.showSetupWizard": {
+						"type": "boolean",
+						"default": true,
+						"markdownDescription": "Specifies whether to show the commit signing setup wizard when signing is not configured",
+						"scope": "window",
+						"order": 10
+					},
+					"gitlens.signing.showSignatureBadges": {
```
ianhattendorf, line 5682: This needs to be respected.

package.json
```diff
@@ -5664,6 +5664,34 @@
...
+						"type": "boolean",
+						"default": true,
+						"markdownDescription": "Specifies whether to show signature verification badges on commits in the Commit Graph and other views",
+						"scope": "window",
+						"order": 20
+					},
+					"gitlens.signing.enableKeyGeneration": {
```
ianhattendorf, line 5689: TODO: support this in this PR?

package.json
```diff
@@ -5664,6 +5664,34 @@
 						"markdownDeprecationMessage": "Deprecated. Use the pre-release of GitLens instead"
 					}
 				}
+			},
+			{
+				"id": "commit-signing",
+				"title": "Commit Signing",
+				"order": 1400,
+				"properties": {
+					"gitlens.signing.showSetupWizard": {
```
ianhattendorf: This needs to be respected (or just removed, since it's just triggered from the command palette).

ianhattendorf:
Regarding the signature icon/tooltip for inline blame hovers: I don't believe it's possible to get a tooltip without making the icon a blue link. Not sure what we want to do here or if I'm missing something, I'll take another look next week.

image
image

package.json
ianhattendorf: Let's push this to a follow up PR to prioritize getting the core features merged first.

src/git/parsers/signatureParser.ts
```diff
@@ -0,0 +1,109 @@
...
+ * @returns TrustLevel
+ */
+function mapTrustLevel(trustStr: string | undefined): TrustLevel {
+	if (!trustStr) return 'unknown';
+
+	const normalized = trustStr.toLowerCase().trim();
+	switch (normalized) {
+		case 'ultimate':
+			return 'ultimate';
+		case 'fully':
```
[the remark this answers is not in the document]
ianhattendorf: I believe Git uses `fully`, not `full`: https://github.com/git/git/blob/ab689ea7f91ab0858e85776f31102203d3ea7b83/gpg-interface.c#L208

`fully` is mapped to `full` to align with GPG and the tense of the other trust levels.

src/env/node/git/sub-providers/patch.ts
```diff
@@ -191,34 +217,95 @@ export class PatchGitSubProvider implements GitPatchSubProvider {
...
+					this.container.telemetry.sendEvent(
+						'commit/signing/failed',
+						{ reason: this.getSigningFailureReason(signingError.reason), format: signingFormat },
+						options?.source,
+					);
+					throw signingError;
+				}
+			}
+
+			debugger;
```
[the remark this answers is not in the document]
ianhattendorf: This debugger statement precedes this PR.

ianhattendorf:
augment review

src/env/node/git/sub-providers/config.ts
```diff
@@ -147,4 +164,114 @@ export class ConfigGitSubProvider implements GitConfigSubProvider {
...
+			updates.push(this.setConfig(repoPath, 'gpg.program', config.gpgProgram, options));
+		}
+		if (config.sshProgram !== undefined) {
+			updates.push(this.setConfig(repoPath, 'gpg.ssh.program', config.sshProgram, options));
+		}
+		if (config.allowedSignersFile !== undefined) {
+			updates.push(this.setConfig(repoPath, 'gpg.ssh.allowedSignersFile', config.allowedSignersFile, options));
+		}
+
+		await Promise.all(updates);
```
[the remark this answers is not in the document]
eamodio: I think we should just use Promise.allSettled at least here -- not sure we need to do more than that.
```suggestion
		await Promise.allSettled(updates);
```

src/env/node/git/sub-providers/config.ts
```diff
@@ -147,4 +164,114 @@ export class ConfigGitSubProvider implements GitConfigSubProvider {
...
+
+	@log()
+	async setSigningConfig(
+		repoPath: string,
+		config: Partial<SigningConfig>,
+		options?: { global?: boolean },
+	): Promise<void> {
+		const updates: Promise<void>[] = [];
+
+		if (config.enabled !== undefined) {
```
eamodio: Unless we need to explicitly look for `undefined` we should opt for the `!= null` pattern we use in most places

src/env/node/git/sub-providers/patch.ts
```diff
@@ -170,13 +190,19 @@ export class PatchGitSubProvider implements GitPatchSubProvider {
...
 	): Promise<string> {
 		const scope = getLogScope();
 
 		if (!patch.endsWith('\n')) {
 			patch += '\n';
 		}
 
+		// Check if we should sign
+		const signingConfig = await this.provider.config.getSigningConfig?.(repoPath);
+		const shouldSign = options?.sign ?? signingConfig?.enabled ?? false;
```
eamodio: Should we move this into the `try` and execute it in parallel with the first git.exec call?

src/env/node/git/sub-providers/config.ts
```diff
@@ -147,4 +164,114 @@ export class ConfigGitSubProvider implements GitConfigSubProvider {
 
 		return gitDir;
 	}
+
+	@log()
+	async getSigningConfig(repoPath: string): Promise<SigningConfig> {
```
eamodio, line 206: Thinking that we should cache this in the GitCache -- thoughts?

src/git/gitProvider.ts
```diff
@@ -453,10 +456,23 @@ export interface GitPausedOperationsSubProvider {
...
+	setConfig?(
+		repoPath: string,
+		key: GitConfigKeys,
+		value: string | undefined,
+		options?: { global?: boolean },
+	): Promise<void>;
 	getCurrentUser(repoPath: string): Promise<GitUser | undefined>;
 	getDefaultWorktreePath?(repoPath: string): Promise<string | undefined>;
 	getGitDir?(repoPath: string): Promise<GitDir | undefined>;
+	getSigningConfig?(repoPath: string): Promise<import('./models/signature').SigningConfig>;
```
eamodio: AI seems to do this too often -- this should just be an import at the top of the file. Same with the instances below

src/env/node/git/sub-providers/config.ts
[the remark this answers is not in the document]
ianhattendorf: Looks like git uses `repo_config_set_multivar_in_file_gently` when writing config values which locks the config with [hold_lock_file_for_update](https://github.com/git/git/blob/ab689ea7f91ab0858e85776f31102203d3ea7b83/lockfile.h#L188C19-L188C44), which will error out if there's a lockfile, not wait (timeout is set to 0). So it's possible we fail to write some of these config values if they happen concurrently.

I think we need to await them sequentially to avoid this.

src/git/gitProvider.ts
ianhattendorf: Interesting, I've had that happen with some LSP actions before as well. Might be from the AI using LSP actions.

src/env/node/git/sub-providers/config.ts
ianhattendorf: Hmm, I don't think I'd want to do that unless we could invalidate the cache when the git config changes on disk (repo config, user config, global config, etc.). I don't think we currently have watchers on the gitconfig (outside of maybe the current repo), right?

Is that something we might be interested in? I'd be worried about the extra complication, but maybe it's worth it. Maybe in a future PR?

eamodio:
> Regarding the signature icon/tooltip for inline blame hovers: I don't believe it's possible to get a tooltip without making the icon a blue link. Not sure what we want to do here or if I'm missing something, I'll take another look next week.
> 
> image image

Yeah, only links get tooltips I'm pretty sure

eamodio:
This is ready IMO, though I've disabled the setup wizard for now.

PR #4888, opened 2026-01-16
Updates commit message instructions in auto-compose

### LINKED ISSUE #4870: Pass commit message generation prompt into composer prompt
To ensure that auto-compose is following the same rules for commit messages as the generate commit command, we should pass the prompt that is being used for generating commits into the prompt for auto-composing commits.

This should include any custom instructions a user has added for generating commit messages.

ramin-t:
The commit message section of the auto-compose prompt now more closely matches the existing detailed instructions we had in the generate commit message action.

Note that this cannot be an exact copy-paste/injection of the `generate-commitMessage` prompt itself into the `generate-commits` prompt because there are some subtle differences:
- The message is output as a single string rather than a separated summary and explanation section
- The "explanation" part had to be renamed to "details" so the AI agent does not confuse it with the separate "explanation" it needs to output for each commit to the composer.

Closes #4870


ramin-t:
Note that the user can provide custom instructions for commit messages in three places now:
1. Generate commit message setting
2. Generate commits setting
3. Input box when clicking auto-compose

Updated the instructions to prioritize them in order from lowest to highest in this order.

Issue #4943, opened 2026-01-30
4942: Cached values from the previous account


sergeibbb:
### Steps to reproduce
- connect one GitHub account on GKDev
- sync account on GitLens, reload window, open launchpad -> you get this GitHub account cached
- go to GKDev disconnect GitHub and connect another account.
- sync accounts on GitLens, open launchpad

Expected result
- launchpad gets the **new** GitHub account as "currentAccount"

Current result
- launchpad gets the **old cached** GitHub account as "currentAccount"

sergeibbb:
Hi @eamodio 

What do you think about this problem? Any advice on the right approach to solve? I've shared my thoughts at the "possible solutions" section of the description.

sergeibbb:
Hi @axosoft-ramint 

What do yo think about this ticket?

ramin-t:
@sergeibbb If an account or auth change with an integration is detected during sync (which would cause us to change the stored auth session), right there we should invalidate all caching we have tied to that account for that integration.

Ideally this is tied to account where we can, rather than new auth for the same account, but that can be difficult to detect in some cases.

sergeibbb:
@axosoft-ramint 

> right there we should invalidate all caching we have tied to that account for that integration.

So, rather than modifying etag, you suggest us to force invalidation when we detected the account update, right?

I doubt, because what's difference with "signing-out" and "signing-in" case that is handled currently? Now we do not invalidate caches on logout/login — instead we use different etags: `${integration.id}:${integration.maybeConnected ?? false}`
(e.g. `gitlab:true`, `gitlab:false`).

> Ideally this is tied to account where we can, rather than new auth for the same account, but that can be difficult to detect in some cases.

There is the cached _account_ and _other values._ So we can handle them differently:
- With _etag_ technique we can tie the account's etag to the session, and all other fields' to the account value.
- With invalidation strategy we can force account to refresh when the session is updated
    - and when we see that the account is updated then we can invalidate all other values.

Invalidating sounds harder than _etag_, because sounds like invalidating is detached from retrieving the new value. Imagine:
- we see a new session -> we invalidate the account. But that does not mean immediate retrieving the new account value.
- then when it's time to retrieve the current account, we cannot detect whether it's updated or not, because the old value has been dropped from cache earlier.

nikolay-1986:
Verifiacation: Passed

Scenario 1: Switching GitHub accounts (main bug)

1. Connect the first GitHub account via GKDev.
[...]

Issue #4942, opened 2026-01-30
Invalidate cache on account update


sergeibbb:
### Issue 1. Cached values from the previous account

Steps to reproduce
- connect one GitHub account on GKDev
- sync account on GitLens, reload window, open launchpad -> you get this GitHub account cached
- go to GKDev disconnect GitHub and connect another account.
- sync accounts on GitLens, open launchpad

Expected result
- launchpad gets the **new** GitHub account as "currentAccount"

Current result
- launchpad gets the **old cached** GitHub account as "currentAccount"

### Issue 2. swallowing of an Authentication Error
- connect GitLab account on GKDev
- sync account on GitLens
- revoke the GK application on GitLab
- reload VSCode window -> that makes the current account set to "undefined"
- open launchpad

Expected result:
- no PRs
- **Auth error indication**

Actual result:
- **no error indication,** just no PRs:
    <img width="400" height="398" alt="Image" src="https://github.com/user-attachments/assets/85453947-2847-4f87-80cb-e10bc41483f2" />

## GitLens and VSCode
- GitLens: 17.9 -> pre-release: 2026.1.3004
- VSCode: 1.109.0-insider

## Related issues
- This is a follow-up of #4492
- Sub-issues:
    - #4943
    - #4944
- Follow-ups:
    - #4979
    - #4980

Issue #4731, opened 2025-11-06
Launchpad does not show Gyo's PRs


sergeibbb:
## The problem
Launchpad shows nothing:

<img width="400" height="278" alt="Image" src="https://github.com/user-attachments/assets/9bd26b94-8cf8-471d-9384-237ab588d02f" />

Logs tells about a problem of getting items. Looks like there is a problem with transforming objects from GitHub to GitLens:

<img width="800" alt="Image" src="https://github.com/user-attachments/assets/e40b33a3-95db-4b0a-ba77-07a90213827a" />

## The investigation

I suggest that we should create a custom assembly with more logging and sent to Gyo. He launches and shares the logs. We look at the output and probably find the problem.

### GitLens Version

17.6.2

### VS Code Version

1.105.1

PR #4945, opened 2026-01-30
Controls cache invalidation or persistance on errors

sergeibbb:
# Description
Solves #4942

Controls cache invalidation or persistence on errors.
The goal is to be able to preserve some types of errors in cache, to avoid retrieving the same failures repeatedly.

Be careful, because it makes #4943 more noticeable by preserving the cached Auth Error even if user re-connects. Workaround for user: reload the window. Probably we do not want to merge this PR until #4943 is done.

## Background
Initially it's been done as part of #4731 but while revealing more and more nuances I decided to ship in smaller changes to make things reviewed and merged faster.

# Checklist

### DIFF
--- a/src/cache.ts
+++ b/src/cache.ts
@@ -90,16 +90,13 @@ export class CacheProvider implements Disposable {
 		) {
 			const cacheController = new CacheController();
 			const { value, expiresAt } = cacheable(cacheController);
-			if (isPromise(value)) {
-				void value.finally(() => {
-					if (cacheController.invalidated) {
-						this.delete(cache, key);
-					}
-				});
-			}
-			return this.set<T>(cache, key, value, etag, expiresAt, options?.expireOnError)?.value as CacheResult<
-				CacheValue<T>
-			>;
+			const setResult = this.set<T>(cache, key, value, etag, expiresAt, options?.expireOnError);
+			void setResult?.promise?.finally(() => {
+				if (cacheController.invalidated) {
+					this.delete(cache, key);
+				}
+			});
+			return setResult?.item.value as CacheResult<CacheValue<T>>;
 		}
 
 		return item.value as CacheResult<CacheValue<T>>;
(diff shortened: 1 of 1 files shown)


src/cache.ts
```diff
@@ -249,17 +262,14 @@ export class CacheProvider implements Disposable {
...
-			void value.then(
-				v => {
-					this.set(cache, key, v, etag, expiresAt);
-				},
-				() => {
-					this.delete(cache, key);
-				},
-			);
+			void value.then(v => this.set(cache, key, v, etag, expiresAt, expireOnError));
+			if (expireOnError !== false) {
```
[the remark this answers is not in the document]
sergeibbb: This is intentional for having some rejected promise cached forever. However, we might add some improvements in #4943.

src/plus/integrations/models/integration.ts
```diff
@@ -603,19 +603,31 @@ export abstract class IntegrationBase<
...
 						this.handleProviderException('getCurrentAccount', ex, { scope: scope });
-						return undefined;
+
+						// Invalidate the cache on error, except for auth errors
+						if (!(ex instanceof AuthenticationError)) {
+							cacheable.invalidate();
+						}
+
+						// Re-throw to the caller
+						throw ex;
```
ramin-t, line 625: We are throwing the error up the chain now on every error except for cancellations, where before these changes we would swallow the error and return `undefined`.

Just want to make sure this is intended and we have considered side effects of this logic change.

sergeibbb: @axosoft-ramint 

Yes — it's intended. Actually the main goal of the current PR is being able to throw while keep caching the result (even an error result).

The outcome is the following:
- Once the `undefined` was cached, the Launchpad stopped indicating that there is a problem with GitLab authentication token: it just showed an empty list. We wanted to throw, but before these changes the error result was not cached so we would hit the same endpoint again and again, receiving the same problematic result.
- Now, when it throws, it will display the note that an error has happened. Even if it's some other (non-auth) error, it seems useful to keep the clarity for user and indicate that part of PRs is not shown due to an error.

src/cache.ts
```diff
@@ -85,8 +88,15 @@ export class CacheProvider implements Disposable {
...
-			return this.set<T>(cache, key, value, etag, expiresAt)?.value as CacheResult<CacheValue<T>>;
+			const cacheController = new CacheController();
+			const { value, expiresAt } = cacheable(cacheController);
+			const setResult = this.set<T>(cache, key, value, etag, expiresAt, options?.expireOnError);
+			void setResult?.promise?.finally(() => {
+				if (cacheController.invalidated) {
+					this.delete(cache, key);
+				}
+			});
+			return setResult?.item.value as CacheResult<CacheValue<T>>;
```
ianhattendorf, line 99: Not sure how we typically deal with unhandled promise rejections, but if `options.expireOnError` is `false`, `set` will not add a `.catch()` to the promise.

Would it make sense to add a catch-all `promise.catch(() => {})` either here or in `set`?

sergeibbb: @ianhattendorf 

Fixed. I've added to `set`.

axosoft-ramint (approved):
Thanks for the explanation!

PR #4956, opened 2026-02-04
Moves gk metadata to a dedicated git config file

eamodio:
### Purpose
Reduces noise in the primary Git configuration by migrating GitKraken-specific metadata to a dedicated storage location. This prevents extension-specific data from cluttering user-facing config files and avoids triggering unnecessary repository change events for standard Git operations.

### Changes
- Introduces a new configuration file located at `.git/gk/config` specifically for extension metadata like branch merge targets, associated issues, and access timestamps.
- Implements specialized configuration providers with built-in fallbacks to ensure backward compatibility with existing metadata stored in the primary Git config.
- Updates repository watchers to specifically monitor the new metadata file for changes.
- Optimizes metadata caching by sharing GitKraken configuration across worktrees through the common `.git` directory.
- Refines configuration types to clearly distinguish between core Git settings and extension-specific properties.

### DIFF
--- a/src/env/node/git/sub-providers/branches.ts
+++ b/src/env/node/git/sub-providers/branches.ts
@@ -542,26 +542,18 @@ export class BranchesGitSubProvider implements GitBranchesSubProvider {
 						if (data == null) {
 							const defaultBranch =
 								(await this.provider.config.getConfig(repoPath, 'init.defaultBranch')) ?? 'main';
-							const branchConfig = await this.provider.config.getConfigRegex(
+							const configMap = await this.provider.config.getConfigRegex(
 								repoPath,
 								`branch\\.${defaultBranch}\\.+`,
 								{ runGitLocally: true },
 							);
 
-							let remote;
-							let remoteBranch;
+							const remote = configMap.get(`branch.${defaultBranch}.remote`);
+							const merge = configMap.get(`branch.${defaultBranch}.merge`);
+							const remoteBranch = merge?.startsWith('refs/heads/')
+								? merge.substring('refs/heads/'.length)
+								: undefined;
 
-							if (branchConfig) {
-								let match = /^branch\..+\.remote\s(.+)$/m.exec(branchConfig);
-								if (match != null) {
-									remote = match[1];
-								}
-
-								match = /^branch\..+\.merge\srefs\/heads\/(.+)$/m.exec(branchConfig);
-								if (match != null) {
-									remoteBranch = match[1];
-								}
-							}
 							data = [
 								`${defaultBranch}${remote && remoteBranch ? `\n${remote}/${remoteBranch}` : ''}`,
 								undefined,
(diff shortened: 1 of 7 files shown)


eamodio:
augment review

src/env/node/git/sub-providers/branches.ts
```diff
@@ -1135,37 +1127,23 @@ export class BranchesGitSubProvider implements GitBranchesSubProvider {
...
-						update = false;
-						break;
-					} else if (key === 'vscode-merge-base') {
-						mergeBase = value;
-						update = true;
-						continue;
-					}
-				}
+			// getGkConfig has built-in fallback to regular config for backward compatibility
+			let mergeBase = await this.provider.config.getGkConfig(repoPath, `branch.${ref}.gk-merge-base`);
```
ianhattendorf, line 1131: Would it make sense to have a list of keys/patterns that are directed to the gk file? That way `{get,set}Config` can check the list, so that which keys are fetched/saved in `.git/config` vs. `.git/gk/config` is managed centrally vs. decided at each point of use (via `{get,set}GkConfig`).

src/env/node/git/sub-providers/config.ts
```diff
@@ -36,67 +56,109 @@ export class ConfigGitSubProvider implements GitConfigSubProvider {
...
+
+		if (options?.file) {
+			args.push('-f', options.file);
+		} else {
+			const global = options?.global || repoPath == null;
+			args.push(global ? '--global' : '--local');
+		}
+
+		if (value == null) {
+			args.push('--unset', key);
```
ianhattendorf, line 148: I'm wondering if we might have issues with new values being written to the new location when there's currently a value in the old location. If we later delete the config key, instead of there not being a value there's still a value in the old location that will be picked up.

src/env/node/git/sub-providers/branches.ts
eamodio: Not sure I'm following — we have:

https://github.com/gitkraken/vscode-gitlens/blob/1eeba930277185330a815ba3c28aa399688f37e5/src/constants.ts#L50-L83

Which only let you to call the correct method.

ianhattendorf: My thinking is to just have a single method to call: `getConfig`. That can then call out to something like `getConfigCore` for the normal git keys or `getGkConfig` for our keys.

Just an idea to reduce the API surface slightly by avoiding some duplication, but if that doesn't work or isn't ideal for some reason for GL that's fine too.

eamodio: Ah, that is the way I had it originally, but it felt to "magical" especially with the regex support. So I'd rather the callers make the conscious decision of where to store a particular config setting

src/env/node/git/sub-providers/config.ts
eamodio: Yeah, I added a one-time best effort migration, and no more fallbacks

eamodio:
I added a one-time best effort migration, and no more fallbacks.

Issue #1803, opened 2022-02-07
Gitlens crashes extension host when clicking on submodule commit change file in treeview


ulugbekna:
- GitLens Version: 11.7.0
- Git Version: 2.32.0

- VSCode Version: 
```
1.64.0
5554b12acf27056905806867f251c859323ff7e9
x64
```
- OS Version: macOS Big Sur 11.6 (20G165)

Hi, when I click on a file that represents the update of a commit of a git submodule, the extension host terminates because of an exception thrown by gitlens

```
  ERR [Extension Host] (node:13650) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
(Use `Electron --trace-deprecation ...` to show where the warning was created)
workbench.desktop.main.js:75 [Extension Host] (node:13650) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.(Use `Electron --trace-deprecation ...` to show where the warning was created) (at writeOut (internal/process/warning.js:43:3))
I @ workbench.desktop.main.js:75
workbench.desktop.main.js:3036 Extension host (LocalProcess) terminated unexpectedly. Code: 7, Signal: null
_onExtensionHostCrashed @ workbench.desktop.main.js:3036
workbench.desktop.main.js:578   ERR Extension host (LocalProcess) terminated unexpectedly. The following extensions were running: vscode.microsoft-authentication, vscode.debug-auto-launch, vscode.git-base, GitHub.vscode-pull-request-github, vscodevim.vim, ms-vscode-remote.remote-containers, ocamllabs.ocaml-platform, vscode.github-authentication, vscode.git, vscode.github, vscode.emmet, vscode.merge-conflict, eamodio.gitlens, ms-vscode-remote.remote-wsl
workbench.desktop.main.js:788 Extension host terminated unexpectedly.
onDidChangeNotification @ workbench.desktop.main.js:788
[...]

PR #4960, opened 2026-02-06
Adds basic support for Git submodules

### LINKED ISSUE #1048: GitLens cannot display submodule diffs
- GitLens Version: 10.2.1
  
- VSCode Version: 1.45.1
- OS Version: Ubuntu 20.04

Steps to Reproduce:
1. Create git repository with submodule(s), create a commit involving a submodule.
2. In branch history locate and expand that commit.  Click on the submodule.

Result: 
Error "Unable to open '\<submodule\> \<hash\> ... "  
Looks like VSCode is trying to compare  submodule directory as if it were a file.

Expected:
Display submodule hash diff, the way VSCode's git integration does this:
```
diff --git a/submodule b/submodule
--- a/submodule
+++ b/submodule
@@ -1 +1 @@
-Subproject commit b1601e42183b31af1afc6ceedd4cb9b3efb27f0d
+Subproject commit 388ba770105a27e5589e55ad56dcf9dbd407d134
```

eamodio:
Improves the handling of Git submodules by treating them as first-class repository entities alongside worktrees. This update ensures that submodules (that are discovered by VS Code) are correctly identified and provides better visual differentiation within the user interface.

### Key Enhancements

- **Submodule Discovery:** Enhances repository tracking by identifying nested submodules using `rev-parse` and tracking parent repository relationships.
- **Improved UI/UX:** Distinguishes submodules in the repository tree and quick picks with specific icons and tooltips. 
- **Contextual Menu Control:** Updates menu "when" clauses to filter out submodule-specific files from certain actions where they are not yet supported, such as file-level sharing or deep linking.
- **Diff & Navigation:** Adds logic to handle `160000` file modes, allowing for better navigation and diffing of subproject commits.

Closes #1048, #1803

### DIFF
--- a/src/env/node/git/sub-providers/commits.ts
+++ b/src/env/node/git/sub-providers/commits.ts
@@ -158,6 +158,7 @@ export class CommitsGitSubProvider implements GitCommitsSubProvider {
 					undefined,
 					undefined,
 					f.mode,
+					f.oid ? { oid: f.oid, previousOid: f.previousOid } : undefined,
 				),
 		);
 
@@ -800,9 +801,11 @@ export class CommitsGitSubProvider implements GitCommitsSubProvider {
 			renames: options?.renames ?? configuration.get('advanced.fileHistoryFollowsRenames'),
 		};
 
+		let type: 'file' | 'folder' | 'submodule' = options.isFolder ? 'folder' : 'file';
 		if (isFolderGlob(relativePath)) {
 			relativePath = stripFolderGlob(relativePath);
 			options.isFolder = true;
+			type = 'folder';
 		} else if (options.isFolder == null) {
 			const tree = await this.provider.revision.getTreeEntryForRevision(repoPath, rev || 'HEAD', relativePath);
 			if (cancellation?.isCancellationRequested) throw new CancellationError();
@@ -810,18 +813,20 @@ export class CommitsGitSubProvider implements GitCommitsSubProvider {
 			if (tree?.type === 'commit') {
 				// It's a submodule — line ranges and rename tracking don't apply, and the rev may reference
 				// the submodule's internal SHA which isn't valid in the parent repo
+				type = 'submodule';
 				options.range = undefined;
 				options.renames = false;
-				rev = undefined;
+				// rev = undefined;
 			} else {
-				options.isFolder = tree?.type === 'tree';
+				type = tree?.type === 'tree' ? 'folder' : 'file';
+				options.isFolder = type === 'folder';
 			}
 		}
 
 		let cacheKey: string | undefined;
 		if (
-			// Don't cache folders
-			!options.isFolder &&
+			// Only cache files
+			type === 'file' &&
 			options.authors == null &&
 			options.cursor == null &&
 			options.filters == null &&
@@ -1442,6 +1447,7 @@ export function createCommitFileset(
 				undefined,
 				f.range ? { startLine: f.range.startLine, endLine: f.range.endLine } : undefined,
 				f.mode,
+				f.oid ? { oid: f.oid, previousOid: f.previousOid } : undefined,
 			),
 	);
 
--- a/src/commands/diffWithPrevious.ts
+++ b/src/commands/diffWithPrevious.ts
@@ -41,7 +41,7 @@ export class DiffWithPreviousCommand extends ActiveEditorCommand {
 		args = { ...args };
 		if (args.uri == null) {
 			uri = getCommandUri(uri, editor);
-			if (uri == null) return;
+			if (uri == null && args.commit == null) return;
 		} else {
 			uri = args.uri;
 		}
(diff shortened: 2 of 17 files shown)


eamodio:
There are many caveats and gaps here — this gets some basic submodule support to largely align with VS Code's support.

src/env/node/git/sub-providers/commits.ts
```diff
@@ -629,8 +632,25 @@ export class CommitsGitSubProvider implements GitCommitsSubProvider {
...
-			Logger.error(ex, scope);
 			if (isCancellationError(ex)) throw ex;
+
+			// Check if this is a "bad object" error due to a submodule's internal SHA
+			const pathspec = options?.path?.pathspec;
+			if (rev && pathspec && ex instanceof Error && GitErrors.badObject.test(ex.message)) {
+				const tree = await this.provider.revision.getTreeEntryForRevision(repoPath, 'HEAD', pathspec);
+				if (tree?.type === 'commit') {
+					// It's a submodule - retry without the rev and without rename tracking
+					return this.getLogCore(
```
ianhattendorf, line 644: Is it possible we enter an infinite loop here, or will `GitErrors.badObject.test(ex.message)` never match if we pass `rev: undefined` and `renames: false`?

src/env/node/git/sub-providers/commits.ts
```diff
@@ -629,8 +632,25 @@ export class CommitsGitSubProvider implements GitCommitsSubProvider {
...
+						repoPath,
+						undefined,
+						{ ...options, path: { ...options.path, pathspec: pathspec, renames: false } },
+						cancellation,
+						additionalArgs,
+					);
+				}
+			}
+
+			Logger.error(ex, scope);
```
ianhattendorf, line 654: I hit the below debugger statement by:
1. Opening https://github.com/ianhattendorf/submodule-test
2. Opening the commit graph
3. Selecting the commit `Add nodegit` d847aa7
4. Clicking `.gitmodules` under `2 files changed` in `inspect: commit details`, and then `nodegit`

```
[GitLens      ] [2026-02-11 00:49:30.007] [GIT:←12      ] (spawn) [/home/ian/dev/test/submodule-test] git log --format=%x1E%H%x1D%aN%x1D%aE%x1D%at%x1D%cN%x1D%cE%x1D%ct%x1D%P%x1D%D%x1D%B%x1D -M --use-mailmap --no-merges -n41 d847aa7f8e6125c4038e1827089a815b7c329b6d -L 2,2:nodegit -- • file nodegit has only 1 lines [9ms]
logger.ts:138
[GitLens      ] [2026-02-11 00:49:30.007] [          884] Git.LogParserWithPatch.parseAsync [10ms]
logger.ts:99
[GitLens      ] [2026-02-11 00:49:30.008] [  881 →   882] CommitsGitSubProvider.getLogCore  GitError: fatal: file nodegit has only 1 lines {cmd: '/usr/bin/git -c core.quotepath=false -c color.…25c4038e1827089a815b7c329b6d -L 2,2:nodegit --', exitCode: 128, stdout: '', stderr: 'fatal: file nodegit has only 1 lines', original: RunError: Error (128): fatal: file nodegit has only 1 lines, …}
```

src/git/models/repository.ts
```diff
@@ -203,7 +203,9 @@ export class Repository implements Disposable {
 		public readonly root: boolean,
 		closed: boolean = false,
 	) {
-		if (_gitDir?.commonUri != null) {
+		if (_gitDir?.parentUri != null) {
+			this._type = 'submodule';
+		} else if (_gitDir?.commonUri != null) {
 			this._type = 'worktree';
```
ianhattendorf: What if a repo is both a submodule and a worktree?

eamodio: Yeah — I figured if it was both we'd largely just want to treat it as a submodule (which is also what VS Code does). Though we could remove the `type` prop and just leave the `isSubmodule` and `isWorktree`  props. Or we could leave the `type` as the "primary", but change the `isSubmodule` and `isWorktree` props to not use the type.

What do you think?

src/env/node/git/sub-providers/commits.ts
eamodio: Shouldn't be possible, since the retry sets `rev` to `undefined` so it won't enter again

src/env/node/git/sub-providers/commits.ts
eamodio: This should be resolved in f13e505033dad068619e0509d71b99bb743dd3cf

src/env/node/git/sub-providers/commits.ts
ianhattendorf: Totally spaced and missed the `rev` check above :upside_down_face:

src/git/models/repository.ts
ianhattendorf: Let's leave it for now then, especially since that's essentially what VS Code does. Just wanted to bring it up as something that stood out to me.

ianhattendorf (approved):
I'm still hitting the debugger line, but everything appears to be working so let's get this in and we can follow up in the future if necessary.

PR #5001, opened 2026-02-23
Fixes selection highlight delay when selecting composer commits with large commits

### LINKED ISSUE #4872: Recompose performance degrades when each commit is selected
In a UX session with Eric, we discovered that composer takes longer to render changes as more commits are selected during recomposition.

ramin-t:
Closes #4872

Reduces the time needed to highlight a commit in the commits panel of the composer when it is selected or shift-selected.

Does this by:
1. Using caching/memoization to reduce redundant calculations of commits and hunks and diff/file counts in both the app and commits panel.
2. Delaying the event firing from the commits panel to the app using double-RAF
3. Fixing some redundant processing that happens on diff files when initially loaded, and uses scheduling for updates
4. Removing the transition from commit highlighting when selected

### DIFF
--- a/src/webviews/apps/plus/composer/components/commits-panel.ts
+++ b/src/webviews/apps/plus/composer/components/commits-panel.ts
@@ -1024,20 +1024,16 @@ export class CommitsPanel extends LitElement {
 
 	/**
 	 * Dispatches an event after the browser has painted the current frame.
-	 * Uses double `requestAnimationFrame`: the first fires before the current
-	 * frame's paint, the second fires before the NEXT frame's paint — by which
-	 * point the current frame has been painted. This guarantees that local
-	 * selection state is visually rendered before the parent starts heavy work
-	 * (e.g. rendering diffs in the details panel).
+	 * Uses `setTimeout(0)` — per the spec the browser runs the rendering
+	 * step (style, layout, paint) between the current macrotask and the
+	 * next, so local DOM changes are painted before this callback fires.
 	 */
 	private dispatchAfterPaint(event: Event) {
 		const gen = ++this._pendingDispatch;
-		requestAnimationFrame(() => {
-			requestAnimationFrame(() => {
-				if (gen !== this._pendingDispatch || !this.isConnected) return;
-				this.dispatchEvent(event);
-			});
-		});
+		setTimeout(() => {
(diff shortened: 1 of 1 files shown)


src/webviews/apps/plus/composer/components/app.ts
```diff
@@ -1202,12 +1202,22 @@ export class ComposerApp extends LitElement {
...
+
 	private get hunksWithAssignments(): ComposerHunk[] {
 		if (!this.state?.hunks || !this.state?.commits) {
 			return [];
 		}
 
-		return updateHunkAssignments(this.state.hunks, this.state.commits);
+		const { hunks, commits } = this.state;
+		const cached = this._hunksWithAssignmentsCache;
+		if (cached?.hunks === hunks && cached?.commits === commits) {
```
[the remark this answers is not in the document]
ramin-t: We don't mutate hunk content in-place, so this should be fine.

ramin-t:
augment review

ramin-t:
augment review

src/webviews/apps/plus/composer/components/commits-panel.ts
```diff
@@ -925,14 +1008,38 @@ export class CommitsPanel extends LitElement {
...
+	 * (e.g. rendering diffs in the details panel).
+	 */
+	private dispatchAfterPaint(event: Event) {
+		const gen = ++this._pendingDispatch;
+		requestAnimationFrame(() => {
+			requestAnimationFrame(() => {
+				if (gen !== this._pendingDispatch || !this.isConnected) return;
+				this.dispatchEvent(event);
+			});
+		});
```
d13: Does the double `requestAnimationFrame` provide better efficiency/performance over using `setTimeout(0)`?

d13 (approved):
Looks good, but was curious about using double `requestAnimationFrame` versus using `setTimeout(0)`.

src/webviews/apps/plus/composer/components/commits-panel.ts
ramin-t: I had initially thought doubleRAF had a better guarantee of the logic hitting the exact timing I wanted before your comment, but after doing a bit of research, it looks like setTimeout(0) should get us there (past the first paint) pretty reliably and might even be a touch faster, so I'll switch to it. Thanks!

PR #5012, opened 2026-02-27
Fixes AuthenticationError meta-info not appearing in logs

### LINKED ISSUE #4880: 4879: Provide the Authentication error with token type and scopes
Provide token type and scopes, to the Authentication error, so when it happen it's easier for us to reproduce

GK gives us information of whether the token is PAT or OAuth. Also when we get the VSCode's GitHub token we can know it as well.

To sum up, when an AuthenticationError appears in log we want to see:
- integration type
- token type
- scopes

### Steps to reproduce

- Connect a GitHub integration
- Go to [GitHub.com authorized applications settings](https://github.com/settings/applications) and revoke the application
- Navigate git content in GitLens until you start seeing Authentication errors:
    ```
    Error: Your 'github' credentials are either invalid or expired
    ```
- We want to add more details there, for example, if we've sent user through the re-authentication flow we want to know whether it helps or not.

### Verification

Example of a produced error log after the improvement:

```
Error: Your 'github' credentials are either invalid or expired (token details: self-managed, @4cb, [repo,read:user,user:email])

or

Error: Your 'github' credentials are either invalid or expired (token details: cloud,  oauth, @e75, expires at 2060-01-30T03:52:47.500Z, [repo,read:user,user:email])
```

- [x] real token does not appear in logs
- [ ] undefined values do not pollute the message, e.g. you do not see something like this:
    ```
    (token details: self-managed, , cloud, , [repo])
    ```
- [ ] While additional info is presented in logs, it does not pollute the pop-up message:
    Image
    Image
- [ ] Make sure that the Graph keeps working, e.g. `github.searchCommitShas()` has been affected by the change, so it's worth to check that it's not broken.
- [ ] Make sure that various `paged` results keep working, since the `providersApi.ts -> ProvidersApi.getPagedResult()` is affected too.

sergeibbb:
Fixes #4880

After the logging migration to VS Code's LogOutputChannel (bce3282), the token details stopped being logged because LogOutputChannel uses `.message` rather than `toString()`.

- Includes authInfo in the Error's `.message` property for logging
- Adds `userMessage` property with clean text for user-facing dialogs
- Updates GitHub and GitLab providers to use `userMessage` in error popups
- Removes the now-redundant `toString()` override

# Checklist

### DIFF
--- a/src/errors.ts
+++ b/src/errors.ts
@@ -54,7 +54,10 @@ export class AuthenticationError extends Error {
 	readonly id: string;
 	readonly original?: Error;
 	readonly reason: AuthenticationErrorReason | undefined;
+	/** Token details for debugging (included in `.message` for logging) */
 	readonly authInfo: string;
+	/** User-friendly message without token details */
+	readonly userMessage: string;
 
 	constructor(info: TokenInfo, reason?: AuthenticationErrorReason, original?: Error);
 	constructor(info: TokenInfo, message?: string, original?: Error);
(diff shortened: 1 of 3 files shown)


eamodio:
I think we should back this out, and instead fix the behavior change where we pass the Error object directly to the VS Code logger -- looking at the VS Code logger code, I don't like how they are logging errors, so we should go back to our existing behavior.

https://github.com/gitkraken/vscode-gitlens/blob/04b6535518548b03da1e03a837bfa3c49221ac1a/src/system/logger.ts#L221

So we'd just need to get rid of the `ex instanceof Error` check and always do the `String(ex)`, which should fix everything

PR #5016, opened 2026-03-02
Adds tel events for conflict-prone git ops, settings/diff metadata for AI events

ramin-t:
- Adds new git op events for git commands that would cause conflicts and new events for when conflicts occur as they are run
- Adds diff info, custom instruction setting info, and diff hash to the metadata for ai/generate events: generate commits (composer's auto-compose) and generate commit message
- Adds diff hash to composer telemetry metadata/context


ramin-t:
Since this is fairly low risk and only updates telemetry, I will merge it

PR #5035, opened 2026-03-10
Fixes stale cached data on account switch

sergeibbb:
Fixes: #4943

Prevents cached data from a previous account from displaying after an account switch by incorporating a fingerprint of the active authentication session into cache ETags.

This ensures that cached items tied to an integration or resource are re-fetched when the underlying account's access token changes, invalidating data specific to the old account. A 64-bit FNV-1a hash is introduced to generate robust session fingerprints.

Also adds a `peek` method to the cache to safely retrieve resolved cached values without triggering a fetch.

# Checklist

### DIFF
--- a/src/system/__tests__/hash.test.ts
+++ b/src/system/__tests__/hash.test.ts
@@ -1,0 +1,72 @@
+import * as assert from 'assert';
+import { fnv1aHash, fnv1aHash64 } from '../hash.js';
+
+/**
+ * Test vectors from RFC 9923 (FNV Non-Cryptographic Hash Algorithm), Section 8.3.
+ * The RFC defines 4 test strings: "", "a", "foobar", "Hello!\x01\xFF\xED".
+ * The 4th string contains raw bytes 0xFF/0xED which map to different charCodeAt
+ * values than the byte values, so it is excluded from these tests.
+ *
+ * The 32-bit implementation uses `(hash * prime) | 0` which loses precision
+ * when the product exceeds 2^53 (JS safe integer limit). The "" and "a" vectors
+ * match the RFC; "foobar" diverges. The 64-bit implementation avoids this by
+ * splitting into hi/lo 32-bit halves and matches the RFC vectors exactly.
+ *
+ * @see https://www.rfc-editor.org/rfc/rfc9923.html — Section 8.3 (FNV Test Code)
+ */
+
+suite('Hash Test Suite', () => {
+	// FNV-1a 32-bit test vectors: [input, expected hash, source]
+	// RFC 9923 §8.3: FNV32svalues = { 0x811c9dc5, 0xe40c292c, 0xbf9cf968, 0xfd9d3881 }
+	const fnv1a32Vectors: [string, number][] = [
+		['', 0x811c9dc5], // RFC 9923 §8.3 — https://www.rfc-editor.org/rfc/rfc9923.html
+		['a', 0xe40c292c], // RFC 9923 §8.3 — https://www.rfc-editor.org/rfc/rfc9923.html
+		['foobar', 0xbf9cf968], // RFC 9923 §8.3 — https://www.rfc-editor.org/rfc/rfc9923.html
+	];
+
+	// FNV-1a 64-bit test vectors: [input, expected hash]
+	// RFC 9923 §8.3: FNV64svalues = { 0xcbf29ce484222325, 0xaf63dc4c8601ec8c, 0x85944171f73967e8, 0xbd51ea7094ee6fa1 }
+	const fnv1a64Vectors: [string, string][] = [
+		['', 'cbf29ce484222325'], // RFC 9923 §8.3 — https://www.rfc-editor.org/rfc/rfc9923.html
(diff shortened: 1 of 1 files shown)


src/cache.ts
```diff
@@ -318,6 +325,26 @@ export class CacheProvider implements Disposable {
...
+			etag: `${integration.id}:${integration.maybeConnected ?? false}:${integration.sessionFingerprint ?? ''}`,
+		};
+	}
+
+	/** ETag for resource caches — uses cached account ID when available, falls back to sessionFingerprint */
+	private getResourceKeyAndEtag(resource: ResourceDescriptor, integration?: GitHostIntegration | IntegrationBase) {
+		let fingerprint = '';
+		if (integration != null) {
+			fingerprint =
+				this.peek('currentAccount', `id:${integration.id}`)?.id ?? integration.sessionFingerprint ?? '';
```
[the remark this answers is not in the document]
sergeibbb: No change needed. The window is tiny, strictly better than before, and closes on the next getCurrentAccount call. Fixing it would couple the cache layer to auth lifecycle for marginal gain.

src/plus/integrations/models/integration.ts
```diff
@@ -127,6 +128,16 @@ export abstract class IntegrationBase<
 		return this._session === undefined ? undefined : this._session !== null;
 	}
 
+	/** Hash of the current session's access token. Changes on any token change (account switch or refresh). */
+	private _sessionFingerprint: { session: ProviderAuthenticationSession; hash: string } | undefined;
+	get sessionFingerprint(): string | undefined {
+		if (this._session == null) return undefined;
+		if (this._sessionFingerprint?.session !== this._session) {
```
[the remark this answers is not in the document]
sergeibbb: The concern is that if a provider mutates accessToken on the same session object, the fingerprint would go stale. No change needed. VS Code's AuthenticationSession objects are immutable — the authentication API returns a new session object on any change (token refresh, account switch). The !== reference check is correct and idiomatic.

src/cache.ts
```diff
@@ -318,6 +325,26 @@ export class CacheProvider implements Disposable {
...
+		}
+		return {
+			key: resource.key,
+			etag: `${resource.key}:${integration?.maybeConnected ?? false}:${fingerprint}`,
```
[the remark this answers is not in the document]
sergeibbb: Copilot suggests embedding the full integration etag (which already includes sessionFingerprint) into the resource etag, eliminating the peek entirely. This would close the transient window but at a cost: the integration etag includes maybeConnected and the full fingerprint, meaning token refreshes (same account) would also bust all resource caches unnecessarily. The current design intentionally prefers account ID (stable across refreshes) for resource caches. No change needed — the suggested code is correct but trades cache efficiency for a marginal consistency improvement.

src/system/hash.ts
```diff
@@ -10,3 +10,33 @@ export function fnv1aHash(str: string): number {
...
+/**
+ * FNV-1a 64-bit hash — fast, non-cryptographic hash with excellent distribution.
+ * Uses split hi/lo 32-bit integers to avoid BigInt allocation overhead.
+ * The 64-bit prime 0x100000001b3 decomposes as (0x100 << 32) + 0x1b3,
+ * keeping all intermediate products under 2^41 (safe for JS number math).
+ *
+ * Why 64-bit instead of 32-bit: the 32-bit variant's birthday-bound collision
+ * probability (P ≈ 1 - e^(-k² / 2^33)) becomes problematic at scale — at 5M
+ * users generating ~8,760 hourly tokens/year, ~10 users/year would experience
+ * a consecutive-token collision.
```
ianhattendorf: I don't think this comment is accurate for how we're using it. Locally, we shouldn't have more than a handful of user accounts to potentially collide, so 32-bit would be fine.

ianhattendorf: I would also argue that when implementing a hash function ourselves, we should have at least a few test cases using known test vectors (e.g. from https://www.ietf.org/archive/id/draft-eastlake-fnv-21.html#name-a-few-test-vectors).

I realize the 32-bit implementation doesn't have any tests either, but we can always leave the code better than we found it 🙂

sergeibbb: @ianhattendorf 

I've covered the functions with tests.

Also I've updated the description to make it more explanatory:

> The probability of successful token regeneration without a hash collision is
> `P1 ≈ 1 - 2^-32`, which is very high.
> But if the token is updated hourly, then the probability for a user
> of having a year without consecutive collisions:
> `PY = P1^hours`
> which measn that probability of having at least one collision is:
> `EY = 1 - PY = 2 * 10^-6`
> Which means that **2 users of 1M** have a chance to have it.
>
> It does not mean that they open an issue, they probably won't even notice.
> But it might happen with a real person in real life.
>
> By using `fnv1aHash64` we reduce the chance to 0.

sergeibbb: @ianhattendorf 

Also I've updated 32bit version by using `Math.imul` instead of usual multiplication to satisfy specs.

PR #5167, opened 2026-04-24
Centralize signing-error classification across git-cli operations

### LINKED ISSUE #5166: Centralize signing-error classification across git-cli operations
## Problem

`SigningError` (typed reasons `passphraseFailed | noKey | gpgNotFound | sshNotFound | unknown`) is defined in `packages/git/src/errors.ts` but is only thrown from **one** place today: `packages/git-cli/src/providers/patch.ts` inside `createUnreachableCommitForPatchCore`. Every other signing-capable operation in `packages/git-cli/src/providers/operations.ts` — `commit`, `merge`, `pull`, `rebase`, `revert`, `cherryPick` — funnels its catch through `getGitCommandError` (`packages/git-cli/src/exec/git.ts`), whose error-to-reason map contains zero signing regexes. As a result:

- A `-S` failure during any of those operations surfaces as `CommitError { reason: 'other' }` / `MergeError { reason: 'other' }` / etc., with the signing info buried in `.original.stderr`.
- The `hooks.commits.onSigningFailed` hook (declared in `packages/git/src/context.ts`) **never** fires from those paths.
- External consumers have to duplicate the stderr-matching logic that already exists in patch.ts.

The library clearly knows how to classify these errors — the ladder in patch.ts does it — but the classifier is private, duplicated-in-waiting, and not applied to the paths that need it.

## Acceptance

- `SigningError` thrown from all six signing-capable operations in `OperationsGitSubProvider`, not just patch.ts.
- `hooks.commits.onSigningFailed` fires from every signing-capable path with `(reason, format, source)`.
- Signing classifier exported so external consumers can drop forked stderr-matching.
- Existing typed reasons on `CommitError` / `MergeError` / `PullError` etc. continue to work for non-signing failures (no regressions).
- Signing regexes covered by unit tests in `gitErrors.test.ts`.

ianhattendorf:
Closes #5166.

## Summary

- Adds signing stderr regexes to `GitErrors` and exports pure `classifySigningError` / `inferSigningFormatFromError` helpers from `exec/git.ts` as the single source of truth — replaces the inline ladder in `patch.ts` and tightens the previously over-broad `error: gpg` substring to an anchored `/^error:\s*gpg failed to sign the data/im`
- Adds `throwIfSigningError` to `OperationsGitSubProvider` and wires it into every signing-capable catch (`commit`, `merge`, `pull`, `rebase`, `revert`, `cherryPick`), firing `hooks.commits.onSigningFailed(reason, format, source)` and throwing a typed `SigningError` so those operations no longer collapse signing failures into `CommitError { reason: 'other' }` / `MergeError { reason: 'other' }` / etc.

## Test plan

- [x] Manual end-to-end smoke: extension host with `commit.gpgsign=true` and an invalid `user.signingkey`, run **Git: Commit** — expect `SigningError` surface + `commit/signing/failed` telemetry event, not a generic "Unable to commit"

### DIFF
--- a/packages/git-cli/src/exec/__tests__/gitErrors.test.ts
+++ b/packages/git-cli/src/exec/__tests__/gitErrors.test.ts
@@ -1,6 +1,13 @@
 import * as assert from 'assert';
 import { RunError } from '../exec.errors.js';
-import { getGitCommandError, GitError, GitErrors, GitWarnings } from '../git.js';
+import {
+	classifySigningError,
+	getGitCommandError,
+	GitError,
+	GitErrors,
+	GitWarnings,
+	inferSigningFormatFromError,
+} from '../git.js';
 
 /**
  * Helper: creates a GitError wrapping a RunError with the given stderr.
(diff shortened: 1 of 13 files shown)


julianmesa-gitkraken (commented):
One spot looks like it was missed when wiring `SigningError` through the UX layer:

**`src/git/gitRepositoryService.ts:369-373` (`pullCore`)**

The four QuickCommand handlers (`cherry-pick`, `merge`, `rebase`, `revert`) were updated to let `SigningError` pass through with its own message, but the pull handler at the service level still only checks `PullError.is(ex)`:

```ts
if (PullError.is(ex)) {
    void showGitErrorMessage(ex);
} else {
    void showGitErrorMessage(ex, 'Unable to pull');
}
```

Now that `ops.pull` can throw `SigningError` (e.g. `pull --rebase` with `commit.gpgsign=true` and a broken signing setup), it falls into the `else` branch and gets masked with `Unable to pull`, which defeats the centralization the PR is aiming for. IMO same pattern as the other four:

```ts
if (PullError.is(ex) || SigningError.is(ex)) {
    void showGitErrorMessage(ex);
} else {
    void showGitErrorMessage(ex, 'Unable to pull');
}
```

The why: the whole point of throwing a typed `SigningError` from `ops.pull` is so the UI surfaces the specific signing-failure message (`Unable to sign commit because…`) instead of a generic op-level fallback. Without this, signing failures during pull look identical to any other pull error to the user.

ianhattendorf:
> One spot looks like it was missed when wiring `SigningError` through the UX layer:
> ...

Good catch, added that and force pushed (2 line change: import + if statement in `pullCore`).

Issue #5143, opened 2026-04-15
Introduce new commit details experience in Commit Graph


trevor-polidore:
### Description

Introduce a revamped commit details experience integrated in the graph with:

Relates to this exploration issue https://github.com/gitkraken/vscode-gitlens/issues/4876

PR #5184, opened 2026-04-29
Remove broken "Inspect Details" context menu from Graph view

### LINKED ISSUE #5183: Graph: Remove broken "Inspect Details" context menu item
## Description

The **"Inspect Details"** context menu item in the Graph view does not work — the embedded details panel never opens when clicked.

The root cause is that the `gl-graph-request-inspect` DOM event is dispatched on `gl-graph-apphost` (an ancestor element) but the listener sits on `div.graph__workspace` inside `gl-graph-app` (a descendant). Since DOM events bubble **up** the tree, the event never reaches the listener.

Rather than fixing the broken plumbing, this item is being **removed** — the embedded details panel is already accessible via double-click on a commit row, making the context menu entry redundant.

## Likely source

Introduced in 0ebf629cf1 ("Embeds details panel directly into the graph").

## Parent Issue

Part of #5143

## Verification steps
- [ ] Right-click a commit row in Graph → "Inspect Details" should NOT appear
- [ ] Double-click a commit row → embedded details panel still opens
- [ ] Build passes (`pnpm run build`)

sergeibbb:
## Summary

- Removes the **"Inspect Details"** context menu item from the Graph view commit rows
- The menu item never worked — the `gl-graph-request-inspect` DOM event was dispatched on `gl-graph-apphost` (ancestor) but the listener was on `div.graph__workspace` (descendant). Since events bubble up, the event never reached the listener
- The embedded details panel is already accessible via **double-click**, making this entry redundant

## Likely source

Broken since 0ebf629cf1 ("Embeds details panel directly into the graph").

## What's removed

- `gitlens.graph.showInDetailsView` command definition and menu entry from `contributions.json`
- `openInDetailsView` command handler from `graphWebview.ts`
- `DidRequestInspectNotification` IPC notification from `protocol.ts`
- Notification handler from `stateProvider.ts`
- `handleRequestInspect` method and template listener from `graph-app.ts`

## Related issues

- Closes #5183
- Part of #5143

### DIFF
--- a/contributions.json
+++ b/contributions.json
@@ -3662,19 +3662,6 @@
 				]
 			}
 		},
-		"gitlens.graph.showInDetailsView": {
-			"label": "Inspect Details",
-			"icon": "$(eye)",
-			"menus": {
-				"webview/context": [
-					{
-						"when": "webviewItem =~ /gitlens:(commit|stash|wip)\\b/ && !listMultiSelection",
-						"group": "3_gitlens_explore",
-						"order": 0
-					}
-				]
-			}
-		},
 		"gitlens.graph.soloBranch": {
 			"label": "Solo Branch",
 			"commandPalette": "gitlens:enabled",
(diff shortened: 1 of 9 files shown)


sergeibbb:
@eamodio @d13 @ianhattendorf 

It should be real quick to review: deletions only.

eamodio (approved):

PR #5199, opened 2026-05-07
Consolidate CLI and agent IPC into a shared service

### LINKED ISSUE #5198: Consolidate CLI and agent IPC into a shared service
## Problem

GitLens currently runs two independent HTTP IPC servers in the extension host:

- One for the GitKraken CLI (`GkCliIntegrationProvider` → `createIpcServer` from the agents package)
- One for agent session hooks (`AgentIpcServer`, also in the agents package)

Each server allocates its own port, generates its own bearer token, and writes its own discovery file. The two servers don't share lifecycle or configuration, even though they're both consumed by `gk` binaries and peer GitLens windows on the same host.

This duplication:

- Wastes resources (two HTTP servers, two ports, two file watchers)
- Splinters the workspace-paths source-of-truth (each server tracks its own copy)
- Makes refresh-on-workspace-change inconsistent (one refreshes, the other doesn't)
- Leaves several footguns on rapid configuration changes (no concurrency guard on `start()`, no event fired when the discovery-file write fails, an unnecessary per-`ping` repository lookup once both capabilities share `wrapHandler`)

## Goal

Replace the two servers with a single shared `IpcService` owned by the host. The CLI and agent capabilities register handlers and publish their own discovery files against this shared server. Discovery directories stay distinct so older `gk` binaries and peer GitLens windows continue reading what they expect, and the agents package owns the `workspacePaths` it advertises (no implicit dependency on `workspace.workspaceFolders` at the IPC layer).

ianhattendorf:
Closes #5198

## Summary

- Extracts the existing IPC server into a new `@gitlens/ipc` package and consolidates the CLI and agent capabilities onto a single shared HTTP server (one port, one bearer token), with each capability publishing to its own discovery directory so older `gk` binaries and peer GitLens windows continue to read what they expect
- Owns server lifecycle, handler registration, and discovery-file writes in a single `IpcService` (with a browser stub) that the CLI integration and agents provider both register against, instead of each spinning up a separate server
- Inverts the agents package's host dependency via an `IpcRegistrar` interface and threads `workspacePaths` through `publishAgents` so the agents package owns what its discovery file advertises
- Refreshes the CLI discovery file on workspace folder changes (previously only the agents file was refreshed)
- Gates `start()` on the CLI integration with `@gate()` so concurrent configuration changes can't race into duplicate handler registrations
- Fires `gk:cli:ipc:started` whenever the IPC server is up so MCP providers don't sit idle for the 30 s wait when the discovery-file write fails
- Skips the per-call repository lookup for `ping` so it stays cheap as a liveness check

## Test plan

- [x] `pnpm run build:packages` succeeds
- [x] `pnpm exec tsc --noEmit -p tsconfig.node.json` and `tsconfig.browser.json` are clean for changed code
- [x] Toggle `gitkraken.mcp.autoEnabled` rapidly — no `IPC handler '<name>' is already registered` errors and no unhandled rejections in the GitLens output panel
- [x] With write-protected `tmpdir/gitkraken/gitlens`, MCP providers don't sit idle for 30 s after activation; `gk:cli:ipc:started` still fires (with `discoveryFilePath: undefined`)
- [x] `gk ping` does NOT trigger a `getRepository` call; other CLI commands still do
- [x] On workspace folder change: only ONE write per change to the agents discovery file; the CLI discovery file still updates
- [x] Two GitLens windows can still discover and query each other's agent sessions via `tmpdir/gitkraken/gitlens/agents/`


ianhattendorf:
augment review

d13 (approved):

PR #5221, opened 2026-05-13
Fix Graph Details publish branch for branches without upstream

### LINKED ISSUE #5220: Graph Details: Push fails with 'Unable to push' for branches without upstream tracking
## Problem

When a branch has no upstream (or the upstream ref can't be resolved — common in linked worktrees), clicking **Push** or **Publish Branch** in the Graph Details panel's WIP section fails with:

```
Error: Unable to push
OperationsGitSubProvider.push — failed
```

The same push works correctly from the **Branches View** because it uses the Git Wizard which detects missing upstreams and offers to publish.

## Root Cause

In `gl-graph-details-panel.ts`, both `handlePublishBranch` and `handlePush` call the same `services.repository.push(repoPath)` method, which goes through `RepoActions.push(repoPath)` without passing a branch reference. The Git Wizard *does* detect `branch.upstream == null` and shows a publish UI (see `push.ts` line 213), but in linked worktrees, `revParse` fails to read `refs/remotes/origin/<branch>` from the per-worktree git directory (it looks in `.git/worktrees/<name>/refs/remotes/` instead of the shared `.git/refs/remotes/`), causing the branch to appear as having an upstream with `ahead: 0, behind: 0`.

The wizard then takes the normal push path (no `--set-upstream`) and `OperationsGitSubProvider.push()` throws because `remoteName` and `upstreamName` are both null.

Additionally, `handlePublishBranch` (triggered by the "Publish Branch" button in the empty state) is identical to `handlePush` — it doesn't pass publish-specific options, so even when the panel correctly detects an unpublished branch and shows "Publish", the handler still attempts a normal push.

## Reproduction

1. Open a linked worktree in the dev host (`.wt` folder)
2. Select the WIP row in the Graph
3. Click Push or Publish in the details panel
4. Observe "Unable to push" error

## Fix Approach

1. **Immediate fix**: Add a dedicated `publishBranch` RPC method that resolves the current branch and passes it as a reference to `RepoActions.push`, matching the pattern used by the graph webview's host-side `publishBranch` command. Update `handlePublishBranch` to use it.

2. **Root cause (separate)**: The worktree remote ref resolution bug in `revParse` that reads from per-worktree refs instead of the shared common dir.

## Key Files

- `src/webviews/apps/plus/graph/components/gl-graph-details-panel.ts` — handlers
- `src/webviews/rpc/services/repository.ts` — RPC service
- `packages/git-cli/src/providers/operations.ts` — push logic (throw site)
- `src/commands/git/push.ts` — wizard publish detection

## Verification steps

- [x] Open a linked worktree with no upstream in the dev host
- [x] Select the WIP row in the Graph
- [x] Click "Publish Branch" in the details panel — should show the publish wizard instead of throwing
- [x] Click "Publish" in the WIP empty-state pane — should also show the publish wizard
- [x] Verify regular Push still works for branches with an upstream (no regression)
- [x] After publishing, verify the graph updates the branch label to show upstream tracking without needing a manual refresh

sergeibbb:
## Summary

Fixes #5220

- Adds a dedicated `publishBranch(repoPath)` RPC method that resolves the current branch and passes it as a reference to `RepoActions.push`, matching the pattern used by the graph webview's host-side `publishBranch` command
- Updates `handlePublishBranch` in the Graph Details panel to call the new method instead of the generic `push`
- Expands the repository change notification after a successful publish to include `config` and `heads` (not just `remotes`), so the graph updates the branch label to reflect the new upstream tracking

Previously, the "Publish Branch" button in the Graph Details panel's WIP section called the same `push(repoPath)` as the regular Push button. This failed in linked worktrees where upstream ref resolution returns stale data, causing `OperationsGitSubProvider` to throw "Unable to push".

Additionally, after a successful publish, the graph showed the new remote branch but didn't update the local branch label because the operations provider only notified about `remotes` changes, missing the `config` and `heads` updates needed for the branch to reflect its new upstream.

### DIFF
--- a/packages/git-cli/src/providers/operations.ts
+++ b/packages/git-cli/src/providers/operations.ts
@@ -586,7 +586,10 @@ export class OperationsGitSubProvider implements GitOperationsSubProvider {
 			}
 
 			this.context.hooks?.cache?.onReset?.(repoPath, 'branches');
-			this.context.hooks?.repository?.onChanged?.(repoPath, ['remotes']);
+			this.context.hooks?.repository?.onChanged?.(
+				repoPath,
+				options?.publish != null ? ['config', 'heads', 'remotes'] : ['remotes'],
+			);
 		} catch (ex) {
 			scope?.error(ex);
 			throw ex;
--- a/src/webviews/apps/plus/graph/components/gl-graph-details-panel.ts
+++ b/src/webviews/apps/plus/graph/components/gl-graph-details-panel.ts
@@ -1187,7 +1187,7 @@ export class GlGraphDetailsPanel extends SignalWatcher(LitElement) {
 
 	private handleCreateBranch = () => this._actions.createBranch(this.effectiveRepoPath);
 
-	private handlePublishBranch = () => void this._actions.services.repository.push(this.effectiveRepoPath!);
+	private handlePublishBranch = () => void this._actions.services.repository.publishBranch(this.effectiveRepoPath!);
 
 	private handlePull = () => void this._actions.services.repository.pull(this.effectiveRepoPath!);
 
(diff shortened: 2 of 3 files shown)


eamodio (approved):
LGTM

PR #5262, opened 2026-05-20
Replace agent-session quickpick with a smart dispatcher

ianhattendorf:
## Summary

Replaces the action-pick quickpick that fired on "open agent session" with a deterministic dispatcher that picks the right action based on session host + workspace state:

- **In-workspace + extension-hosted** → open in the Claude Code VS Code extension
- **In-workspace + CLI-hosted** → focus the owning terminal window via `pid`
- **Out-of-workspace** → ask the peer GitLens window that already has the workspace open to pre-open the session in its Claude Code extension (new `agents/sessions/open` IPC route), then `vscode.openFolder` to focus that peer's window

Host classification reads `~/.claude/sessions/<pid>.json` for the `entrypoint` field, with a stale-file guard (file's `pid` field must match the filename) and a sensible "extension if installed, else CLI" fallback when the metadata file is missing.

The peer-notify call is capped at 500 ms so an unhealthy peer can't stall the click — the POST keeps running in the background after `vscode.openFolder` lands.

## Test plan
- [x] Unit tests: `pnpm run test:packages` (4 new `agents/sessions/open` IPC handler cases, 3 new `notifyPeerOpenSession` cases) and `pnpm run test` (7 new `classifyClaudeSessionHost` cases) all pass
- [x] In-workspace + extension-hosted session: clicking "open agent session" routes to the Claude Code extension editor pane
- [x] In-workspace + CLI-hosted session: clicking routes to the terminal window (with parent-PID walk for tmux-style setups)
- [x] Out-of-workspace session with a peer window holding the workspace open: peer's Claude Code extension receives the session before the focus switch lands
- [x] Out-of-workspace session with no peer window: `vscode.openFolder` falls through and the current window switches to the folder
- [x] Extension-hosted session whose extension command throws: surfaces the "Unable to open agent session" warning rather than focusing the VS Code window in a misleading no-op
- [x] Unhealthy peer (paused/unresponsive but not RST): click resolves within ~500 ms instead of waiting the full per-fetch timeout

### DIFF
--- a/src/agents/agentStatusService.ts
+++ b/src/agents/agentStatusService.ts
@@ -8,7 +8,7 @@ import { registerCommand } from '../system/-webview/command.js';
 import type { AgentSessionState, AgentSessionWorktreeMetadata } from './models/agentSessionState.js';
 import { getSessionDisplayName, serializeAgentSession } from './models/agentSessionState.js';
 import type { AgentSession, AgentSessionProvider, PermissionDecision, PermissionSuggestion } from './provider.js';
-import { isClaudeExtensionAvailable } from './utils/-webview/claudeExtension.js';
+import { isClaudeExtensionAvailable, tryOpenClaudeSession } from './utils/-webview/claudeExtension.js';
 
 export class AgentStatusService implements Disposable {
 	private readonly _onDidChange = new EventEmitter<void>();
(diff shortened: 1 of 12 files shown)


src/agents/agentStatusService.ts
```diff
@@ -442,78 +443,94 @@ export class AgentStatusService implements Disposable {
...
+			await commands.executeCommand('claude-vscode.editor.open', sessionId);
+			return true;
+		} catch {
+			try {
+				await commands.executeCommand('claude-vscode.sidebar.open');
+				return true;
+			} catch {
+				return false;
+			}
 		}
```
d13: We should probably add some logging here.

src/agents/agentStatusService.ts
```diff
@@ -442,78 +443,94 @@ export class AgentStatusService implements Disposable {
...
 
-		let action: string;
-		if (actions.length === 1) {
-			action = actions[0].action;
-		} else {
-			const actionPick = await window.showQuickPick(actions, {
-				placeHolder: `Action for ${getSessionDisplayName(session, this.getWorktreeMetadataForSession(session)?.name)}`,
-			});
-			if (actionPick == null) return;
+		void window.showWarningMessage('Unable to open agent session.');
```
d13: We should probably include Logging here too

d13 (approved):

PR #5305, opened 2026-06-02
Fix push to use upstream tracking branch when names differ

### LINKED ISSUE #5304: Graph push buttons silently fail when branch upstream tracks a differently-named remote branch
## Problem

When a local branch tracks a remote branch with a **different name** (e.g., local `feature/foo` has upstream `origin/main`), clicking any of the three "Push" buttons in the Commit Graph does nothing — the branch is not pushed to the remote, and no error is shown.

### Steps to reproduce

1. Create a new branch based on a remote branch so that git sets up tracking to the remote branch:
   ```
   git checkout -b feature/my-feature origin/main
   ```
   This creates `feature/my-feature` with the git config:
   ```
   [branch "feature/my-feature"]
       remote = origin
       merge = refs/heads/main
   ```
2. Make a commit so the working directory is clean.
3. Open the Commit Graph. The branch shows a "Push 1↑" indicator.
4. Click any of the three push buttons (header toolbar, commit details panel header on WIP row, or "Next Actions" section).
5. A quickpick appears offering **"Push"** and **"Force Push"**.
6. Select "Push".

**Actual result:** Nothing visibly happens — the "Push 1↑" indicator remains, and no remote branch appears to be created. (In reality, `git push origin feature/my-feature` does run and creates `origin/feature/my-feature`, but the upstream tracking is not updated from `origin/main` to `origin/feature/my-feature`, so the Graph still shows the branch as having unpushed commits.)

**Expected result:**
Act as VSCode: push to the current upstream (in this example `main`). That sounds correct behavior because:
- the interface shows "ahead-behind" relative to the current upstream
- it decides that the "push" button should because it sees the active upstream in configs.
- VSCode's Source Control panel works exactly as described.

Therefore, pushing to that exact remote branch instead of creating a new one looks correct.

### Root cause

In `src/commands/git/push.ts`, the `confirmStep()` method's non-branch-reference path (used when the Graph calls `RepoActions.push(repoPath)`) decides between the "Publish" and "Push" flows based solely on whether there are commits ahead of the upstream:

```typescript
if (status?.upstream?.state.ahead === 0) {
    // Publish flow (only entered when upstream is null or 0 ahead)
    ...
} else {
    // Push flow — shows Push / Force Push quickpick with empty flags []
    ...
}
```

When the branch tracks `origin/main` and has commits ahead, the code enters the `else` block and shows "Push / Force Push". Selecting "Push" calls `execute()` with empty flags `[]`, which calls `pushAll()` **without** the `publish` option. The operations layer runs `git push origin feature/my-feature` which creates the remote branch, but since `--set-upstream` was not used, the local tracking remains pointed at `origin/main`. The Graph still sees the branch as ahead of `origin/main`, making it appear as if nothing happened.

The condition does not account for the case where the upstream branch name differs from the local branch name — it should detect that `getBranchNameWithoutRemote(status.upstream.name)` (e.g., `main`) does not match `status.branch` (e.g., `feature/my-feature`) and route to the Publish flow instead.

### Affected code

- **File:** `src/commands/git/push.ts`
- **Method:** `PushGitCommand.confirmStep()`, line ~328 (the `status?.upstream?.state.ahead === 0` condition)
- **Execution path:** Graph push button → `RepoActions.push(repoPath)` → `executeGitCommand({ command: 'push' })` → `PushGitCommand.confirmStep()` (non-branch-reference path, since `state.reference` is undefined)

### Additional notes

- The same issue likely affects the `isBranchReference` path (line ~203) — when a branch reference is passed with an upstream whose name doesn't match, the code checks `branch?.upstream?.state.ahead` and falls into the regular push flow instead of publish. However, that path is less commonly hit from the Graph.
- When the upstream is completely `null` (no tracking at all), the Graph doesn't show push buttons at all, so the `upstream == null` case is not the scenario users encounter.

sergeibbb:
Fixes #5304

## Summary

image

- When a branch tracks a differently-named upstream (e.g., `feature/foo` → `origin/main`), the push operation now constructs a refspec `feature/foo:main` so git pushes to the correct remote branch
- Previously, `git push origin feature/foo` was executed, which created a new `origin/feature/foo` branch instead of pushing to `origin/main` — leaving the upstream tracking unchanged and the Graph still showing unpushed commits

## Test plan

- [ ] Create a branch tracking a differently-named remote: `git checkout -b feature/test origin/master`
- [ ] Commit a change
- [ ] Open the Commit Graph and click any of the 3 push buttons (header toolbar, WIP details header, or "Next Actions" panel)
- [ ] Verify push goes to `origin/master`, not `origin/feature/test`
- [ ] Verify the "Push N↑" indicator disappears after push
- [ ] Regression: push a branch with matching upstream name (e.g., `main` → `origin/main`) — should work as before

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 
 ### Fixed
 
+- Fixes push from the _Commit Graph_ silently pushing to a wrong remote branch when the local branch tracks a differently-named upstream (e.g., `feature/foo` tracking `origin/main`) &mdash; now correctly pushes to the configured upstream branch instead of creating a new remote branch ([#5304](https://github.com/gitkraken/vscode-gitlens/issues/5304))
 - Fixes the working changes (WIP) _Generate Commit Message_ in the _Commit Graph_ losing its result when the selection changes while a message is generating &mdash; the generation now continues and the message lands in the originating worktree's commit input (or its saved draft if you've navigated away) ([#5295](https://github.com/gitkraken/vscode-gitlens/issues/5295))
 - Fixes the commit details metadata bar in the _Commit Graph_ and _Inspect_ view not surfacing a reachable tag when a commit is contained only in tags (and not any branch) &mdash; now shows the most relevant tag with a tag icon and color, matching how branches are shown ([#5293](https://github.com/gitkraken/vscode-gitlens/issues/5293))
 - Fixes incorrect GitLab avatars being shown when multiple GitLab users share the same name ([#2205](https://github.com/gitkraken/vscode-gitlens/issues/2205))
--- a/packages/git-cli/src/providers/__tests__/integration/operations.test.ts
+++ b/packages/git-cli/src/providers/__tests__/integration/operations.test.ts
@@ -1,11 +1,12 @@
 import * as assert from 'assert';
 import { execFileSync } from 'node:child_process';
-import { writeFileSync } from 'node:fs';
+import { mkdtempSync, rmSync, writeFileSync } from 'node:fs';
+import { tmpdir } from 'node:os';
 import { join } from 'node:path';
 import type { SigningErrorReason } from '@gitlens/git/errors.js';
 import { CommitError, MergeError, SigningError } from '@gitlens/git/errors.js';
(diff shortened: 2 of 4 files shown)


sergeibbb:
augment review

eamodio (approved):

PR #5324, opened 2026-06-05
Extracts integrations into the @gitlens/integrations workspace package

eamodio:
This needs a thorough review/testing. Agents have gone through and come up clean, but it needs human attention — especially at the boundary layers (which I personally haven't look at after the latest round of significant changes as I ran out of time before I leave for vacation).

Extracts src/plus/integrations/ into a cloud-only @gitlens/integrations workspace package consumed through a single runtime port (IntegrationServiceContext: storage/account/config/http/cache/repositories + optional hooks) and composed via createIntegrationManager(ctx). The package is vscode-free; the host wires its services through an adapter under src/plus/integrations/host/.

- Unifies the former subscription/cloudConnection/authentication/uris ports into one account port and relocates the GK-Dev connect/manage OAuth round-trips host-side (gkDevFlows).
- Moves the per-provider API clients into the package; constructs cloud auth providers in-package while keeping createAuthenticationProvider as an optional consumer extension point.
- De-leaks the host gitlens. secret-key prefix (host re-prepends; stored keys stay byte-identical).
- Keeps the self-managed Pro paywall on the GitHub/GitLab Enterprise/self-hosted host connect sites only (Bitbucket Server / Azure DevOps Server stay free).
- Groups outbound hooks by domain; makes isIntegrationsEnabled an optional, default-enabled config port.


sergeibbb:
Hi @eamodio 

I'm reviewing it. I see this note: _"External consumers should import from this entry point only (index.js), not from the package's internal subpaths."_ and at the same time there are a lot of imports from subpaths are still there. As I understand it's left as a follow-up, isn't it?

sergeibbb:
**Pervasive `this.authenticationService.ctx.*` transitive access** in providers (Jira, Azure, Bitbucket, BitbucketServer, GitLab) and in `integrationService.ts:463`. All providers reach for `ctx` through the auth service instead of receiving it directly. This works because it's the same `ctx` object, but it's an architectural smell -- the auth service's `ctx` is readonly but exposed as a public field, creating a coupling that will resist future refactoring (e.g., giving providers their own narrower context). Not blocking, but worth noting for future cleanup.

**Why?**

Here's the picture. There are ~50 call sites doing `this.authenticationService.ctx.<something>`. They break down into a few categories:

**What providers actually need from `ctx`:**
- **`ctx.storage`** — Jira, Bitbucket, BitbucketServer, AzureDevOps cache provider-specific data (orgs, projects, accounts)
- **`ctx.cache`** — `gitHostIntegration.ts` uses the cache for PRs, issues, metadata, accounts
- **`ctx.hooks`** — GitLab calls `onReauthenticationRequired`, AzureDevOps calls `ui.onError`
- **`ctx.repositories`** — Bitbucket scans open remotes for PR search
- **`ctx.config`** — `IntegrationBase.ensureSession` checks `isIntegrationsEnabled`
- **`ctx.account`** — `IntegrationBase.access()` calls `isTrialOrPaid()`, auth provider checks `getAccount()`

**What `IntegrationBase` itself needs** (the base class in `models/integration.ts`) — this is the bulk: `storage` (workspace connected-key flags), `hooks` (session events, disconnect confirmation, failure notifications), `config` (integrations enabled check), `account` (access gating).

The chain is: providers receive `authenticationService` in their constructor (via `IntegrationBase`), and `authenticationService.ctx` is a public `readonly` field. So every provider reaches `ctx` by hopping through the auth service, even though the auth service has nothing to do with storage or caching — it's just the object that happens to carry the reference.

**Why this happened**: `IntegrationBase`'s constructor takes `(authenticationService, getProvidersApi, didChangeConnection)`. Before the extraction, these classes reached `Container` directly. The extraction replaced `Container` with `ctx`, but threaded it through `authenticationService` rather than giving it to providers directly — probably to minimize constructor signature changes.

**Is it a real problem?** For this PR, no. It's the same `ctx` object everywhere, so behavior is correct. The concern is about future refactoring: if you ever want to give providers a narrower context (e.g., Jira only needs `storage` + `hooks`, not `cache` + `repositories`), you'd have to change every provider's constructor. But that's a deliberate narrowing that would happen regardless of how `ctx` is passed today.

In short: it's a cosmetic coupling, not a behavioral one. The "smell" is real but mild — it's a reasonable trade-off for a clean extraction that doesn't inflate every provider constructor.

sergeibbb:
Went through the main integration flows in the debug host: connected a cloud integration via OAuth, disconnected and reconnected it, opened Launchpad to verify issues/PRs load, checked the Settings webview autolinks section — all working as expected, no regressions.

packages/plus/integrations/src/providers/utils.ts
```diff
@@ -35,16 +55,16 @@ export function getEntityIdentifierInput(entity: Issue | PullRequest | Launchpad
 
 	let provider = fromStringToEntityIdentifierProviderType(entity.provider.id);
 	let domain = null;
-	if (provider === EntityIdentifierProviderType.Github && !isGitHubDotCom(entity.provider.domain)) {
+	if (provider === EntityIdentifierProviderType.Github && !isGitHubDotCom(entity.provider.domain!)) {
```
sergeibbb: One more question here, that is flagged by the AI analyzed. Should we fix the force-cast to non-null here?

image

It can be simply done by relaxing `isGitHubDotCom` argument type:

image

The `equalsIgnoreCase` accepts null values and the behavior does not change.

If by some reason we do not want `isGitHubDotCom` to accept nulls, then maybe do this?

image

Btw. the AI suggests the opposite: skip entering the `if` block when the domain is null, because the `!`-marking looks like the assumption that the legal state then it's not null. That highlight the ambiguity.

So I think that by removing "!" we make the decision of what the `null|undefined` value should mean here:
1. we just accept it and let the API provider handle it later
2 we think that the state is illegal, then even if we assume it never happens, I think we should catch that state in code and fail early: just to nail the expectation.

I prefer #1 just let `isGitHubDotCom` to accept nulls:
- it does not change the current behaviour
- highlight that we are tolerant to nulls here and ready to accept them.

sergeibbb (approved):

PR #5368, opened 2026-06-19
Fixes silent error handling in Graph commit message generation

sergeibbb:
## Summary

- Shows an error notification when the Graph details "Generate Commit Message" fails with an untyped provider error (e.g. Copilot credit limit)
- Adds `Provider Error` and `Provider Unavailable` AI simulator modes for reproducing untyped provider failures in debug builds

Closes gitkraken/vscode-gitlens-private#89

## Approach

**Implemented (Option A — targeted fix):** Added `showGenericErrorMessage()` to the Graph's `generateCommitMessage` catch block in `graphInspectServices.ts`, matching the existing SCM command behavior in `generateCommitMessage.ts`.

> Note: originally implemented in `graphWebview.ts`; the handler has since moved to `graphInspectServices.ts` (Home view removal relocated the Graph details services), so the fix now lands there.

**Alternative considered (Option B — systemic fix):** Add a fallback notification in `sendRequest()` (`aiProviderService.ts`) for all unclassified errors before rethrowing, so no caller has to handle it individually. This would cover all current and future AI surfaces (Graph, SCM, Composer, Review, etc.) but is a broader change in a critical path.

## Test plan

Verified live against a debug build (subscription simulator → Pro, AI simulator → the modes below):

- [x] Provider Error mode: Graph → select WIP → sparkle "Generate Commit Message" → error notification shown (previously: silent, error only in the output channel)
- [x] SCM "Generate Commit Message" with the same mode → same error notification (surfaces now match)
- [x] Cancel mode → no spurious error notification (cancellation is returned as `'cancelled'`, never rethrown)
- [x] Provider Unavailable mode → silent on both surfaces, no crash (expected — no error to report; the "AI is degraded" indicator is out of scope, tracked by the issue's remaining items)
- [x] Typed AIError modes (Error, Quota) still show their specific notifications from both surfaces — and only once on the Graph (typed errors are notified inside `sendRequest` and never reach the new catch, so no double notification)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 
 ### Fixed
 
+- Fixes the _Commit Graph_ commit details _Generate Commit Message_ silently doing nothing when the AI provider fails with an untyped error (e.g. Copilot credit limit) &mdash; the failure is now surfaced as an error notification, matching the _Source Control_ panel
 - Fixes the _Commit Graph_ retaining outdated parent connections and file statistics after deepening a shallow clone or changing Git replacement refs
 - Fixes closing _Compose_ or _Review_ after restarting discarding the saved work, and restores _Resume_ when returning after navigating away
 - Fixes contextual tips already seen in one _Commit Graph_ view showing again after another view saves its seen tips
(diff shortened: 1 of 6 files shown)


sergeibbb:
augment review

PR #5494, opened 2026-07-10
Adds an SSH Allowed Signers editor

### LINKED ISSUE #5469: Build an SSH allowed_signers file to verify SSH-signed commits
### Problem

GitLens already parses commit signatures and renders status badges, but **SSH-signed commits can never reach the "Signed & Verified / Trusted" state**. They stay at *"Unverified Signer"* or *"Missing Key"* because Git can only validate an SSH signature when `gpg.ssh.allowedSignersFile` points at an `allowed_signers` file that lists the signer's email **and full public key**. Most users never maintain that file by hand — and `git log` only exposes fingerprints (`SHA256:…`), not the full key needed for an entry.

### Proposal

Add an **Edit SSH Allowed Signers** editor that builds/maintains the `allowed_signers` file for you:

- **Discovers candidate signers** from two sources and cross-checks them for provenance:
  - **Local commits** — the full public key is embedded in each SSH-signed commit's SSHSIG blob, so it can be extracted offline for any host.
  - **Provider APIs** — when a GitHub or GitLab integration is connected, each signer's registered SSH signing keys are fetched and matched, marking keys as *signed-here*, *provider-verified*, or *both*.
- **Lets you review and choose** which signers to include (avatar, email, fingerprint, provenance, signed-commit count).
- **Writes/merges** the file without clobbering existing manual entries (idempotent), and optionally points `gpg.ssh.allowedSignersFile` at it (globally or for the current repository).

### Entry points

- Command Palette: **GitLens: Edit SSH Allowed Signers**
- An **Add to allowed signers…** action shown on an unverified SSH signature in a commit's signature details (_Inspect_ view / commit hover), scoped to that commit's repository.

### Outcome

Once the file is generated, previously *"Unverified Signer / Missing Key"* SSH-signed commits flip to **"Signed & Verified / Trusted"** in GitLens.

ianhattendorf:
# Description

Closes #5469.

SSH-signed commits can never reach the **"Signed & Verified / Trusted"** state in GitLens because Git can only validate an SSH signature when `gpg.ssh.allowedSignersFile` lists the signer's email **and full public key** — and `git log` only exposes fingerprints. This PR adds an **Edit SSH Allowed Signers** editor that builds/maintains that file for you.

## What it does

- **Discovers candidate signers** from two sources, cross-checked for provenance:
  - **Local commits** — extracts the full public key embedded in each SSH-signed commit's SSHSIG blob (offline, any host), read in a single batched `git cat-file --batch`.
  - **Provider APIs** — when a GitHub or GitLab integration is connected, fetches each signer's registered SSH signing keys and matches them, marking keys as *signed-here*, *provider-verified*, or *both*.
- **Streams a progress-aware loading page** while scanning, then lets you review signers (avatar, email, fingerprint, provenance, signed-commit count) and choose which to include.
- **Pre-selection**: only API-verified signers are pre-checked by default; opening from a commit's *Add to allowed signers…* action pre-checks all (explicit intent to trust).
- **Writes/merges** the file idempotently without clobbering existing manual entries, and optionally points `gpg.ssh.allowedSignersFile` at it (globally or per-repo).

image
image

## Entry points

- Command Palette: **GitLens: Edit SSH Allowed Signers**
- An **Add to allowed signers…** action on an unverified SSH signature in a commit's signature details (_Inspect_ view / commit hover), scoped to that commit's repository.

## Verification

- Unit tests for the SSHSIG parser and `allowed_signers` merge logic.
- **Exercised end-to-end in a running instance** (command palette open → discovery → signer list; provenance-based default pre-checking; `preselectAll` via the signature-action link; toggle; Save → file write + idempotent 2-entry merge with no clobber; signature-action render + hover-underline).


ianhattendorf:
augment review

Issue #5391, opened 2026-06-24
Explore consolidating views


d13:

trevor-polidore:
Turning the Graph into the GitLens panel:
- [x] Place the avatar/header bar from home on top of the graph #5411 
- [x] Ask on entry the preferred user layout (left panel, or bottom) #5412 
- [x] When in vertical layout:
    - [x] https://github.com/gitkraken/vscode-gitlens/issues/5402
    - [x] https://github.com/gitkraken/vscode-gitlens/issues/5403
    - [x] Adjust columns in graph appropriately (e.g. use compact graph column, hide SHA column) (implemented in the new graph engine)
- #5447 
  - [x] Default to unpined sidebar
  - [x] Replace ‘Pro’ badge with start work
- [x] Adjust graph gate design to fit in smaller area
- [x] https://github.com/gitkraken/vscode-gitlens/issues/5401
- [x] ~~Hide gitlens ‘Home’ icon in graph header bar~~

sergeibbb:
@trevor-polidore 

By Graph replacing the Home as a main view do we mean that some day the following happens:

1. the current Home is removed completely
2. Graph gets its new default place in the Side Bar view on the left

What of above is true?

cc @d13 @eamodio

nikolay-1986:
## Verification: Passed

Rollup for this epic. All nine sub-issues are now verified: seven already carried `verified ✔`, and the
remaining two were verified in this run against `main` @ `0b3834c56`, dev build, VS Code 1.132.0, on a
real signed-in account.

### Covered

| Sub-issue | Status |
| --- | --- |
| #5411 Home view avatar/header bar at the top of the Graph | previously verified |
| #5412 Ask users on entry which layout they prefer | previously verified |
| #5447 Graph sidebar unpinned by default, Pro badge → Start Work | previously verified |
| #5449 Inline mode for the Graph header bar | previously verified |
| **#5500 Narrow layout: compact graph column, hidden SHA** | **verified in this run — Passed** |
| #5513 Graph header account/integrations state after initial load | previously verified |
| #5519 Graph gate design in a smaller area | previously verified |
| #5545 Graph as the main view, default to sidebar | previously verified |
| **#5610 Account/org switcher usable when the Graph is gated** | **verified in this run — Passed** |

**#5500** — the behavior ships via the new Graph engine, not via the draft PR that was closed unmerged
(none of that PR is in `main`). Below a graph width of 520px the whole graph switches to its stacked
[...]

PR #5501, opened 2026-07-14
Adjusts Commit Graph columns for narrow layouts

### LINKED ISSUE #5500: Graph vertical layout: use compact graph column and hide SHA column
Sub-issue of #5391 — _"When in vertical layout: Adjust columns in graph appropriately (e.g. use compact graph column, hide SHA column)"_.

## Goal

When the Commit Graph view is **horizontally narrow**, optimize the columns for the reduced space:

- switch the graph column to `compact` mode, and
- hide the SHA column.
- hide Changes column.

The user's own wide-view column preferences must be preserved — these are view-driven overrides, not settings changes.

## Requirements / behavior

- **The trigger is the width of the graph view, not the details-panel placement.** Narrowness is keyed off the width-driven signal (`_autoEffectiveLocation`, the same overall graph-view width breakpoint + hysteresis used for the `auto` details-location decision), which is tracked regardless of where the details panel is pinned. So:
  - a genuinely narrow view gets compact columns + hidden SHA **whatever** the details pin, and
  - an explicit details-on-**bottom** pin on a **wide** editor is **not** treated as narrow — columns stay full.
- In a wide view, columns render from the user's stored settings unchanged.
- The overrides are **not persisted**: when the user resizes/reorders a column while narrow, the compact/hidden overrides are stripped before the column config is saved, so wide-view preferences survive.
- Width changes re-apply/remove the overrides reactively (no reload).

## Column menus while narrow

The managed toggles (Show/Hide SHA, Show/Hide Changes, Compact/Expanded Graph column) are **disabled** in the column header / settings context menus while the Graph is narrow, and the menu reflects the effective (overridden) state rather than the stored one. A permanently-disabled info line — _"Some Options Return When the Graph Is Wider"_ — appears at the top of the columns group to explain why (VS Code menus don't support tooltips, so a disabled info item is the conventional substitute). Unmanaged column toggles (author, date, message, ref) stay fully functional. Implemented via a `columns:narrow` context token appended webview-side + `enablement` clauses on the managed commands.

## Possible future refinement

The width signal is the **overall** graph-view width. When the details panel is pinned to the **right**, it eats horizontal space, so the graph *pane* can be cramped while the overall width is still above the breakpoint (→ no compaction). If that becomes a problem, drive off the graph pane's own measured width instead of the overall view width.

## Implementation

Branch `feature/5391-graph-columns-adjustment`:
- `graph-wrapper.ts` — `effectiveColumns` getter (compact graph + hidden SHA/Changes when narrow, identity-cached), strip-on-persist in `onColumnsChanged`, and `effectiveGraphContext` (rewrites the column-menu context tokens to effective state + appends `columns:narrow`); takes a width-driven `narrow` prop.
- `graph-app.ts` — `isGraphViewNarrow` (width-based, decoupled from the details pin) passed to the wrapper.
- `contributions.json` — `enablement` guards on the six managed column commands + the disabled `gitlens.graph.columnsNarrowInfo` info line.

sergeibbb:
## Summary

Adjusts the Commit Graph's columns when the graph view is horizontally narrow: uses a compact graph column and hides the SHA and Changes columns so the limited horizontal space is used effectively. Closes #5500 (sub-issue of #5391).

The trigger is the graph's **actual width**, not where the details panel is pinned — so an explicit details-on-bottom pin on a wide editor is *not* treated as narrow, and a genuinely narrow graph gets the treatment regardless of the pin. Reuses the same width breakpoint (with hysteresis) already used for the `auto` details-location decision.

## Behavior

- **Narrow**: graph column → `compact` mode, SHA and Changes columns hidden.
- **Wide**: columns render from the user's stored settings, unchanged.
- **Not persisted**: resizing/reordering a column while narrow strips the compact/hidden overrides before saving, so the user's wide-view preferences survive.
- **Reactive**: width changes apply/remove the overrides immediately, no reload.
- **Column menus while narrow**: the managed toggles (Show/Hide SHA, Show/Hide Changes, Compact/Expanded Graph Column) are disabled in the column header / settings-gear context menus, and reflect the *effective* (overridden) state rather than the stored one. A disabled info line — _"Some Options Return When the Graph Is Wider"_ — explains why (VS Code menus don't support tooltips, so a disabled info item is the conventional substitute). Unmanaged toggles (author, date, message, ref) stay fully functional.

## Implementation

- `src/webviews/apps/plus/graph/graph-app.ts` — `isGraphViewNarrow` getter, keyed off the width-driven `_autoEffectiveLocation` signal (decoupled from the details-panel pin); passed to the wrapper.
- `src/webviews/apps/plus/graph/graph-wrapper/graph-wrapper.ts`:
  - `effectiveColumns` — applies the narrow overrides, identity-cached to avoid prop churn on unrelated renders.
  - `onColumnsChanged` — strips the overrides before persisting.
  - `effectiveGraphContext` — rewrites the column-menu context tokens to the effective state and appends a `columns:narrow` token; identity-cached (the React engine ingests `context` by reference).
- `contributions.json` — `enablement` guards on the six managed column commands (keyed off `columns:narrow`) + a new disabled `gitlens.graph.columnsNarrowInfo` info-line command.
- `package.json` / `src/constants.commands.generated.ts` — regenerated from `contributions.json`.
- `CHANGELOG.md` — entry under **Changed**.

## Test plan

Verified live via the vscode-inspector MCP, both engines (React and Lit):
- [x] Narrow graph view (792px): graph column compact, SHA + Changes headers hidden, columns not persisted after resize, `data-vscode-context` carries `columns:narrow` + rewritten tokens.
- [x] Wide graph page (1152px+): no overrides, no `columns:narrow` token, stored settings unaffected.
- [x] Switching width reactively applies/removes overrides without a reload.
- [x] `pnpm run build` (type-check + unit-test compile) passes clean.
- [ ] Manual: right-click the column header while narrow — confirm the managed items render grayed out and the info line appears (native VS Code context menu, not verifiable via DOM automation).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 
 - Changes the default _Commit Graph_ lane colors for dark and high-contrast themes to a new vibrant, perceptually-uniform palette &mdash; every lane shares the same perceived brightness so no lane visually dominates; light themes keep the previous colors, and any `gitlens.graphLane1Color`&ndash;`gitlens.graphLane10Color` customizations in `workbench.colorCustomizations` are still honored
 - Changes the _Commit Graph_ sidebar to be unpinned by default &mdash; the sidebar now floats over the graph and auto-collapses when it loses focus; pin it (or set `gitlens.graph.sidebar.pinned` to `true`) to restore the previous shared-space layout ([#5447](https://github.com/gitkraken/vscode-gitlens/issues/5447))
+- Changes the _Commit Graph_ to automatically use a compact graph column and hide the SHA and Changes columns when the graph is narrow &mdash; based on the graph's actual width rather than where the details panel is pinned, so your column layout for a wide graph is preserved and restored once it's wide again; while narrow, the corresponding column menu items are disabled with a note that they return when the graph is wider ([#5500](https://github.com/gitkraken/vscode-gitlens/issues/5500), [#5391](https://github.com/gitkraken/vscode-gitlens/issues/5391))
 
 ### Removed
 
--- a/contributions.json
+++ b/contributions.json
@@ -3016,7 +3016,7 @@
 		},
 		"gitlens.graph.columnChangesOff": {
 			"label": "Hide Changes Column",
-			"enablement": "webviewItemValue =~ /\\bcolumns:canHide\\b/",
+			"enablement": "webviewItemValue =~ /\\bcolumns:canHide\\b/ && !(webviewItemValue =~ /\\bcolumns:narrow\\b/)",
 			"menus": {
 				"webview/context": [
 					{
@@ -3029,6 +3029,7 @@
 		},
 		"gitlens.graph.columnChangesOn": {
 			"label": "Show Changes Column",
+			"enablement": "!(webviewItemValue =~ /\\bcolumns:narrow\\b/)",
 			"menus": {
 				"webview/context": [
 					{
@@ -3066,6 +3067,7 @@
 		},
 		"gitlens.graph.columnGraphCompact": {
 			"label": "Use Compact Graph Column",
+			"enablement": "!(webviewItemValue =~ /\\bcolumns:narrow\\b/)",
 			"menus": {
 				"webview/context": [
 					{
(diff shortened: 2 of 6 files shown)


sergeibbb:
Closing since it's already implemented in the new Graph engine

PR #5503, opened 2026-07-14
Fixes self-managed cloud token requests using the domain as a token id (#5497)

### LINKED ISSUE #5497: Self-managed cloud connections request `v1/provider-tokens/tokens/<domain>` instead of `/tokens/<uuid>`
### Summary
For a self-managed **cloud** integration (e.g. GitLab Self-Managed), GitLens repeatedly calls
`GET /v1/provider-tokens/tokens/<host-domain>` — passing the **host domain** where the backend's
`/v1/provider-tokens/tokens/:id` endpoint expects a **UUID token id**. The backend rejects it (originally 5xx
"invalid uuid" from the DB, now 4xx) and GitLens keeps re-issuing it. Surfaced via backend monitoring (DataDog).

The same root cause has a second, more destructive symptom: a **forced re-sync can delete the working token and leave
the connection permanently broken** (details in "Symptom B" below).

Introduced by #5430 (`aead3e398`, "Adds multiple simultaneous connections per provider integration").

### Severity
P1 — repeated failing backend requests per session refresh; token fetch for the connection never succeeds, and a forced
re-sync can permanently break an otherwise-working self-managed connection.

### Root cause (shared)
`ConfiguredIntegrationDescriptor.id` is, for legacy/migrated single connections, the **domain** (self-managed) — this is
intentional for secret-key stability (see `packages/plus/integrations/src/authentication/models.ts:84-91`).

`getConfiguredConnectionId` (`packages/plus/integrations/src/authentication/configuredIntegrationService.ts:490-494`)
returns that `id`, and its doc promises the result is safe to use "as a real per-connection token id" — but it never
checks, so for a legacy self-managed connection it returns the domain. `getCloudSession`
(`.../integrationAuthenticationProvider.ts:278-287`) then feeds that domain into `getConnectionSession(...)` as
`connectionId`, which takes the `/tokens/{connectionId}` branch (`.../cloudIntegrationService.ts:131-133`) →
`v1/provider-tokens/tokens/{domain}`.

For a legacy single self-managed connection the correct request is the provider-scoped `v1/provider-tokens/{provider}`
(the `connectionId == null` branch).

---

### Symptom A — repeated `/tokens/<domain>` requests
The cycle is self-sustaining: when the backend returns a token with no `tokenId`, `getCloudSession` stores the session
with `id = resolveConnectionId(...) = domain` (`.../integrationAuthenticationProvider.ts:324`), so the next fetch
re-reads a domain-keyed descriptor and repeats the bad request.

**Reproduction**
1. Connect a self-managed cloud provider whose backend token has no `tokenId` (legacy / mid-rollout).
2. Trigger any session get/refresh (open Launchpad, a PR/issue read, periodic sync).
3. Observe repeated `GET /v1/provider-tokens/tokens/<domain>` → 4xx/5xx.

---

### Symptom B — forced re-sync can permanently break the connection
`deleteSession` was narrowed by #5430 to delete only the cloud secret and **keep** the configured descriptor
(`preserveConfigured`) — `.../integrationAuthenticationProvider.ts:79-98`. Its sole caller is the forced-resync path
`syncCloudConnection('connected', forceSync=true)` (`packages/plus/integrations/src/models/integration.ts:352-374`),
which deletes then immediately refetches via `ensureSession({ sync:true })` → `getCloudSession`.

- Before #5430: `deleteStoredSessions(id, descriptor, undefined)` deleted **both** secrets **and** removed the descriptor
  → clean disconnected state on failure.
- After #5430: only the cloud secret is deleted and the descriptor is kept.

If the refetch fails, the descriptor survives with no backing secret: the provider is still reported connected, reads
silently return nothing, and no reconnect is prompted.

- **Transient (any cloud provider):** GK API 5xx / offline during refetch → shown connected, reads empty until the next
  successful sync (self-heals).
- **Permanent (self-managed legacy):** the working domain-keyed cloud secret is deleted, the refetch hits
  `/tokens/{domain}` (Symptom A) and fails, and reconcile cannot restore an id-less legacy connection
  (`packages/plus/integrations/src/integrationService.ts:1057-1058` filters out connections with no backend id and
  early-returns). Net: a forced sync **permanently breaks a working connection** until manual disconnect/reconnect.

---

### Proposed fix
1. **Root cause (fixes Symptom A and the permanent case of Symptom B):** make `getConfiguredConnectionId` honor its
   contract — return the id only when it is a real token id (not the domain), else `undefined` so the caller falls
   through to the provider-scoped primary endpoint:

   ```ts
   const connection = candidates?.find(c => c.primary) ?? candidates?.[0];
   return connection != null && connection.id !== connection.domain ? connection.id : undefined;
   ```

   Valid because legacy self-managed descriptors have `id === domain`, while real multi-account connections have a UUID
   `id` distinct from the host `domain`. Only one caller (`getCloudSession`), so the change is contained. Add a
   regression test asserting a legacy domain-keyed self-managed connection uses the provider-scoped path.

2. **Symptom B failure-path hardening (independent):** on a failed forced-resync refetch, don't leave a descriptor with
   no backing secret reported as connected — e.g. don't `preserveConfigured` when the follow-up `ensureSession` returns
   nothing, or reconcile-prune token-less descriptors.

### Notes
Only self-managed **cloud** integrations are affected (`isCloudGitSelfManagedHostIntegrationId`). Cloud providers with a
canonical domain and connections that carry a real backend `tokenId` are unaffected.

ianhattendorf:
Fixes #5497.

## Problem

Self-managed cloud integrations (e.g. GitHub Enterprise, GitLab Self-Hosted) were requesting `v1/provider-tokens/tokens/<domain>` instead of `/tokens/<uuid>`, which the backend rejects as a non-uuid — repeated on every session refresh.

`getConfiguredConnectionId` returned a legacy/migrated connection's descriptor id, which for self-managed hosts is backfilled from the *domain* (see `ConfiguredIntegrationDescriptor.id`). Using that as a per-connection token id produced the bad request.

A related failure mode: a forced re-sync deletes the cloud secret up front but preserves the descriptor so the connection keeps reporting connected while a fresh token is fetched. If that refetch failed, the connection was left reported as connected with **no backing token** — permanent for self-managed legacy connections, since reconcile can't restore an id-less connection.

## Fix

- **`getConfiguredConnectionId`** now only returns an id that is a real per-connection token id (`id !== domain`); legacy connections fall through to the provider-scoped primary endpoint instead of targeting `/tokens/<domain>`.
- **`deleteSession`** gains a `preserveConfigured` option, and `syncCloudConnection` drops the orphaned descriptor when a forced re-sync's replacement fetch returns nothing — a clean disconnect on failure instead of a token-less "connected" state (matches pre-multi-account behavior).

## Testing

- Added `configuredIntegrationService` unit tests: `getConfiguredConnectionId` returns `undefined` for a legacy domain-keyed connection, and the real token id for a genuine multi-account connection.
- Added a `reconcileConnections` test: a failed forced re-sync drops the connection (descriptor + secret) instead of leaving it token-less.
- Both new tests were confirmed to fail without the corresponding source change.
- `pnpm run build` (type-check + lint) passes; all 147 integrations package tests pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 
 - Removes the _Pro_ feature badge from the _Commit Graph_ header &mdash; the _Start New_ menu now occupies that area ([#5447](https://github.com/gitkraken/vscode-gitlens/issues/5447))
 
+### Fixed
+
+- Fixes self-managed cloud integrations (e.g. GitHub Enterprise, GitLab Self-Hosted) repeatedly issuing a token request that the server rejects on every session refresh, and legacy connections getting stuck reported as connected with no usable token &mdash; the per-connection token fetch no longer uses the host domain as the token id, and a connection whose token can no longer be fetched is now cleanly disconnected instead of left token-less ([#5497](https://github.com/gitkraken/vscode-gitlens/issues/5497))
+
 ## [18.3.0] - 2026-07-09
 
 ### Added
--- a/packages/plus/integrations/src/__tests__/configuredIntegrationService.test.ts
+++ b/packages/plus/integrations/src/__tests__/configuredIntegrationService.test.ts
@@ -85,6 +85,103 @@ suite('ConfiguredIntegrationService — multi-account (#5430)', () => {
 		assert.equal(session.accessToken, 'ent');
 	});
 
+	test('getConfiguredConnectionId returns undefined for a legacy self-managed connection keyed by domain', async () => {
+		const runtime = createFakeRuntime();
+		// Legacy/migrated single connection: hydration backfills the descriptor id from the domain, so
+		// `id === domain` (see ConfiguredIntegrationDescriptor.id). That is NOT a real backend token id, and
+		// returning it would target `v1/provider-tokens/tokens/{domain}` (rejected by the backend as a non-uuid).
+		await runtime.storage.store('integrations:configured', {
+			'cloud-github-enterprise': [
+				{ cloud: true, integrationId: 'cloud-github-enterprise', domain: 'gh.example.com', scopes: 'repo' },
+			],
+		});
+
+		const service = new ConfiguredIntegrationService(runtime);
+
+		assert.equal(
+			service.getConfiguredConnectionId(
+				GitSelfManagedHostIntegrationId.CloudGitHubEnterprise,
+				'gh.example.com',
+				true,
+			),
+			undefined,
+			'a domain-keyed legacy id must not be returned as a token id',
+		);
+	});
+
+	test('getConfiguredConnectionId returns the token id for a real multi-account self-managed connection', async () => {
+		const runtime = createFakeRuntime();
+		await runtime.storage.store('integrations:configured', {
+			'cloud-github-enterprise': [
+				{
+					id: 'ghe-a1',
+					cloud: true,
+					integrationId: 'cloud-github-enterprise',
+					domain: 'ghe-a.example.com',
+					scopes: 'repo',
+					primary: true,
+				},
+			],
+		});
+
+		const service = new ConfiguredIntegrationService(runtime);
+
+		assert.equal(
+			service.getConfiguredConnectionId(
+				GitSelfManagedHostIntegrationId.CloudGitHubEnterprise,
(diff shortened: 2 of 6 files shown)


julianmesa-gitkraken (approved):
LGTM

PR #5517, opened 2026-07-17
Fixes live updating of subscription and organization state in webviews

### LINKED ISSUE #5513: Graph header account/integrations state doesn't update after initial load
### Description

Follow-up to #5411 — the account/integrations header at the top of the Graph populates its state when the webview loads, but never reflects changes after that:

- Sign-in/sign-out: signing out from the Graph header itself updates the Home view correctly, while the Graph header keeps showing the previous account
- Switching organization (e.g. via the Home view) doesn't update the Graph header
- Plan/trial state changes — including the Subscription Simulator — are ignored; only a full webview reload picks up the new state
- The integrations chip also presents stale account/plan-gated content after such changes

Since the header state is frozen at load time, everything downstream is stale: the plan tier label, the upgrade CTA, and the plan-gated presentation of the integrations chip.

Applies to both header presentations — the full-width bar and the inline mode added in #5449 — since both feed from the same contexts.

### Root cause (verified)

The per-webview `SubscriptionService` RPC service seeds its reactive signals (`subscriptionState`, `hasAccountState`, `avatarState`, `organizationsCountState`) once in its constructor (`src/webviews/rpc/services/subscription.ts:80-84`). The only code that ever updates them afterwards lives inside the `container.subscription.onDidChange` handler — and that handler is registered **lazily** by `createRpcEventSubscription` (`src/webviews/rpc/eventVisibilityBuffer.ts:226-229`): it is armed only when a webview client calls `onSubscriptionChanged(handler)`.

The Graph app never does. `initAccountContexts` (`src/webviews/apps/plus/graph/graph-app.ts:564-643`) awaits the bridged signal proxies and installs them into the shared contexts, assuming the signal bridge keeps them fresh — but nothing ever updates the host-side signals for the Graph's service instance, so they stay frozen at their construction-time values.

The Home view works only by accident: `src/webviews/apps/home/events.ts:134-138` subscribes to `onSubscriptionChanged` for what its comment claims are "side effects only" (`refreshPromo`) — that incidental call is what arms the container listener and keeps Home's signals live. The comment's claim that state is "handled by signal bridges" is wrong without that subscription.

Notes from verification:

- The Graph's **integrations data channel is live** (it does subscribe to `onIntegrationsChanged`, `graph-app.ts:618-627`); the integrations chip appears stale because its rendering is gated on the frozen subscription signal (`integrations-chip.ts:209-217, 243`). Fixing the subscription signal should resolve most of the integrations symptom too.
- **Commit Details has the same latent bug**: `src/webviews/apps/commitDetails/events.ts:63` removed its `onSubscriptionChanged` call with the same mistaken "bridged from host" comment, so its `SubscriptionService` signals are frozen the same way.
- **`orgSettingsState` is frozen in every webview**: no code anywhere calls `onOrgSettingsChanged`, so the `onDidChangeContext` listener (`subscription.ts:101-121`) never registers. The Graph masks this via its separate legacy `DidChangeOrgSettings` IPC path, but Home (`active-work.ts:337`, `branch-card.ts`) and Commit Details consume the frozen signal for AI/drafts gating.

### Fix direction

Register the `container.subscription.onDidChange` (and `onDidChangeContext`) listeners **eagerly in the `SubscriptionService` constructor**, decoupling signal freshness from RPC-event subscription — this fixes the Graph, Commit Details, and the org-settings staleness everywhere at once. A per-webview workaround (adding an `onSubscriptionChanged` call to `graph-app.ts` like Home's) would fix only the Graph and leave the trap armed for the next consumer.

sergeibbb:
# Description

Closes #5513

The account/subscription header at the top of the Commit Graph populated its state on webview load and never updated afterwards — sign-in/sign-out, organization switches, plan/trial changes, and the Subscription Simulator were all ignored until a full webview reload.

**Root cause**: the per-webview `SubscriptionService`'s reactive signals were only updated inside the `container.subscription.onDidChange` handler, which `createRpcEventSubscription` registers lazily — only when a client subscribes a handler to the RPC change event. The Graph app never subscribes (it reads the bridged signals directly), so its service instance's signals stayed frozen at construction-time values. The Home view stayed live only because it subscribes to `onSubscriptionChanged` for an unrelated side effect (`refreshPromo`), incidentally arming the listener.

## Changes

- Registers the `container.subscription.onDidChange` and `onDidChangeContext` listeners eagerly in the `SubscriptionService` constructor, decoupling signal freshness from RPC-event subscription; the lazy RPC-event subscriptions now only forward buffered events
- Makes `SubscriptionService` disposable; `proxyServices` (moved to a new dependency-free `rpc/services/proxy.ts` module) collects disposable services behind a non-enumerable symbol, and the webview controller releases them at teardown via the new `disposeServices` — deliberately outside `SubscriptionTracker`, whose `reset()` on RPC reconnection would otherwise orphan the eager listeners
- Also fixes, via the same change: frozen account state in Commit Details (same latent bug — it removed its `onSubscriptionChanged` call trusting signals were self-fresh), and frozen organization AI/drafts settings (`orgSettingsState`) in every webview (`onOrgSettingsChanged` has no callers anywhere)
- Corrects the misleading "state handled by signal bridges" comments in `home/events.ts` and `commitDetails/events.ts` to point at where the freshness guarantee actually lives

## Testing

- `pnpm run check`: clean
- `pnpm run test`: 1290 passing, including 3 new tests for the service disposal mechanism (`servicesProxy.test.ts`)
- Live-verified in a dev instance with `gitlens.graph.experimental.homeHeader.enabled`: `gitlens.plus.simulate.subscription` flipped the Graph account chip Community → Pro → Pro Trial live with no reload, including simulator re-runs (the case called out in the issue)
- Note: `SubscriptionService` itself is not unit-testable today — importing it pulls the `context.ts → command.ts → container.ts` chain, which cannot initialize in the self-contained test bundle (pre-existing circular-init limitation); the test file documents this

# Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 
 ### Fixed
 
+- Fixes the _Commit Graph_ header's account/subscription state never updating after the view loads &mdash; sign-in/sign-out, organization switches, and plan changes (including the subscription simulator) now update the header live instead of requiring a reload; the same underlying fix also keeps account and organization state live in _Commit Details_ and organization AI/drafts settings live in the _Home_ view ([#5513](https://github.com/gitkraken/vscode-gitlens/issues/5513))
 - Fixes the _Commit Graph_ working changes (WIP) details continuing to show files after they were committed or discarded outside of VS Code (e.g. from a terminal)
 - Fixes the _Commit Graph_ view's progress indicator repeatedly flashing while nothing appears to change (most noticeable after the VS Code window regains focus) &mdash; last-fetched updates now coalesce into a single pending update instead of queuing one per `FETCH_HEAD` change
 
--- a/src/webviews/apps/commitDetails/events.ts
+++ b/src/webviews/apps/commitDetails/events.ts
@@ -60,8 +60,8 @@ export function setupSubscriptions(
 				handleRepositoryChanged(state, event, actions),
 			),
 		() => services.config.onConfigChanged(() => handleConfigChanged(actions)),
-		// Note: onSubscriptionChanged removed — hasAccount signal bridged from host
-		// Note: onOrgSettingsChanged removed — orgSettings signal bridged from host
+		// Note: onSubscriptionChanged/onOrgSettingsChanged removed — the bridged hasAccount and
+		// orgSettings signals are kept fresh by SubscriptionService's eager listeners (#5513)
 		() =>
 			services.integrations.onIntegrationsChanged(data => handleIntegrationsChanged(state, data.hasAnyConnected)),
 		() => services.ai.onModelChanged(model => state.aiModel.set(model)),
--- a/src/webviews/apps/home/events.ts
+++ b/src/webviews/apps/home/events.ts
@@ -131,7 +131,9 @@ export function setupSubscriptions(
 		// Generic events — from WebviewEventsService
 		// ============================================================
 
-		// Subscription changed — state handled by signal bridges, keep side effects only
+		// Subscription changed — state flows via the bridged signals, whose freshness is
+		// guaranteed by SubscriptionService's eager listeners (#5513); this subscription
+		// exists solely for its side effects and is safe to remove if they go away
 		() =>
 			services.subscription.onSubscriptionChanged(() => {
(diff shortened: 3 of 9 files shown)


sergeibbb:
augment review

Issue #5449, opened 2026-07-06
Add an inline mode to the Graph header bar for vertically constrained views


sergeibbb:
### Description

Follow-up to #5411 — from the GitLens standup on 2026-07-06.

- The Graph header bar should support two modes, switching based on the available vertical space:
    - **Bar mode** (current behavior) — a full-width bar at the top of the Graph, used when there is enough vertical space (e.g. when the Graph is the main view of the GitLens panel)
    - **Inline mode** — used when the view is vertically constrained (e.g. the Graph in the bottom panel), so the header doesn't take up yet another full row
- In inline mode, consolidate the account badge and the integrations entry into the top right of the existing Graph header (e.g. integrated into the existing header bar as buttons, near "Start Work")

sergeibbb:
@eamodio 

I've collapsed the account header on vertically narrow views, here is a video. How do you find it? Is it what's expected?


https://github.com/user-attachments/assets/46edcb59-921e-4471-9593-8510e66dfa88

cc @d13

nikolay-1986:
**Verification: Passed** ✅

Verified statically against `main`, build clean (`pnpm run build` exit 0), and confirmed live via the vscode-inspector (Paid subscription simulation + `graph.experimental.homeHeader.enabled`, driving `resize_window` across the 400/460 boundary while measuring the real `.graph` height). No functional bugs.

Commits reviewed: `c1c492e97` (inline-mode logic + observer robustness), `b90729e03` (compact chip presentation), `2756ad5cc` (chips integrated into the graph header).

**Live results** (window → measured `.graph` height → observed state):

| State | Window | `.graph` height | `_headerInlineMode` | Account bar | Inline compact chips |
| --- | --- | --- | --- | --- | --- |
| Baseline | default | 808 | `false` | present | — |
| Short (enter) | 1200×480 | **388** (<400) | `true` | removed | account + integrations, `compact=true` |
| Dead-band | 1200×525 | **433** (401–459) | `true` (holds) | removed | present |
| Back tall (exit) | 1200×570 | **478** (>460) | `false` | back | — |
| Narrow + short | 560×480 | 388, width 220 | `true` | removed | present but clipped |

- **Hysteresis is correct**: enters inline below 400, holds through the 401–459 dead-band (no revert at 433), exits above 460. No oscillation.
[...]

PR #5521, opened 2026-07-21
Adapt the Commit Graph feature-gate to constrained areas

### LINKED ISSUE #5519: Adjust graph gate design to fit in smaller area
### Description

Sub-issue of #5391 (view consolidation). Part of the Graph-as-main-view work: when the Graph is placed in a vertically/horizontally constrained area (e.g. bottom panel, narrower side panel), the current feature-gate design (Pro upsell shown when not entitled) doesn't adapt and doesn't fit well in the smaller area.

Adjust the gate's layout/content so it degrades gracefully at smaller sizes, consistent with the other Graph-in-constrained-space work already done for the header/sidebar (#5449, #5500, #5513).

sergeibbb:
Closes #5519 (part of #5391)

## Summary

Makes the Commit Graph's Pro feature-gate (`gl-graph-gate`) degrade gracefully when the Graph is docked in a constrained area — a narrow side bar or the short bottom panel — consistent with the other Graph-in-constrained-space work (#5449, #5500, #5513). Previously the gate was sized in viewport units, used a fixed two-column feature list, and — because the CTA is focused on open — scrolled the user to a mid-content position where neither the heading nor the Upgrade button was reliably visible.

**The full-height dialog (full-height editor tab, maximized panel) is unchanged.** Compaction keys off actual available height rather than placement, so a short split-editor pane (below ~450px) intentionally compacts as well. Every adaptation is scoped behind a container query, and no feature items are ever hidden — constrained layouts compact and scroll instead.

- Normal view:

    Image

- Horizontally narrow view: one column

    Image

- Vertically narrow view: collapsing titles to one line witht the text, floating around icons, smaller intervals

    Image

- Image

- Both: smaller gaps, floating texts, one column

    Image

- https://github.com/user-attachments/assets/d10e4b5a-8d41-40a4-8db7-4ee7aa89841c

## What changes

**Narrow width (side bar)**
- Feature list collapses two columns → one intrinsically (`repeat(auto-fit, minmax(min(32rem, 100%), 1fr))`).
- Card tracks the gated view's width via container units (`100cqi`) instead of `90vw` (the window's width — wrong basis inside a dock).
- Tighter content padding and title size.

**Short height (bottom panel)**
- Header collapses to one line (title + lede).
- Trial message + promo collapse to one line; footer paragraph/rule rhythm tightens.
- Feature-list horizontal gaps halve.
- Each item's icon floats (`float: inline-start`) so its copy flows around and under it instead of sitting in a rigid second column; the run-in title is sized to match the body.

**Both**
- The CTA focus no longer scrolls the gate (`focus({ preventScroll: true })`), so it opens at the top with the heading and context visible.

## Mechanism

`gl-feature-gate`'s `:host` becomes a `container-type: size` query container (safe — it's an `inset: 0` overlay, so its size is content-independent in both axes; the top-layer `<dialog>` resolves the container via its flat-tree ancestry, verified live). Adaptations live behind `@container (max-width: 45rem)` and `@container (max-height: 45rem)`. This matches the container-query pattern already used by `gl-details-header`, `gl-wip-tree-pane`, and the graph minimap — not viewport media queries.

Timeline also imports `featureGateContentStyles` and mounts the wrapper, so it **shares** these adaptations in its constrained Visual File History view — verified live to render correctly (one-line header, compacted title/padding; the `.list` float/gap rules are inert as timeline has no feature list). The rebase conflict-indicator and merge-target-status mount the wrapper but render their own content and don't import the content styles, so only `.content` padding / plus-state rhythm can reach them.

## Validation

- Reproduced live via `gitlens.plus.simulate.subscription` (trial-expired) + simulated private-repo visibility, across editor / side-bar / bottom-panel placements.
- Verified computed layout at each breakpoint: normal dialog identical to before (two-column flex list, full-size header); panel/side-bar compact as intended; no feature items hidden.
- Verified timeline (the other consumer of the content styles) shares the adaptation and renders correctly in its constrained Visual File History view; the rebase conflict-indicator and merge-target-status don't import the content styles and render their own content.
- `pnpm run check` clean.

## Out of scope

`gl-graph-access-account` (the account-less "Get Started" screen) has the same constrained-space problem but is a separate component — candidate follow-up.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 
 - Changes the default _Commit Graph_ lane colors for dark and high-contrast themes to a new vibrant, perceptually-uniform palette &mdash; every lane shares the same perceived brightness so no lane visually dominates; light themes keep the previous colors, and any `gitlens.graphLane1Color`&ndash;`gitlens.graphLane10Color` customizations in `workbench.colorCustomizations` are still honored
 - Changes the _Commit Graph_ sidebar to be unpinned by default &mdash; the sidebar now floats over the graph and auto-collapses when it loses focus; pin it (or set `gitlens.graph.sidebar.pinned` to `true`) to restore the previous shared-space layout ([#5447](https://github.com/gitkraken/vscode-gitlens/issues/5447))
+- Changes the Pro feature gate to adapt to constrained spaces &mdash; in narrow placements (e.g. the side bar) the feature list collapses to a single column, and in short placements (e.g. the bottom panel) the header, messaging, and spacing compact so the pitch and _Upgrade_ action stay legible without hiding any content; full-size placements are unchanged ([#5519](https://github.com/gitkraken/vscode-gitlens/issues/5519))
 
 ### Fixed
 
--- a/src/webviews/apps/plus/graph/gate.ts
+++ b/src/webviews/apps/plus/graph/gate.ts
@@ -20,7 +20,10 @@ export class GlGraphGate extends SignalWatcher(LitElement) {
 		featureGateContentStyles,
 		css`
 			gl-feature-gate::part(section) {
-				width: 90vw;
+				/* Container units (not vw): size against the gate overlay's own box (the
+				   feature-gate host) rather than the webview viewport, so the card keeps tracking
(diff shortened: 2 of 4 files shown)


sergeibbb:
augment review

sergeibbb:
augment review

Issue #5411, opened 2026-06-29
Place home view avatar/header bar at the top of the Graph view


trevor-polidore:
### Description

- Move the avatar bar (with account and integrations info) from the top of the home view to the top of the Graph view. This keeps account and integrations info always available when the Graph view becomes the main view of the GitLens panel
    - Ensure responsive behavior for different layouts & sizing

<img width="548" height="197" alt="Image" src="https://github.com/user-attachments/assets/69f2f63f-6cba-421e-8737-dccff903a0b0" />

nikolay-1986:
**Verification: Passed ✅**

Verified commit 5759802d6 live (built from `main`, fresh instance, Commit Graph open).

- The account + integrations bar (`gl-account-bar`, extracted from `gl-home-header`) renders at the **top of the Graph view, directly above the graph header** — confirmed in the DOM (the bar precedes `gl-graph-header`; the bar occupies the top strip and the header starts immediately below it) and visually via screenshot: the account chip on the left (account/subscription state + Upgrade) and the integrations chip on the right (Connect + provider icons), spanning the full width above the header/search and the graph.
- The bar contains the `gl-account-chip` and `gl-integrations-chip`, wired to the subscription/integrations/ai contexts the Graph app now provides and populates from the host.
- Init is guarded end-to-end (a rejected service or failed subscription can't break the graph) with late-completion/disconnect teardown to avoid leaking host subscriptions, and `show()` is guarded against an unresolved query on both components.

No functional bugs found.

PR #5546, opened 2026-07-23
Adds an inline mode to the Graph header bar for vertically constrained views

### LINKED ISSUE #5449: Add an inline mode to the Graph header bar for vertically constrained views

sergeibbb:
Closes #5449 (sub-issue of #5391; follow-up to #5411)

## Summary

Makes the experimental Graph account header adaptive to vertical space (gated by `gitlens.graph.experimental.homeHeader.enabled`):

- **Bar mode** (unchanged current behavior) — the full-width account/integrations bar above the graph header, when the view has room
- **Inline mode** — when the graph webview is vertically constrained (e.g. docked in the bottom panel), the bar gives up its dedicated row and its chips render as compact buttons in the top right of the existing graph header, next to the Start New menu

### How it works

- Mode is driven by the existing `.graph` ResizeObserver with enter/exit hysteresis (`headerInlineEnterPx = 400` / `headerInlineExitPx = 460`), mirroring the `_autoEffectiveLocation` width pattern — no new observers, no height container queries (deliberately avoided: `container-type: size` containment is a layout-loop risk in the graph's observer ecosystem)
- The chips consume the shared subscription/integrations/AI contexts provided at the `GraphApp` level, which cross the header's shadow boundary — no rewiring; context init remains flag-gated and mode-independent, so constrained-at-startup hydrates correctly
- New opt-in `compact` attribute on `gl-account-chip` (avatar-only) and `gl-integrations-chip` (icons-only), with `aria-label`s preserving the accessible names — default off, Home view untouched
- With no repository open the header row (the inline chips' host) isn't rendered, so the bar stays even when constrained
- Fixes a pre-existing observer lifecycle bug exposed by the feature: `ensureGraphObserved` latched a boolean, so after a sign-out/sign-in cycle the recreated `.graph` element was never re-observed, freezing all size-driven behavior (this also froze the `auto` details location on main). Now identity-tracked with re-observe on element change and reset on disconnect.

### Threshold note for review

400/460px were chosen against typical bottom-panel heights and validated live, but the boundary is a product call — feedback welcome before this graduates from the experimental flag.

## Test plan (live-verified)

- [x] Bar mode at height ≥ exit threshold: full-width bar, no chips in header
- [x] Inline mode at height < enter threshold: bar gone, both chips in header right group, fitting within the viewport
- [x] Hysteresis: no flapping across the 400–460 dead band in either direction
- [x] Constrained-at-startup: chips hydrate live data when the first-ever mode is inline
- [x] Both chip popovers open fully on-screen in a 338px-tall viewport (right-edge aligned)
- [x] Sign-out → sign-in → resize: mode switching keeps working (observer re-attach fix)
- [x] Flag off: no bar, no chips, at any height
- [x] Home view header unaffected (chips non-compact there); bar-mode chips unaffected
- [x] Console clean across mode flips
- [ ] Web (vscode.dev) smoke — no environment-specific code touched, but untested

### DIFF
--- a/src/webviews/apps/plus/graph/graph-app.ts
+++ b/src/webviews/apps/plus/graph/graph-app.ts
@@ -146,6 +146,13 @@ const detailsMaxPct = 80;
 const detailsAutoBottomEnterPx = 820;
 const detailsAutoBottomExitPx = 920;
 
+// Height thresholds (px) for the account bar's inline mode (issue #5449) — below `enter` the bar
+// gives up its own row and its chips move into the graph header's right group; above `exit` it
+// returns to a full-width row. The dead-band between them prevents flicker when the view is
+// dragged across the boundary (e.g. resizing the bottom panel).
+const headerInlineEnterPx = 400;
+const headerInlineExitPx = 460;
+
 const minimapDefaultPx = 40;
 const minimapMaxPct = 40;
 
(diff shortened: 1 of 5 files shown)


sergeibbb:
augment review

sergeibbb:
augment review

PR #5565, opened 2026-07-24
Use the provider's canonical identity in resolveRepository (#5558)

### LINKED ISSUE #5558: resolveRepository builds the identity from the input URL and discards the provider's canonical repo, so renames are undetectable
Found in a full review pass of the `core` branch (at `708112435`), comparing `IntegrationService.resolveRepository` against the gkcli `repo resolve` semantics it replaces (kepler#1329).

## Problem

`resolveRepository` (`packages/plus/integrations/src/integrationService.ts`, around the `getRepoInfo` call) resolves the repo through the provider and then builds the returned identity **from the owner/name parsed out of the input remote URL**, discarding the response:

```ts
const repo = await integration.getRepoInfo({ owner, name, project, connectionId });
...
const identity: RepositoryIdentity = {
    providerId: id,
    domain: domain,
    owner: owner,   // <- from the URL matcher, not from `repo`
    name: name,     // <- from the URL matcher, not from `repo`
    ...
};
```

The `ProviderRepository` returned by `getRepoInfo` carries the **canonical** `namespace`/`name` from the provider (provider-apis `normalizeRepository`: `namespace: repository.owner.login, name: repository.name`; GitHub's REST/GraphQL repo lookup follows the 301 rename redirect, so a stale old name resolves to the new canonical identity). That canonical data is fetched and then thrown away.

## Why this matters for Kepler (kepler#1329, kepler#999/#1010)

The whole point of `gk repo resolve` in Kepler is **canonicalization**: `ResolvedRepoIdentity.renamed` (`kepler/src/backend/provider/repo-resolution.ts:35-39`) marks a host-side rename whose stale local remote still points at the old name, and drives the `{ canonical }` policy in `correlate.ts` (use the provider's canonical name instead of the literal remote name). gkcli computes it by comparing input vs response:

```go
// gkcli internal/providers/repository/github.go:116
renamed := !strings.EqualFold(resolved.Owner, input.Owner) || !strings.EqualFold(resolved.Name, input.Name)
```

With the current facade, a `CoreGitLensProviderAdapter` would always receive the literal input identity marked `resolved`, `renamed` would always be `false`, and the kepler#999 regression (correlations silently missing on renamed repos) would come back with no signal.

## Suggested fix

- Populate `RepositoryIdentity.owner`/`name` from `repo.namespace`/`repo.name` (the canonical values), keeping the parsed values only as fallback when the provider response omits them.
- Add `renamed: boolean` to `RepositoryIdentity`, computed as a case-insensitive comparison of the parsed input owner/name vs the canonical response (mirroring gkcli).
- Keep `remoteUrl` as the original input so the caller can still key the resolution back to the remote it asked about.

One caveat to verify per provider: Bitbucket Server/Azure lookups may echo the input casing rather than a canonical one; the case-insensitive compare covers that, matching gkcli's `EqualFold`.

Part of the Kepler migration epic (gitkraken/kepler#1322, sub-issue kepler#1329).

julianmesa-gitkraken:
## Summary

`IntegrationService.resolveRepository` resolved the repo through the provider and then built the returned `RepositoryIdentity` from the owner/name parsed out of the **input** remote URL, discarding the canonical `namespace`/`name` in the provider response. GitHub's REST/GraphQL repo lookup follows the 301 rename redirect, so a stale old name resolves to the new canonical identity; throwing that away made host-side renames undetectable.

This matters for the Kepler migration (kepler#1329, kepler#999/#1010): the whole point of `gk repo resolve` is canonicalization. Without the canonical values and a rename signal, a `CoreGitLensProviderAdapter` would always see the literal input identity marked `resolved` with `renamed` always `false`, reintroducing the kepler#999 regression (correlations silently missing on renamed repos).

## Changes

- Populate `RepositoryIdentity.owner`/`name` from `repo.namespace`/`repo.name` (canonical), falling back to the parsed values only when the response omits them (truthy `||`, so an empty-string response degrades to the parsed remote rather than emitting an empty identity or a spurious rename).
- Add `renamed: boolean` to `RepositoryIdentity`, computed as a case-insensitive compare of the parsed input owner/name vs the canonical response. This mirrors gkcli's `EqualFold`, so hosts that echo the input casing (e.g. Bitbucket Server/Azure) aren't flagged.
- Keep `remoteUrl` as the original input so the caller can key the resolution back to the remote it asked about.

`project` intentionally stays the parsed value and isn't part of the `renamed` compare: the `getRepoInfo` lookup is scoped by the parsed project, so a mismatching canonical project can't come back (it would 404 into `not-found` instead), matching the issue's owner/name scope.

## Tests

Extended `resolveRepository.test.ts`:
- canonical owner/name come from the provider response, `renamed: true` on a name change, `remoteUrl` preserved
- owner-only rename (repo transferred between accounts) flags `renamed`
- case-only difference does **not** flag `renamed`
- response omitting owner/name falls back to the parsed remote, `renamed: false`

Full package test suite passes; `pnpm run check` and `tsc -b` are clean.

Closes #5558

### DIFF
--- a/packages/plus/integrations/src/__tests__/resolveRepository.test.ts
+++ b/packages/plus/integrations/src/__tests__/resolveRepository.test.ts
@@ -88,6 +88,77 @@ suite('resolveRepository (#5438)', () => {
 		}
 	});
 
+	test('builds the identity from the canonical provider response, marking a rename', async () => {
+		const manager = createIntegrationManager(createFakeRuntime());
+		const gh = await connect(manager, GitCloudHostIntegrationId.GitHub, 'github.com');
+		// The remote points at the stale `octocat/old-name`; the provider follows the 301 redirect and
+		// returns the canonical `octocat/hello`.
+		stubGetRepo(gh, () =>
+			Promise.resolve({ id: 'r1', namespace: 'octocat', name: 'hello' } as unknown as ProviderRepository),
+		);
+
+		const result = await manager.resolveRepository({ remoteUrl: 'https://github.com/octocat/old-name.git' });
+		assert.equal(result.resolution.status, 'resolved');
+		assert.equal(result.resolution.identity?.owner, 'octocat', 'owner comes from the canonical response');
+		assert.equal(result.resolution.identity?.name, 'hello', 'name comes from the canonical response');
+		assert.equal(result.resolution.identity?.renamed, true, 'a differing canonical name flags renamed');
+		assert.equal(
+			result.resolution.identity?.remoteUrl,
+			'https://github.com/octocat/old-name.git',
+			'remoteUrl keeps the original input',
+		);
+
+		manager.dispose();
+	});
+
+	test('flags renamed when only the owner changed (repo transferred between accounts)', async () => {
+		const manager = createIntegrationManager(createFakeRuntime());
+		const gh = await connect(manager, GitCloudHostIntegrationId.GitHub, 'github.com');
+		// The repo kept its name but moved to a new owner; the name-only rename test above keeps the owner
+		// constant, so this exercises the owner side of the OR compare.
+		stubGetRepo(gh, () =>
+			Promise.resolve({ id: 'r1', namespace: 'new-org', name: 'hello' } as unknown as ProviderRepository),
+		);
+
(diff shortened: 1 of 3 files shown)


julianmesa-gitkraken:
augment review

Issue #5588, opened 2026-07-28
Scope picker and file list go stale when a working-changes row disappears, so a review covers files the user cannot see or exclude


nikolay-1986:
### Summary

When a working-changes row disappears from the scope picker (e.g. the user stages everything while a mode is open), the picker's highlight, the stored scope, and the Files Changed curation list all drift apart. The next review then runs on files that were never listed — so they cannot be excluded from the review.

### Details

With a review scoped to unstaged changes only and staged changes also present, staging the remaining unstaged files removes the `Unstaged changes` row. Three things then disagree:

- The picker highlights `Staged changes` as included, because the stored row ids no longer resolve and the range falls back to its defaults.
- The stored scope is still unstaged-only — the selection-to-range sync bails out when the ids don't resolve instead of re-deriving or re-emitting, so nothing updates the scope.
- The curation list still shows the previously scoped files; the scope file list is refetched on a branch-state change or the panel's manual Refresh, but not on a working-tree change.

Starting the review from this state re-derives the scope from what the picker shows (staged), and the run covers three files while the list shows two. The extra file has no row in the curation list, so its checkbox does not exist and the user has no way to exclude it. The stale list also under-reports the file count the panel advertises before the run.

### Impact

Reviews silently cover files the user never saw and cannot deselect, and the panel's file count misreports the scope. Users who stage or unstage while a mode is open — a normal thing to do mid-review — hit this. Low-to-medium severity; the manual Refresh button recovers it, but nothing indicates that a refresh is needed.

### Validation

1. In a repo, stage a change to one file and leave an unstaged change in a second file.
2. Select the Working Changes row in the Commit Graph and open the Review panel.
3. Move the bottom scope handle up so the scope is unstaged-only — the file list shows the unstaged file.
4. From a terminal, stage everything (`git add -A`).
5. Observe the picker and the list: the picker now highlights `Staged changes`, while the file list still shows the pre-staging set.
6. Click Start Review.
7. Expected: the picker, the scope, and the list agree, and the review covers exactly the listed files. Actual: the review covers the full staged set, including a file that was never listed and could not be excluded.

### Suggested fix

When the stored range ids no longer resolve, re-derive the range and reconcile the stored scope (and refetch the scope file list) instead of returning early. Refreshing the scope file list on a working-tree change while a mode is open would keep the curation list and the run in agreement.

### Notes

Live-verified on `main` against a real repository. Pre-existing; found while verifying #5577.

[...]

PR #5600, opened 2026-07-28
Fix compose scope picker showing a forbidden unstaged-only selection

### LINKED ISSUE #5587: Compose scope picker can show an unstaged-only selection it forbids, then silently composes staged changes anyway
### Summary

The compose scope picker enforces its "a selection containing unstaged must also contain staged" floor only while the user is moving a handle. When the row list changes underneath a selection that was valid before, the picker keeps rendering an unstaged-only selection — a state its own floor forbids — and the compose run then silently widens the scope to include staged changes without telling the user.

### Details

Reproduced by staging a file while compose sits idle:

1. With only unstaged changes present, compose opens with an unstaged-only scope. There is no staged row yet, so the selection is legitimate.
2. Staging a file from outside the panel adds a `Staged changes` row. The floor now says the bottom edge may not sit above that row, but the rendered selection still ends on `Unstaged changes` — `Staged changes` renders greyed out/excluded.
3. Clicking Compose flips the scope to staged + unstaged behind the scenes and composes the staged file too — three commits for three files, where the picker showed only the unstaged ones as included.

The floor is applied to interactive movement (drag, keyboard, row click) but not when the range is derived from the controlled selection, so nothing reconciles the displayed range after a working-tree change. The compose run path then corrects the scope after the fact, which keeps the engine correct but makes the picker's state a lie.

### Impact

A user who deliberately narrowed compose to unstaged work gets staged work folded into the generated commits with no warning, and no visible cue that the excluded row was re-included. The follow-through action is Commit Changes, so the mismatch lands in real commits. Medium severity.

### Validation

1. In a repo, make an unstaged change to one tracked file and ensure nothing is staged.
2. Select the Working Changes row in the Commit Graph and open the Compose panel — the scope is unstaged-only and no staged row exists.
3. From a terminal, stage a change to a different file.
4. Observe the picker: a `Staged changes` row appears, excluded, while `Unstaged changes` stays included.
5. Click Compose.
6. Expected: either the picker reconciles to include the staged row (with the constraint visible), or compose runs on exactly what is shown. Actual: the scope silently becomes staged + unstaged and the staged file is composed.

### Suggested fix

Reconcile the derived range against the floor whenever the row list or the controlled selection changes, so the displayed selection can never contradict the constraint. If the widening has to happen, surface it — the staged row should render as forcibly included rather than excluded.

### Notes

Live-verified on `main` against a real repository. Pre-existing; found while verifying #5577 (whose compose path is unaffected and correct).

ramin-t:
Closes #5587

### Problem

The compose scope picker enforces a **floor**: in compose mode, a selection that includes the `Unstaged changes` row must also include the `Staged changes` row (unstaged diffs are index-relative, so composing unstaged without staged is ill-defined). That floor was only applied while the user was *moving* an edge — drag, keyboard, or row-click — never when the selected range was derived from the controlled `selection` prop.

As a result, if compose opened with only unstaged changes present (a legitimate unstaged-only scope, since no staged row existed yet) and the user then staged a file from outside the panel, a `Staged changes` row appeared but the rendered selection still ended on `Unstaged changes` — the staged row rendered *excluded*, a state the floor forbids. Clicking Compose then silently widened the scope to include staged and composed the staged file too, with no cue that the excluded row had been re-included.

### Fix

Make the **effective end edge honor the floor** in the one getter every consumer reads, so the invariant "the end never sits above the floor" holds no matter how the end was set — including the controlled-derivation path, not just interactive moves.

- Extracted the pure range/floor logic (`getDefaultStart`/`getDefaultEnd`, `getMinEndIndex`, `resolveStartIndex`, `resolveEndIndex`) into a new lit-free `gl-commits-scope-pane.utils.ts`. `resolveEndIndex` clamps the resolved/derived end **up** to the floor.
- The component's `rangeStart` / `rangeEnd` / `minEndIndex` getters now delegate to those helpers.

This single reconciliation makes the **display** (the staged row now renders as included, and the end handle sits below it), the **keyboard target**, the public **`selectedIds`** getter, and therefore the **compose run** all reflect the truth — the run composes exactly what is shown, with no silent widening. It is also self-healing: if the staged row later disappears, the floor falls back to the range start and the end returns to unstaged.

The `buildScopeFromPicker` / `runCompose` widening backstops are kept as defense-in-depth for non-picker entry points; with this change they are no-ops for the picker flow rather than silent corrections.

### Tests

The extraction is what makes this unit-testable: the component registers a custom element at import and pulls in `lit` + `.css.js`, so it can't be imported in the DOM-less test env — mirroring the existing `signature.utils.ts` / `signature-badge.test.ts` precedent. Added `__tests__/gl-commits-scope-pane.utils.test.ts` (19 cases). The `resolveEndIndex` cases fail against the pre-fix behavior (which returned the unstaged index), so they guard the regression; plus mode/default/fallback coverage.

### Notes

The related idle scope-state / Files-Changed list staleness on a working-tree change (the picker highlight vs. the curation list drifting) is tracked separately by #5588 and is intentionally out of scope here.

### DIFF
--- a/src/webviews/apps/plus/graph/components/__tests__/gl-commits-scope-pane.utils.test.ts
+++ b/src/webviews/apps/plus/graph/components/__tests__/gl-commits-scope-pane.utils.test.ts
@@ -1,0 +1,124 @@
+import * as assert from 'assert';
+import type { ScopeItem, ScopeItemState } from '../gl-commits-scope-pane.js';
+import {
+	getDefaultEnd,
+	getDefaultStart,
+	getMinEndIndex,
+	resolveEndIndex,
+	resolveStartIndex,
+} from '../gl-commits-scope-pane.utils.js';
+
+function item(id: string, state: ScopeItemState): ScopeItem {
+	return { id: id, label: id, state: state };
+}
+
+const unstaged = item('unstaged', 'uncommitted');
+const staged = item('staged', 'uncommitted');
+const commitA = item('a', 'unpushed');
+const commitB = item('b', 'pushed');
+const mergeBase = item('merge-base:x', 'merge-base');
+const loadMore = item('load-more', 'load-more');
+
+suite('gl-commits-scope-pane range/floor utils', () => {
+	suite('getMinEndIndex', () => {
+		test('review mode never floors', () => {
+			const items = [unstaged, staged, commitA];
+			assert.strictEqual(getMinEndIndex('review', items, 0), 0);
+		});
+
+		test('compose with no unstaged row does not floor', () => {
+			const items = [staged, commitA];
+			assert.strictEqual(getMinEndIndex('compose', items, 0), 0);
+		});
+
+		test('compose does not floor when the start is below unstaged (unstaged excluded)', () => {
+			const items = [unstaged, staged, commitA];
+			// rangeStart at staged (1): unstaged is not in the selection, so no floor is imposed.
+			assert.strictEqual(getMinEndIndex('compose', items, 1), 1);
+		});
+
+		test('compose floors the end down to staged when unstaged is included', () => {
+			const items = [unstaged, staged, commitA];
+			assert.strictEqual(getMinEndIndex('compose', items, 0), 1);
+		});
+
+		test('compose with unstaged but no staged row does not floor (unstaged-only is legitimate)', () => {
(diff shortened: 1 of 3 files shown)


ramin-t:
augment review

PR #5601, opened 2026-07-28
Reconcile scope picker, stored scope, and file list on working-tree changes

### LINKED ISSUE #5588: Scope picker and file list go stale when a working-changes row disappears, so a review covers files the user cannot see or exclude

ramin-t:
Closes #5588

## Problem

In the Commit Graph details **Review** (and **Compose**) panel, a WIP-scope selection is driven by three things that must stay in agreement:

- the scope **picker** highlight,
- the stored **`state.scope`**, and
- the **Files Changed** curation list (`scopeFiles`).

When a working-changes row disappears out from under the picker mid-mode — e.g. the review is scoped to *unstaged-only* and the user runs `git add -A` in a terminal — these three drift apart:

- The picker's stored range IDs no longer resolve, so it silently falls back to its auto-derived defaults and highlights **Staged changes**.
- `syncSelectionRange()` bailed out when the controlled selection didn't resolve, so nothing re-emitted `scope-change` — the stored scope stayed *unstaged-only*.
- `scopeFiles` was only refetched on a scope change or manual Refresh, never on a working-tree change — so the curation list kept showing the pre-staging set.

Starting the review from this state re-derives the scope from what the picker actually shows (staged), so the run covers files that were never listed and cannot be excluded, and the panel's advertised file count under-reports the scope. The manual Refresh button recovered it, but nothing indicated a refresh was needed.

## Fix

Two coordinated changes so the picker, the stored scope, and the file list re-converge automatically:

1. **Picker re-emits `scope-change` when its selection can no longer be honored** (`gl-commits-scope-pane.ts`). When the controlled `selection` no longer resolves against the current items, `syncSelectionRange()` now drops the stale stored IDs and flags a reconcile; `updated()` re-emits `scope-change` post-render with the effective (fallback) selection. This reuses the existing `handleScopeChange` → `buildScopeFromPicker` path, which reconciles the stored scope **identity** and refetches the scoped file list. The host's `scopeSelectionEqual` guard keeps benign items ticks silent and prevents any re-emit loop.

2. **Refetch `scopeFiles` on a working-tree change while a WIP-scope mode is open** (`detailsActions.ts`, `applyWipPayload`). Scoped to the active mode's repo and to WIP scopes. This keeps the file **set** fresh when the selection identity stays valid but the underlying files move (a case the picker re-emit doesn't cover).

Together: (1) reconciles scope identity when a row disappears; (2) keeps the file set fresh when the identity is unchanged but the files moved.

The selection-range resolution and default-range derivation were extracted into a small, dependency-free `scopeRange.ts` helper (`resolveSelectionRange` / `computeDefaultRange`) reused by the picker, keeping that logic unit-testable.

## Testing

- Type-check + lint clean (`pnpm run check:fix`).
- New unit tests: `scopeRange.test.ts` (range resolution incl. the row-disappeared case and default derivation) and `detailsActions.test.ts` (the working-tree-tick refetch and its guards — no mode / non-WIP scope / background repo). All pass.
- Manual, per the issue's validation steps: with a review scoped unstaged-only, `git add -A` in a terminal now leaves the picker, scope, and file list in agreement, and Start Review covers exactly the listed files.

### DIFF
--- a/src/webviews/apps/plus/graph/components/__tests__/detailsActions.test.ts
+++ b/src/webviews/apps/plus/graph/components/__tests__/detailsActions.test.ts
@@ -453,6 +453,102 @@ suite('DetailsActions', () => {
 		assert.strictEqual(state.wipStale.get(), false);
 	});
 
+	// A working-tree change (WIP tick) while a wip-scope mode is open must refetch the scoped file
+	// list, or the curation list (and the eventual run) stays pinned to the pre-change file set.
+	function wipFor(repoPath: string): Wip {
+		return {
+			changes: undefined,
+			repositoryCount: 1,
+			repo: { uri: `file://${repoPath}`, name: 'repo', path: repoPath, isWorktree: false },
+			stats: { added: 0, deleted: 0, modified: 0 },
+		};
+	}
+
+	function scopeFilesRecorder(sink: Array<[string, ScopeSelection]>) {
+		return createResource(async (_signal, repoPath: string, scope: ScopeSelection) => {
+			sink.push([repoPath, scope]);
+			return [];
+		});
+	}
+
+	const wipScope: ScopeSelection = { type: 'wip', includeStaged: true, includeUnstaged: false, includeShas: [] };
+
+	test('applyWipPayload refetches scope files on a wip-scope mode tick for the active repo', async () => {
+		const state = createDetailsState();
+		state.activeMode.set('review');
+		state.activeModeRepoPath.set('/repo');
+		state.scope.set(wipScope);
+
+		const fetches: Array<[string, ScopeSelection]> = [];
+		const resources = createResources({
+			wip: createResource(async (_signal, _repoPath: string) => ({ wip: wipFor('/repo') })),
+			scopeFiles: scopeFilesRecorder(fetches),
+		});
+		const actions = new DetailsActions(state, createServices(), resources);
+
+		await actions.refetchWipQuiet('/repo');
+		await new Promise<void>(resolve => setTimeout(resolve, 0));
+
+		assert.deepStrictEqual(fetches, [['/repo', wipScope]]);
+	});
+
+	test('applyWipPayload does not refetch scope files when no mode is active', async () => {
+		const state = createDetailsState();
+		state.activeMode.set(null);
+		state.activeModeRepoPath.set('/repo');
+		state.scope.set(wipScope);
+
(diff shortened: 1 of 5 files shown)


ramin-t:
augment review

PR #5618, opened 2026-07-30
test(integrations): facade base-case trunk + typed fakeProvidersApi (reopens #5595)

IvanCorton:
Same five commits as #5595, byte-identical (tip `44d2b3a19`). #5595 auto-closed during a botched branch operation on my side while re-anchoring onto the moving `core` tip, and GitHub refuses to reopen a PR whose head was force-pushed after close; nothing about the content changed. Full description, verification (539 passing on the core it was rebased onto, plus non-vacuity inversions) and review context live in #5595.

Note: `core` has since moved (facade read-logic refactor + v0.4.100 bump). I will verify the suite on top of current `core` and re-anchor these commits in a follow-up push to THIS PR if anything drifts; treating that as part of review readiness.

### DIFF
--- a/packages/plus/integrations/src/__tests__/facadeBaseCases.test.ts
+++ b/packages/plus/integrations/src/__tests__/facadeBaseCases.test.ts
@@ -674,12 +674,13 @@ suite('ProviderBackend facade — base-case trunk (#5438, #5533, #5549)', () => 
 				// inline author/assignee mapping falls `url` back to `undefined` — a small but real divergence
 				// between the two normalizers worth pinning explicitly.
 				assignees: [{ id: 'rev-1', name: 'Trinity', avatarUrl: undefined, url: undefined }],
-				repository: { owner: 'octocat', repo: 'hello' },
+				repository: { id: 'repo-1', owner: 'octocat', repo: 'hello' },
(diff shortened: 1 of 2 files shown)


IvanCorton:
Re-anchored onto current core tip (ee5f83cf3): the 5 original commits cherry-pick clean (both files are pure additions), plus one reconciliation commit whose body carries the full ledger. Six pins had drifted over the month, every one traced to the intentional change that moved it before updating: cliUnsupported removed by de8aec9a8 (its own test dropped the same assertions), the Jira per-resource paged drain from the same commit, username on PR/issue members from 354cc88c5, and the new search/sort/reviewed filter tables from 5059cbf36 / cdb6ba97b / 5087cc9a9 / 9154096a7 / af6ca0a8f, cross-checked against models.ts and the installed provider-apis 0.57.0 rather than the test's own output. 18/18 passing; tsc -b over the packages, oxlint (both configs) and oxfmt clean.

One observation from the reconciliation, left out of the pins on purpose: toIssueShape never populates IssueMember.username for author/assignees, while the PR path and Azure's fromProviderIssue path do since 354cc88c5. Reads like a gap that commit missed rather than a per-provider choice, but happy to be told otherwise; can file it separately if it's real.

Issue #5450, opened 2026-07-06
Automatic rebase: resolve conflicts end-to-end with AI, then review a summary and undo


ianhattendorf:
**Summary**

Add an "automatic rebase" option that runs a rebase from start to finish, using AI to resolve conflicts at each paused step, then presents a reviewable summary of every conflict encountered and the strategy chosen — with a one-click undo.

**Background / Motivation**

GitLens can already start rebases and resolve the *current* set of conflicts with AI (Resolve mode, #5306). But on a multi-commit rebase the user must manually apply, stage, click Continue, and re-resolve at each paused step. There's no end-to-end automation, no consolidated summary of what the AI decided across the whole operation, and no way to undo a rebase once it has completed.

**Proposed capability (user-facing)**

- Kick off a rebase that proceeds automatically: at each conflicted step the AI resolves, stages, and continues without manual intervention.
- Keep resolutions consistent across steps when the same region conflicts repeatedly.
- Escalate rather than guess: if AI confidence is below a configurable threshold, or a non-conflict pause needs attention, stop and hand off to the existing Resolve panel, pre-populated with what was already resolved.
- Present an end-of-run summary: per file — strategy, confidence, and rationale — grouped by the commit/step where it occurred, with before/after diffs.
- Offer **Undo** on the summary to roll the branch back to its pre-rebase state (validated — refuses if the branch has moved since).

**Impact**

- Benefits users rebasing conflict-prone or long-lived branches, who today resolve conflicts one paused step at a time.
- Turns a repetitive, error-prone manual loop into a single reviewable action with a safety net.
- Builds directly on existing AI conflict resolution / Resolve mode and paused-operation handling.

**Validation**

- Automatic rebase across a branch that conflicts at multiple steps resolves, stages, and continues to completion without manual intervention.
- A low-confidence or failed resolution stops automation and opens the Resolve panel with prior resolutions preserved.
- An externally-modified working tree aborts automation instead of overwriting changes.
- The end-of-run summary lists every file's strategy/confidence/rationale with before/after diffs.
- Undo restores the original branch tip and refuses when the branch has moved on.
- Cancelling mid-run aborts cleanly; the autostash-conflict case is surfaced before offering undo.

**Risk / Regressions**

- Incorrect resolutions could silently corrupt later steps — mitigated by escalating on low confidence rather than falling back to a fixed strategy.
- Undo interacting with autostash pop / dirty working trees.
- Concurrent external repo changes during the loop.
[...]

PR #5623, opened 2026-07-30
Automatic rebase: progress in the Resolve panel, inline reasoning, refined rebase picker

ianhattendorf:
Follow-up to the automatic rebase (#5450), addressing three pieces of review feedback.

## Shows the run's progress in the Resolve panel (`c96d77ea`)

An automatic rebase ran entirely behind a cancellable progress notification, so mid-run you couldn't see which step was running, what had already been decided, or how far along it was — the _Commit Graph_'s **Resolve** panel only opened *after* automation escalated.

The panel is now the run's surface: it opens as the rebase starts and shows the step it's on, a progress bar, each already-resolved conflict (strategy, confidence, reasoning), and the live activity. An escalation becomes an in-place transition instead of a context switch. The notification is gone.

- Adds an `autoRebaseProgress` RPC event fed by the auto-rebase service's change event, mirroring the existing `resolveProgress`. The push channel already existed; nothing but the notification consumed it.
- Adds `AutoRebaseSession.current` so progress surfaces read the step number instead of parsing it out of `progressMessage`.
- Replaces the notification's single _Cancel_ with **Cancel Rebase** (abort + restore) and **Stop Automating**, which reaches the already-implemented `detach` mode — previously unreachable, since the notification hardcoded `abort`.

image

## Reasoning is always visible (`21d04909`)

The reasoning behind a resolution is the text you're being asked to audit, so it shouldn't cost a click. The collapsed _Why this resolution_ disclosure is replaced by the reasoning itself, clamped to three lines with a _…see more_ expander.

- Extracts a shared `renderReasoning()` + `measureReasoningOverflow()` into `resolveDisplay.ts`, replacing two divergent inline implementations.
- Clamps via a **new** class, not `.resolve-file__reasoning` — the panel also uses that for skipped-file, error, and rename/rename messages, none of which may truncate.
- Offers the expander only where text is actually clipped, measured from the laid-out DOM; low-confidence rows still start expanded.

image

## Rebase picker (`7bf6d7fd`)

The confirm step listed six items — three modes, each duplicated solely to carry `--update-refs`. It now offers **Rebase**, **Automatic Rebase**, **Interactive Rebase**, with an _Update Branches_ toggle applying `--update-refs` to whichever is chosen.

- Uses the existing `createToggleButton()` helper: native `toggle.checked` on VS Code >= 1.109, icon-swap fallback below it.
- Refreshes items in place on toggle rather than via `retry()`, which on a confirm step feeds `Directive.Noop` back into the generator and pops the wizard back to branch selection.
- Gates _Automatic Rebase_ on `behind > 0`, matching the plain rebase — it was previously offered with nothing to rebase onto, including when rebasing onto a commit already in HEAD's history (an interactive-only scenario).
- Renames the palette entries and Pro-gate messages to match. Inline contrastive labels are deliberately left alone (\"Continue with AI\" next to a plain \"Continue\") — there the contrast is the information.

image

## AI simulator support (`eb78d86f`)

`conflict-resolution` had no canned response, so it fell through to the generic summary/body payload — not a shape `@gitkraken/conflict-tools` parses, so every file failed and a run escalated on its first step. Exercising the happy path meant hand-authoring JSON against a contract that only exists inside the bundled, minified resolver prompt. Adds a file-level `theirs` default whose confidence clears the `0.8` `ai.autoRebase.confidenceThreshold` default.

## Verification

`check`, `build`, and unit tests (1585 passing) are clean. Beyond that, this was **exercised against a running instance** rather than reviewed statically — which is how the following were confirmed, with measurements:

- `detach` (**Stop Automating**) — its first ever run through any UI: rebase left paused at 3/3, both completed steps preserved and committed.
- `abort` (**Cancel Rebase**) — HEAD restored exactly to the pre-rebase tip, no rebase dir.
- Escalation consumes the handoff in place (the panel is already open, so the mode-entry that normally seeds it is a no-op re-entry).
- A finished session lingers in memory (`dismiss()` is never called), so the terminal view is scoped to runs the panel actually watched — verified both ways.
- Clamp measured at `scrollHeight` 127 vs `clientHeight` 55, correct `aria-expanded`/`aria-controls`.
- Picker: exactly three items in order, `Automatic Rebase` preselected when AI-seeded, toggle rewriting details in place without popping back a step; and only `Interactive Rebase` when rebasing onto a commit already in HEAD's history.
- Console clean in every state.


ianhattendorf:
augment review

Issue #5611, opened 2026-07-29
Excluding an untracked nested repository from a compose has no effect — its entry is still sent to the AI and still applied


ramin-t:
### Summary

In the Commit Graph's Compose panel, unchecking an untracked directory that is itself a git repository does not exclude it — its hunk still reaches the AI provider, and still lands in the applied commits. Normal untracked files exclude correctly. Same root cause as #5603, on a different surface.

### Details

The curation list and the exclusion set carry the path exactly as git reports it for an untracked embedded repository, ending in a slash (`nested-repo/`). Compose's hunks name the same entry without the slash (`nested-repo`, a gitlink entry). Every comparison is an exact string match, so the excluded row never matches.

The Compose and Review panels share one file list — both read the same `scopeFiles` resource (`gl-graph-details-panel.ts:2753`/`:3164`), so the Compose exclusion set has the identical trailing-slash shape #5603 describes.

Where the mismatch bites, all in `src/webviews/plus/graph/compose/integration.ts`:

- `:135` / `:150` — the `hunkFilter` used for plan generation. `userExcluded` holds `nested-repo/` while `ComposeHunk.fileName` is `nested-repo`, so the predicate never matches and the gitlink hunk goes into the prompt.
- `:340` / `:343` — the same filter rebuilt at apply time from `cached.excludedFiles`, so the excluded entry also survives into the commit that actually gets applied.
- `:390` — the `aiExcludedFiles` masking set on the per-commit message-regen path; content redaction misses for the same reason.
- `:505` / `:508` — `moveFilesBetweenCommits` matches `pathSet` against `h.fileName`, so moving that row between draft commits silently does nothing.

Why the hunk carries the slash-less name: Compose requests untracked content via `includeUntracked` (`compose/integration.ts:690`), and `@gitkraken/compose-tools` stages with `git add -A` into a scratch index (`GIT_INDEX_FILE`) before diffing `HEAD` against it. `git add -A` records an untracked embedded repository as a gitlink (mode `160000`), and the resulting `diff --git a/nested-repo b/nested-repo` header has no trailing slash — the same shape #5603 hit in the review diff.

### Impact

The same privacy-facing failure as #5603: a path the user explicitly excluded is still sent to the provider — the directory name and the embedded repository's commit SHA. No file contents leave the machine, because a gitlink carries none.

[...]

PR #5637, opened 2026-07-31
Names an untracked nested repository's row in Files Changed

### LINKED ISSUE #5603: Excluding an untracked nested repository from a review has no effect — its entry is still sent to the AI
### Summary

In the Commit Graph's Review panel, unchecking an untracked directory that is itself a git repository does not exclude it — its entry still reaches the AI provider. Normal untracked files exclude correctly.

### Details

The curation list and the exclusion set carry the path exactly as git reports it for an untracked embedded repository, which ends with a slash (`nested-repo/`). The assembled review diff names the same entry without the slash (`nested-repo`, a gitlink entry that appears because the review now stages untracked files with intent-to-add — see #5586 / #5599). The review's file filter compares the two by exact string, so the excluded row never matches and remains in the payload.

Verified by capturing the request that actually left the extension: with the row excluded the prompt still contained the nested-repository entry, and the payload length was identical to a run where nothing was excluded. Excluding a normal untracked file in the same working tree behaved correctly — that path and its contents disappeared from the payload and the payload shrank.

Only untracked embedded repositories are affected. Plain untracked directories are enumerated file by file, so their rows carry ordinary file paths and exclude correctly.

### Impact

A user who explicitly excludes a path from an AI review still has that path sent to the provider — the directory name and the embedded repository's commit SHA. No file contents leave the machine in this case, because a gitlink carries none, but the exclusion control silently does nothing, which undermines trust in a privacy-facing affordance. The mismatch is systematic rather than timing dependent, so it reproduces on every run. Low severity.

### Validation

1. In a repository with some working changes, create an untracked directory that is itself a git repository (`git init nested-repo`, then commit something inside it).
2. Select the Working Changes row in the Commit Graph and open the Review panel with a scope that includes unstaged changes — `nested-repo` is listed as an untracked (U) row.
3. Uncheck it in Files Changed, leave at least one other row included, and start the review.
4. Expected: the nested repository is not part of the reviewed content. Actual: its entry is still included.

### Suggested fix

Normalize trailing slashes on both sides when filtering the review diff by excluded paths, or drop directory-shaped entries from the review content altogether, since a gitlink is not reviewable in the first place.

### Notes

Live-verified on `main` (429fca690 / #5599) with the outgoing request payload captured. Found while verifying #5586.

Related but distinct from #5588: there the row is missing from the curation list entirely, so it cannot be unchecked; here the row is present and explicitly excluded, and the reconcile added in #5601 does not touch path comparison — this reproduces on current `main`, after that fix.

### LINKED ISSUE #5629: An untracked nested repository's row in Files Changed loses its name in the checkbox tooltip and label
## Summary

In the Files Changed list of the Commit Graph's Review and Compose panels, the row for an untracked directory that is itself a git repository loses its name: its checkbox tooltip reads `Exclude ` / `Include ` with nothing after the verb. Git reports such an entry with a trailing slash (`nested-repo/`), and the row's display strings are derived by taking the basename after the last `/`, which for that path is the empty string.

Clicking the row also does nothing at all — no editor tab, no message.

## Impact

Low. The control still works (unchecking it does exclude the entry, since #5603 / #5611 normalized the comparisons), but the affordance that describes it is nameless — including for screen-reader users, since the tooltip is the checkbox's accessible description. This is the same trailing-slash root as #5603 / #5611: those fixed the comparisons, not the display derivation.

## Validation

1. In a repository with some working changes, create an untracked directory that is itself a git repository (`git init nested-repo`, then commit something inside it).
2. Select the Working Changes row in the Commit Graph and open the Review panel with a scope that includes unstaged changes.
3. Hover the checkbox of the `nested-repo` row.
4. Expected: `Exclude nested-repo`. Actual: `Exclude ` — the name is missing. Unchecking it flips the tooltip to `Include `, equally nameless, while the neighbouring rows read `Exclude file1.txt` / `Exclude new-untracked.txt`.
5. Click the row body: nothing opens.
6. Repeat in the Compose panel — same rows, same tooltip (both panels share one file list).

## Evidence (live, `main` @ `779dac1db`, fresh build, subscription simulator Pro + AI simulator)

Review panel, three rows:

```
{ txt: 'nested-repo U',        tip: 'Exclude ',                  checked: true }
{ txt: 'file1.txt M',          tip: 'Exclude file1.txt',         checked: true }
{ txt: 'new-untracked.txt U',  tip: 'Exclude new-untracked.txt', checked: true }
```

After unchecking the first row: `tip: 'Include '`. Compose panel: identical three rows, identical tooltip. Clicking the row body left `vscode.window.tabGroups` empty with nothing in the logs.

## Code

`src/webviews/apps/shared/components/tree/gl-file-tree-pane.ts:831-832` —

```ts
const pathIndex = file.path.lastIndexOf(glue);
const fileName = pathIndex !== -1 ? file.path.substring(pathIndex + 1) : file.path;
const filePath = flat && pathIndex !== -1 ? file.path.substring(0, pathIndex) : '';
```

For `nested-repo/` this yields `fileName: ''` and `filePath: 'nested-repo'`. `fileName` then feeds the checkbox tooltip (`:861-868`, live-confirmed empty), the file icon's `filename` (`:884`) and the row's `label` (`:901`) — so in list layout the name the user sees is actually the `description` (the directory part), with an empty label beside it. In tree layout (more files than the layout threshold) the same path should instead produce a folder node with an empty-labeled leaf inside it; that variant is derived from the code, not observed.

## Suggested fix

Derive the display name from the path with a trailing separator trimmed first — `normalizePath` is already what the comparison sites use for exactly this shape — so a directory-shaped entry reads as `nested-repo`. Deciding what a click on such a row should do (nothing, reveal in Explorer, or no click target at all) is worth settling at the same time, since a gitlink has no diff to open.

## Notes

Found while verifying #5603 / #5611 on `main` (`779dac1db`).

Related: #5603, #5611.

ramin-t:
Closes #5629

## What the user saw

In the Files Changed list of the Commit Graph's Review and Compose panels, the row for an untracked directory that is itself a git repository had no name. Its checkbox tooltip read `Exclude ` / `Include ` with nothing after the verb, and its label was empty — so the name the user actually saw was the row's *description* (the directory part), with an empty label beside it. The neighbouring rows read `Exclude file1.txt` / `Exclude new-untracked.txt` normally. Clicking the row did nothing at all.

## Why it happened

Git cannot recurse past an embedded repository boundary, so it reports such a directory as a single entry with a **trailing slash** — `nested-repo/`. Every display string was derived by taking the basename after the last `/`:

```ts
const pathIndex = file.path.lastIndexOf(glue);
const fileName = pathIndex !== -1 ? file.path.substring(pathIndex + 1) : file.path;
const filePath = flat && pathIndex !== -1 ? file.path.substring(0, pathIndex) : '';
```

For `nested-repo/` the last `/` is the final character, so `fileName` is `''` and `filePath` is `'nested-repo'` — the two are effectively swapped. That single `fileName` feeds the checkbox tooltip, the file icon's `filename`, and the row `label`, which is why one root cause produced three symptoms at once.

Status is fetched with bare `-u` (`packages/git-cli/src/providers/status.ts:259-264`), which means `all` — ordinary untracked directories *are* recursed into. I verified this against real git with those exact flags:

```
?? nested-repo/          ← nested repo: trailing slash
?? plain-dir/sub/a.txt   ← ordinary untracked dir: fully recursed
```

So a trailing slash identifies the nested-repository case unambiguously, which is what makes the tooltip wording below accurate rather than a guess.

The tree-layout variant has the same root but a different shape: splitting `nested-repo/` on `/` yields a trailing empty segment, so `makeHierarchical` nests an empty-named leaf under a folder of the same name instead of producing one leaf. Compaction does not rescue it — `compactHierarchy` will not merge a folder into a value-bearing child without a `canCompact` predicate.

## How it's resolved

A `trimTrailingSlash` helper in `file-tree-utils.ts`, applied at every site that derives a display name or a tree path:

| Site | Symptom it caused |
| --- | --- |
| `gl-file-tree-pane.ts` `fileToTreeModel` | Empty label, empty checkbox tooltip, file icon fed an empty filename — Review, Compose, Compare and WIP all route through this one component |
| `file-tree-utils.ts` `buildFileTree` | Tree layout nesting an empty-named leaf under a folder |
| `gl-tree-base.ts` (patch details) | Both of the above, in its own copy of the derivation |
| `gl-draft-details.ts` | Defense-in-depth only — see below |

### Deliberately not `normalizePath`

The issue suggested `normalizePath`, and the sibling fixes #5603/#5611 used it. **That precedent does not transfer here**, and this is the one design decision in the PR worth a reviewer's attention.

`normalizePath` rewrites backslashes to `/` unconditionally (`packages/utils/src/path.ts:76`, outside any platform guard). Those sibling fixes normalize *both sides of a comparison*, so the rewrite is symmetric and cancels out. This change derives a **display name and a tree structure** from one side, where nothing cancels. A backslash in a git path is a literal character in a POSIX-legal filename, so `weird\name.ts` would have been relabeled `name.ts` in list layout and split into a folder `weird` containing `name.ts` in tree layout — a regression the pre-existing code did not have, introduced while fixing a display bug. Two tests pin this behavior so the next person doesn't reach for `normalizePath` either.

## Behavior changes

- **Row hover tooltip.** A trailing-slash entry is now marked `(nested repository)`, alongside the existing `(submodule)` marker, and its displayed full path no longer carries the trailing slash. The entry is an untracked directory, not a file, so there is nothing to open or diff — which is why clicking the row does nothing; naming it is what tells the user that, rather than adding click plumbing for an entry with nothing to show. It is not a gitlink at this point: an untracked nested repository is absent from the index entirely (`git status` reports only `?? nested-repo/`, no mode, no SHA) and becomes one only once staged. Clicking remains inert — deliberately.
- **`glue` parameter removed** from both `fileToTreeModel` methods. No call site has ever passed it. Leaving it half-honored — the trim hardcoding `/` while the split still respected the parameter — would have been more misleading than the dead generality it replaced.
- The row's icon is unchanged; it still renders the generic file icon.

## Tests

New `packages/utils/src/__tests__/path.test.ts` covering `trimTrailingSlash` and — because this is the combination that produced the empty-named leaf — `makeHierarchical` composed with it, which had no tests of its own. Includes the ordinary-path and backslash cases so the fix cannot silently over-reach.

`trimTrailingSlash` lives in `packages/utils/src/path.ts` rather than beside its callers because a first attempt to test it in the webview tree hit a hard constraint. `file-tree-utils.ts` carries a side-effect import of `action-chip.js`, which transitively registers the `code-icon` custom element. Each test entry is bundled separately with no code splitting, so a second test importing that module gets its own copy of the registration, and the second `customElements.define` against the one global registry throws — which is exactly how CI failed on the first push. Splitting the bundles would need esbuild's `splitting`, which requires ESM output the CJS test host cannot load. The helper is a pure path operation with no business in a module that registers components, so moving it both fixes the collision and puts it beside the `normalizePath` it is deliberately contrasted with.

**Not covered, stated plainly:**

- `buildFileTree` and `buildFileTooltip` have no direct tests, for the bundling reason above. The first is a thin wrapper over the composition now tested directly; the second appends a fixed marker to a string.
- The `fileToTreeModel` derivation is a private/protected Lit component method, so the label, icon and checkbox-tooltip path has no unit test. Testing it would need a component-instantiation harness this repo does not have for these components, and I did not add one.
- The tree-layout variant remains derived from the code, not observed live — as the issue itself noted. The unit test is what covers it.

## Validation

1. In a repository with some working changes, create an untracked directory that is itself a git repository: `git init nested-repo`, then commit something inside it.
2. Select the Working Changes row in the Commit Graph and open the Review panel with a scope that includes unstaged changes.
3. Hover the checkbox of the `nested-repo` row → expect `Exclude nested-repo` (was `Exclude `).
4. Uncheck it → expect `Include nested-repo` (was `Include `).
5. Hover the row body → expect the tooltip to name it `(nested repository)`.
6. Repeat in the Compose panel — same rows, same tooltip, since both panels share one file list.
7. With more files than the layout threshold (tree layout), confirm `nested-repo` appears as a single leaf rather than a folder containing an empty-named row.

## Related

- #5603 / #5611 — same trailing-slash root cause, fixed at the comparison sites rather than the display derivation.
- `gl-draft-details.ts` is included for consistency, **not** as a reachable bug: its files come from `getDiffFiles(contents)` (`draftsService.ts:175-177,574-575`), parsed from patch text where a gitlink is named without a trailing slash. It was the one `makeHierarchical` site not trimming, and leaving it as the odd one out invites drift.
- `rebase.ts` and `gl-graph-details-panel.ts` have structurally similar path splitting but operate on merge-conflict entries, which an untracked directory cannot be. Deliberately not touched.


ramin-t:
augment review

PR #5647, opened 2026-08-04
Adds repo consultation to AI conflict resolution — closes #5581

### LINKED ISSUE #5581: AI conflict resolution can't consult the repo when a hunk is ambiguous
**Summary**

AI conflict resolution decides from a single request per conflict — it can't read the rest of the file, inspect either side's history, or search the repo when the hunk alone is ambiguous. Give it read-only tool use so it can gather that context before choosing a strategy.

**Background / Motivation**

- #5306 introduced AI resolution (Resolve mode); #5450 automated it across a whole rebase. Both send one request per conflict and take the answer as final.
- `@gitkraken/conflict-tools` already implements six read-only tools — `show_file_at_ref`, `blame`, `grep`, `log`, `show`, `diff` — a bounded step loop, and per-tool output caps. Each falls back to the plain git exec GitLens already supplies, so the tools need no new git plumbing.
- The gap is model-side: GitLens advertises no tool definitions and never returns tool calls, so the loop never engages. Tool-result messages are flattened to plain text on the way in.
- Consequence: ambiguous hunks are judged from local text only. Low-confidence cases escalate to the user by design — but many escalations are answerable from context the model was never allowed to fetch (is the renamed symbol still referenced elsewhere, which side is newer, why a line exists).

**Proposed capability (user-facing)**

- While resolving, AI can consult the repository: read the whole file or either side at its ref, blame/log the lines in question, search for other uses of the symbols involved, diff the two sides.
- Rationale cites the evidence consulted, so the per-file "why this resolution" in Resolve mode and the automatic-rebase summary explains reasoning, not just the verdict.
- Read-only and bounded: a step cap per conflict, scoped to the repository, no working-tree writes — resolution still lands only through the existing apply path.
- Better confidence should mean fewer escalations in automatic rebase; genuinely uncertain cases still escalate rather than guess.

**Dependencies / constraints**

- GitLens's AI request path has no tool-call round-trip today (tool support exists only in the agents layer). This needs tool definitions and multi-turn tool results plumbed through that path, or conflict resolution routed through the agent layer.
- Provider coverage is uneven — BYOK models and VS Code's model API don't uniformly support tools — so it needs a capability check and a graceful fall back to today's single-shot behavior.
- Cost and latency per conflict rise; the step cap and the library's output caps are what keep that bounded. Automatic rebase resolves a step's files in parallel, multiplying tool traffic.

**Impact**

- Fewer escalations and fewer confidently-wrong resolutions on conflicts needing context beyond the hunk — the main quality lever for both Resolve mode and automatic rebase.
- Makes the summary review in #5450 auditable: what the AI looked at, not just what it chose.

**Validation**

- A conflict whose correct resolution depends on code outside the hunk resolves correctly, and the rationale names what was consulted.
- Tool calls stay within the step cap; exhausting it escalates instead of looping.
- A provider without tool support still completes resolution on the existing path.
- Cancelling mid-tool-call aborts cleanly; no working-tree writes originate from tool use.
- Token usage and duration per conflict stay reported, with BYOK usage still attributed to the run.
- AI file-exclusion rules apply to everything tools read, not just the conflicted files.

**Risk / Regressions**

- Runaway tool loops inflating cost and latency without a hard cap.
- Wider context invites edits beyond the conflict region; the existing verifier and the apply-time unmerged-path recheck are the backstop.
- Sensitive content reaching the provider expands from the hunk to whatever tools read.

**Out of scope (possible follow-ons)**

- Write-capable tools (staging, checkout-side) — the library exposes them; this issue stays read-only.
- Verifying resolutions by building or running tests.

Builds on #5306 and #5450. Related: #4444.

ianhattendorf:
# Description

Closes #5581.

AI conflict resolution decided each conflict from a single request, so an ambiguous hunk couldn't be checked against the rest of the file, either side's history, or the wider repo — it either guessed or escalated. Many escalations were answerable from context the model was never allowed to fetch.

`@gitkraken/conflict-tools` already implements the whole agentic loop — six read-only tools (`show_file_at_ref`, `grep`, `blame`, `diff`, `log`, `show`), a bounded step loop, per-tool output caps, and a git-exec fallback for each. **None of it ran.** `createAiModelPort` dropped `params.tools`, flattened tool results to user text, discarded assistant turns carrying only `toolCalls`, and never returned `toolCalls` — so the loop's tool branch was unreachable while the library's own system prompt was already telling the model to use those tools.

## What changed

**Tool calls through the AI provider path.** `AIToolDefinition`/`AIToolCall`, a `'tool'` role, and optional `toolCalls`/`toolCallId`/`toolName` on `AIChatMessage`. The generic's default role union keeps all existing call sites unchanged. OpenAI wire shape lives in `OpenAICompatibleProviderBase` behind hooks alongside the existing `extractSystemPrompt` precedent; Anthropic overrides for `input_schema`/`tool_use`/`tool_result` and batches consecutive tool results into one user turn. The VS Code provider moves from `rsp.text` to `rsp.stream` — `text` yields only text parts and would silently drop tool calls.

**Two safety properties keep this low-risk:**

- Only providers that declare `supportsTools` ever receive tool-shaped messages; everything else keeps today's single-request path. Ollama is opted out explicitly (it overrides `fetch` wholesale), as are the user-supplied-endpoint providers.
- The library already degrades gracefully — a provider that ignores `tools` and returns text just gets parsed, exactly as before. The only hard failure is a provider *rejecting* the field, handled by a one-shot retry without tools that then latches for the session.

**Also fixes the library's Phase 2 refine pass**, which needs tool calls to apply follow-up edits. Without them it burned a request, applied nothing, and had its confidence capped at 0.6 — below the 0.8 automatic-rebase threshold — so every resolution emitting follow-ups always escalated.

**AI file-exclusion rules now apply to everything tools read**, which previously governed nothing in conflict resolution. The `ConflictGitPort` is the only boundary between the model and the repository, so the six read ops are supplied there. The files being resolved are exempt — the library reads them internally for extraction and the prompt's three-way diff, and blinding those while still sending the conflict markers would degrade the resolution without protecting anything. That same shared op is why only *unscoped* diffs are capped: a path-scoped `diff` is the prompt's primary evidence and must not be truncated.

## Notes for review

Two GitKraken proxy behaviours are load-bearing and worth a second pair of eyes:

- Tool calls may carry an opaque `thought_signature` that must be echoed back on the next assistant turn or Anthropic rejects the tool result (Gemini has a Redis fallback; Anthropic does not). The library's `ToolCall` has nowhere to hold it, so signatures are stashed by call id on the port instance and re-attached on replay.
- The proxy rejects a tools request without `GK-Conversation-ID` when it routes to Gemini — its default when no model is pinned — so `conversationId` is now required rather than optional.

`maxSteps` stays at the library default of 15 deliberately: the resolver spends that budget on tool steps *and* on re-prompts after a validation failure, so lowering it converts recoverable retries into hard failures and *more* escalations.

## Verification

Beyond unit tests, this was exercised live against a running instance (both the manual **Resolve** panel and an automatic rebase), confirming: tools advertised, real `git grep`/`blame`/`log` output reaching the model, the empty-content assistant turn surviving, correct `toolCallId` pairing, the `thought_signature` round-trip, `.aiignore`-sourced exclusion blocking a tool read, a parallel multi-call round, and both progress messages rendering in the DOM.

Still unverified and worth knowing: resolution *quality* against a real model (the simulator returns canned answers), and the proxy accepting the payload end-to-end against an Anthropic-backed model specifically — an OpenAI-backed one would pass while leaving the `thought_signature` round-trip broken.

The second commit adds tool-call support to the DEBUG-only AI simulator. Without it the feature is unreachable there and cannot be exercised deterministically at all.

# Checklist


ianhattendorf:
augment review

PR #5663, opened 2026-08-05
Fix comparison diffs always opening at the top of the file

MaEtUgR:
## Problem
Every file opened from a Search & Compare comparison result opens at the very top instead of the first difference. This is very annoying when reviewing differences.

## Solution
Suggested by Claude Code: Dropping the hardcoded range restores the default reveal-first-change behavior for comparison diffs.

ResultsFileNode.getCommand() hardcoded range: editorLineToDiffRange(0). That range flows into DiffWithCommand as an explicit editor selection, which causes vscode.diff() to place the cursor at the very top of the file instead of letting the diff editor reveal the first change, as it does for every other diff entry point.


eamodio:
Thank you for your contribution!

Thank you

Issue #5630, opened 2026-07-31
Review exclusions are applied after the full working-tree diff is built, and the diff cache is keyed on un-normalized paths


nikolay-1986:
## Summary

A review of working changes builds the **whole** working-tree diff first and drops the user's excluded paths from the resulting diff text afterwards. Excluded content is therefore read from disk and materialized in the extension's memory before being discarded, and every untracked entry — excluded or not — is staged into the scratch index on every run.

Separately, the diff cache that keys on that exclusion set uses the raw paths while the filters now compare normalized ones, so two spellings of the same logical exclusion occupy two cache entries.

## Details

**1. Exclusions are applied post-diff.** In `getDiffForScope` the `includeUnstaged` branch enumerates every untracked path (`graphInspectServices.ts:2845`), stages all of them intent-to-add into the scratch index (`:2860`), and diffs with no pathspec (`:2881`). The exclusion set is only consulted later, by `filterDiffFiles` on the assembled text (`:780-790` for the single-pass review, `:920-926` for the two-pass focus-area request).

Consequences, in rough order of importance:

- The contents of a file the user explicitly excluded are still read and held in the diff string for the duration of the request. Nothing leaves the machine — the filter runs before the prompt is built — but for a privacy-facing control the natural expectation is that excluded content is never collected in the first place. It matters more for the case users reach for this control for: a large or sensitive untracked file.
- The work is wasted: staging + diffing content that is guaranteed to be thrown away. It scales with the size of what was excluded.
- Excluding an untracked nested repository still runs `git add -N -- nested-repo/ …` on every review, so git's `warning: adding embedded git repository: nested-repo` (plus its two `hint:` lines) is logged on every run even though the entry is about to be filtered out.

A pathspec-limited diff, or filtering the enumerated untracked list against the exclusion set before staging, would avoid all three. The exclusion set is available at the call site — it is already a parameter of the surrounding `reviewChanges`.

[...]

PR #5673, opened 2026-08-06
Fixes the wrong change types and inflated line counts a two-pass review's first pass ranks from

### LINKED ISSUE #5658: Two-pass review's first pass is given wrong change types and inflated line counts to rank focus areas from
### Summary

The first pass of a two-pass review sends the AI a file manifest and asks it to rank what deserves closer inspection from the change types and line counts in it. Those numbers are wrong: every file is reported as modified — including newly added ones — and the counts include context lines, so they overstate what actually changed.

### Impact

Two-pass is the strategy used for exactly the large change sets where the ranking matters, and pass 1 has nothing but the manifest to go on: no diff content, only paths, change types and counts. Feeding it inflated counts and a uniform "modified" biases which areas it flags and in what order, and the user only ever sees the outcome — the focus areas — with no way to tell the input was off. A newly added file described as modified also invites findings phrased as if there were a previous version to compare against.

Observed live on a working-changes review of one modified tracked file and one new untracked file:

```
[{"path":"file1.txt","status":"M","additions":6006,"deletions":4},
 {"path":"new-untracked-5630.txt","status":"M","additions":1,"deletions":1}]
```

What git reports for the same tree: `file1.txt` **+6004 / −1**, and `new-untracked-5630.txt` is a **new** one-line file with nothing deleted. So the modified file's counts are inflated (the hunk's context lines are counted as changes) and the added file is both mistyped and credited with a deletion that does not exist.

Not related to #5630 — that change moved when exclusions are applied and did not touch how the manifest is built. Found while verifying it.

### Validation

1. In a repository, make a change large enough that a review takes the two-pass path (a few thousand added lines in a tracked file), and add one new untracked file.
2. Select Working Changes in the Commit Graph, open the Review panel with a scope including unstaged changes, and start the review. The panel offers "Review Files" per focus area, which is the two-pass tell.
3. Compare the manifest pass 1 was given against `git diff --numstat` for the same tree, and against the actual status of each file.
4. Expected: the count per file matches what git reports for it, and an added file is reported as added rather than modified. Actual: counts are larger than git's, and every file reads as modified.
5. Confirm the second pass is unaffected — it works from real diff content, so its findings and line citations should stay correct either way.

### Risk

The counts feed the pass-1 ranking, so correcting them changes which focus areas come back and in what order for large change sets — worth eyeballing a couple of real reviews before and after rather than only a synthetic one. The same counting helper is used elsewhere to size and summarize diffs, so a change to it needs those callers checked: anything that only compares magnitudes will be unaffected, anything that displays a number to the user will shift. Renames, binary files, and deleted files are the shapes most likely to expose an off-by-one in a fix.

### Environment

- GitLens: dev build of `main` @ 4b7c5fb52 (would ship in the 19.0 line), found while verifying #5630

Related: #5630.

ramin-t:
Closes #5658

## What the user saw

The first pass of a two-pass review ranks focus areas from a JSON file manifest alone — no diff content, only paths, change types and line counts. Every file in that manifest read as modified, including newly added ones, and the counts overstated what actually changed. On a working-changes review of one modified tracked file plus one new untracked file:

```json
[{"path":"file1.txt","status":"M","additions":6006,"deletions":4},
 {"path":"new-untracked-5630.txt","status":"M","additions":1,"deletions":1}]
```

Git reported `+6004/-1` for the first file, and the second was a new one-line file with nothing deleted.

## Why it happened

Two independent causes, both in how the manifest was assembled.

**The counts came from the hunk headers.** `countDiffInsertionsAndDeletions` summed `hunk.current.count` and `hunk.previous.count` — the totals in `@@ -a,b +c,d @@`, which count *every* line on each side, unchanged context included. A hunk carrying three context lines around one edit therefore reported four additions. Scaled up, a file with a few thousand added lines overstated its additions by however much context its hunks happened to carry.

The same arithmetic produced the phantom deletion. A new file's header is `@@ -0,0 +1 @@`, and `parseHunkHeaderPart` computes its count as `Number(countS) || 1` — so the zero on the old side became a **1**, and an added file was credited with a deletion that does not exist.

**The status was hardcoded.** The manifest built each entry as `{ path, status: 'M', additions, deletions }`. `parseGitDiff` had already determined the real per-file status and it was simply discarded, so an added, deleted, renamed or copied file all reported as modified — which, as the issue notes, invites pass 1 to phrase findings as though a previous version existed to compare against.

## How it's resolved

| Change | Symptom it fixes |
| --- | --- |
| `countDiffInsertionsAndDeletions` counts lines in each hunk's content that begin with `+`/`-` | Context-inflated counts, and the phantom deletion on added files |
| The manifest emits `f.status` instead of `'M'` | Every file reading as modified |

Nothing else changes shape — the manifest still carries exactly `path`, `status`, `additions`, `deletions`, and the pass-1 prompt is untouched. The status letters the manifest now emits (`A`/`M`/`D`/`R`/`C`/`T` from `GitFileIndexStatus`) are the ones git itself uses, and the template already tells the model to rank from "change types and line counts".

Counting off each hunk's content rather than the parsed `hunk.lines` map is deliberate: that map is keyed by file line number and merges a removed/added pair into one `changed` entry, so on an asymmetric block a following context line overwrites the tail of it. Counted from the map, a 3-removed/1-added block reports `+1/-2` where git reports `+1/-3`.

The content is scanned in place rather than split into lines, since splitting a multi-megabyte diff allocates a string per line for a single pass that only ever looks at each line's first character. Measured on a 576 KB, 97-file diff that is ~0.40 ms versus ~0.76 ms and allocates nothing. Neither number is material — this runs once per review, after a `parseGitDiff` it can never skip (~6 ms for the same diff) and before a request that takes seconds — it is just the cheaper option where it costs nothing.

`countDiffLines` is deliberately left alone. Its only caller compares magnitudes against thresholds (`>500`, `>200`, `<50`) that were tuned to context-inclusive numbers, so "fixing" it would silently retune truncation scoring rather than repair anything. Its doc comment now says so.

The other existing caller of the corrected helper, `createFallbackDiffSummary`, prints `+N/-M` per file and inherits the fix. Its numbers shift, correctly — nothing re-parses them, and since accurate counts are at most the old inflated ones, its character-budget check cannot newly fail.

## Behavior changes

**Focus-area ranking will differ for large change sets.** The counts feed the pass-1 ranking, so correcting them changes which areas come back and in what order. That is the point of the fix, but it is a visible change in output, not only a repaired input.

**Pass 2 is unaffected** — it works from real diff content, so its findings and line citations are unchanged either way.

One pre-existing wrinkle that truthful statuses make visible, disclosed so it can be judged rather than discovered: `getDiffForScope` concatenates the unstaged, staged and per-commit diffs, and `parseGitDiff` emits one entry per `diff --git` boundary with no dedup by path. A file touched in both an included commit and the working tree therefore yields two manifest entries. Previously both read `M`; now one can read `A` and the other `M` for the same path. Each entry is individually accurate for its slice, and the duplication mechanism is untouched here. Closing it properly means either deduping by path — which requires deciding what a single status means for a file added in a commit and then modified unstaged — or labelling each entry with its source, so it is left for a follow-up rather than decided incidentally in a bug fix.

## Tests

- 7 new cases in `packages/git/src/parsers/__tests__/diffParser.test.ts` covering the shapes most likely to expose an off-by-one: asymmetric change blocks, added and deleted files, pure renames and binary files (both `0/0`), the `\ No newline at end of file` marker, and an annotated diff, since the manifest parses the `[NNNNN]`-annotated text rather than raw git output.
- Two pre-existing assertions in each diffParser suite encoded the buggy header-count expectations (`4/3` and `5/5`) and were retargeted to what git reports.
- A manifest-level regression test in `src/webviews/plus/graph/__tests__/graphInspectServices.test.ts` forces the two-pass path and asserts the manifest across a modified, added, renamed and binary file together — the rename reports its new path with status `R`, the added file `A` with no deletions, and the binary file `M` with `0/0`.

**Not covered, stated plainly:** the manifest-level test has never been executed locally — `vscode-test` will not run while other VS Code instances are open, so it is type-checked and linted only and CI is its first real run. The parser suites do run (53 and 20 passing).

Separately worth knowing: `pnpm run test:packages` reports success while running zero tests on Windows — the recursive filter matches no projects and the packages' own single-quoted globs reach `cmd` literally. I ran the suites directly instead and left the scripts alone.

## Validation

1. In a scratch repository, commit a file, then add a few thousand lines to it and delete one. Add one new untracked file with a single line.
2. `git add -N <untracked>` (what the review does internally so untracked files appear in the unstaged diff), then capture `git diff`.
3. Compare the manifest built from that diff against `git diff --numstat` and `git diff --name-status`.

Run against that tree, the old code produces:

```json
[{"path":"file1.txt","status":"M","additions":6007,"deletions":4},
 {"path":"new-untracked-5658.txt","status":"M","additions":1,"deletions":1}]
```

reproducing the issue, including the bogus `1/1`. This branch produces:

```json
[{"path":"file1.txt","status":"M","additions":6004,"deletions":1},
 {"path":"new-untracked-5658.txt","status":"A","additions":1,"deletions":0}]
```

matching git exactly.

Also verified against a real 97-file commit in this repository (83 modified, 6 renamed, 6 deleted, 2 added): every count matches `git diff --numstat` and every status matches `--name-status`.

The issue also suggests eyeballing a couple of real reviews before and after, since the ranking shifts. **That has not been done** — it needs a live instance and real model calls.

## Related

- #5630 — adjacent, not the cause; that change moved when exclusions are applied and did not touch how the manifest is built.
- Not fixed here: under `diff.noprefix=true`, `diffRegex` fails to match and `parseGitDiff` silently skips the file, dropping it from the manifest entirely. Every git invocation forces `diff.mnemonicPrefix=false` and `core.quotepath=false` but not `diff.noprefix=false`. The regex and that loop are unchanged by this branch, so it does not own the problem, but a silently omitted file is worth its own issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,6 +59,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 - Fixes searching by author `@username` in GitHub virtual repositories dropping the first character of the username
 - Fixes searching by author or committer `@me` in GitHub virtual repositories skipping the filter entirely for accounts without a display name
 - Fixes the _Commit Graph_'s search history reading and saving under the first-opened repository after switching repositories in a multi-repo workspace
+- Fixes the _Commit Graph_'s _Review_ mode choosing focus areas from wrong file details on large change sets &mdash; newly added files counted as modified, and changed-line counts included unchanged context ([#5658](https://github.com/gitkraken/vscode-gitlens/issues/5658))
 
 ## [19.0.1] - 2026-08-13
 
--- a/packages/git-cli/src/parsers/__tests__/diffParser.test.ts
+++ b/packages/git-cli/src/parsers/__tests__/diffParser.test.ts
@@ -309,9 +309,9 @@ suite('countDiffInsertionsAndDeletions Test Suite', () => {
 		assert.strictEqual(parsed.files.length, 1);
 
 		const { insertions, deletions } = countDiffInsertionsAndDeletions(parsed.files[0]);
-		// previous: 3 lines (deletions context), current: 4 lines (insertions context)
-		assert.strictEqual(insertions, 4, 'Should count current lines as insertions');
-		assert.strictEqual(deletions, 3, 'Should count previous lines as deletions');
+		// One added line, three unchanged context lines that must not be counted
+		assert.strictEqual(insertions, 1, 'Should count only added lines as insertions');
+		assert.strictEqual(deletions, 0, 'Should not count context lines as deletions');
 	});
 
 	test('returns zeros for file without hunks', () => {
@@ -345,9 +345,8 @@ rename to new.ts`;
 		assert.strictEqual(parsed.files.length, 1);
 
 		const { insertions, deletions } = countDiffInsertionsAndDeletions(parsed.files[0]);
-		// First hunk: prev 2, curr 3
-		// Second hunk: prev 3, curr 2
-		assert.strictEqual(insertions, 5, 'Should sum current lines from all hunks');
-		assert.strictEqual(deletions, 5, 'Should sum previous lines from all hunks');
+		// First hunk: one added line; second hunk: one removed line
+		assert.strictEqual(insertions, 1, 'Should sum added lines from all hunks');
+		assert.strictEqual(deletions, 1, 'Should sum removed lines from all hunks');
 	});
 });
--- a/packages/git/src/parsers/__tests__/diffParser.test.ts
+++ b/packages/git/src/parsers/__tests__/diffParser.test.ts
@@ -761,7 +761,7 @@ suite('Diff Parser Test Suite', () => {
 	});
 
 	suite('countDiffInsertionsAndDeletions', () => {
-		test('counts insertions and deletions from hunks', () => {
+		test('counts insertions and deletions from hunks, excluding context lines', () => {
 			const data = [
 				'diff --git a/src/foo.ts b/src/foo.ts',
 				'index abc1234..def5678 100644',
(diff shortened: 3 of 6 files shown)


ramin-t:
augment review

PR #5683, opened 2026-08-07
Adds a one-time out-of-window Pro trial auto-reset on Graph open

sergeibbb:
Implements gitkraken/vscode-gitlens-private#97 (milestone 18.4).

When a signed-in user opens the Commit Graph, GitLens asks the server (`GET user/trial-eligibility`) whether the account may reset its Pro trial outside the standard 90-day window, and if so performs the reset (`POST user/reactivate-trial`), refreshes the subscription, and shows a reactivation toast.

**Behavior**
- Gated on the server's `canResetTrial` — never on local subscription state. Active trials are skipped without settling and re-checked once they lapse (the server never approves a reset during an active trial).
- One attempt per account **per machine** — the `plus:trialReset:${accountId}:attempted` flag is deliberately excluded from settings sync and from `Storage.reset`.
- Paid accounts settle immediately without any requests; a 409 from reactivation (paid-org members) settles as `refused`.
- The toast only fires once the client actually sees the new trial, so a failed post-reset check-in can't pair a success message with a still-gated UI — mirrors the guard on the manual reactivation flow.
- Fails open: server errors and unrecognized payload shapes never consume the attempt. Retries are capped at 3 per session with a 5-minute cooldown; a request that never leaves the client (blocked or unauthenticated) keeps the cooldown without spending a try, while a wire timeout counts against the budget.
- Both requests are pinned to the captured session token, so an account switch mid-flight can't reset the wrong account.
- Telemetry: `subscription/action` with `action: 'auto-reset-trial'` and `outcome: reset | not-eligible | refused | failed | failed-shape`. `failed-shape` is kept separate so a broken payload contract — which disables the promo for everyone — isn't read as an ordinary outage.

**Removal** — this is a launch promo; `src/plus/gk/trialAutoReset.ts` starts with a checklist of everything to delete when it ends, including the one piece that stays (the Graph's Pro-access-flip rebuild, which also covers upgrades and manual reactivation).

Verified live end-to-end on a test account against dev-api: eligibility → reset → fresh 14-day trial → Graph un-gated without reload → toast; repeat opens are no-ops (`resetCount 1/1`). Covered by 18 behavioral unit tests. No CHANGELOG entry on purpose — the promo is silent and is deleted wholesale when it ends.

### DIFF
--- a/docs/telemetry-events.md
+++ b/docs/telemetry-events.md
@@ -6846,6 +6846,17 @@ or
 
 ```typescript
 {
+  // One-time out-of-window trial reset, attempted from Graph state builds; `failed` may repeat within a session (retries), paid accounts emit no event
+  'action': 'auto-reset-trial',
+  // `refused` = the reset 409'd an account the eligibility check approved (e.g. paid-org members); `failed-shape` = the eligibility payload no longer matches what the client reads
+  'outcome': 'reset' | 'not-eligible' | 'refused' | 'failed' | 'failed-shape'
+}
+```
+
+or
+
+```typescript
+{
   'action': 'start-preview-trial:graph',
   'day': number,
   [`day.${number}.startedOn`]: string,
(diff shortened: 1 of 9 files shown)


sergeibbb:
augment review

PR #5703, opened 2026-08-11
Restores the interrupted comparison itself after signing back in

### LINKED ISSUE #5671: Signing back in after an interrupted comparison reopens different references
### Summary

Signing out while a comparison is open and then signing back in returns you to a comparison — but not the one you were looking at. The references are not carried across, so what reopens is the default shape (current branch against the working tree) rather than the two references being compared.

Observed: with `main` compared against `feature` open, signing out showed the comparison-specific sign-in message as intended; signing back in reopened the panel with `main` against `main`.

### Impact

The sign-in screen promises that signing in completes the task you came for, and for the other three tasks it does exactly that. For a comparison it returns something that looks the same — the comparison panel, in the same place — while quietly comparing different references. Nothing indicates the switch, so the likely reading is that the comparison is still the old one until the numbers look wrong; the references then have to be picked again from memory.

Low severity: nothing is lost or changed in the repository, and the panel is one re-selection away from correct.

### Validation

1. Signed in, open the _Commit Graph_ and start a comparison between two named references (not the default current-branch-versus-working-tree shape).
2. Sign out while that comparison is open.
3. Confirm the sign-in screen shows the comparison-specific headline and message — that part works and should keep working.
4. Sign back in.
5. Expected: the comparison reopens on the same two references. Actual: it reopens on the default shape, with the original references gone.
6. Confirm the other three tasks (composing, review, resolving conflicts) still return to what was interrupted, including a comparison-free case — those are unaffected.

### Risk

Carrying the two references needs somewhere to put them, which the current single-target restore shape has no room for — so this is a deliberate limitation rather than an oversight, and widening it touches whatever else that shape feeds. The alternative is to say so on the screen instead, which keeps the promise honest without new plumbing. Either way worth re-checking: the three non-comparison tasks still restore, a comparison opened in its default shape still restores cleanly, and a comparison open at the same time as an active task still yields to that task.

### Environment

- GitLens: dev build of `main` @ `acf25a3a9` (would ship in the 19.0 line)
- Found while verifying #5534

Related: #5534

sergeibbb:
The sign-out capture now carries the open comparison's refs — they already live in the durable branchCompare* signals, so no transport is involved — and the post-sign-in restore feeds them back into openCompareMode instead of rebuilding the default current-branch-vs- working-tree shape. Ref-less compares (no right ref yet, or no repo path) keep the default-shape restore.

(#5671)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,6 +61,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 
 ### Fixed
 
+- Fixes signing back in after a sign-out interrupted a _Commit Graph_ comparison reopening the comparison on different references &mdash; the interrupted comparison's own references are now carried across the sign-in instead of the default current-branch-versus-working-tree shape ([#5671](https://github.com/gitkraken/vscode-gitlens/issues/5671))
 - Fixes rows in the _Commit Graph_ side bar panels nudging their trailing decorations to the left on hover
 - Fixes AI conflict resolution failing outright when the AI backend won't accept tool calls &mdash; it now falls back to resolving without consulting the repository, as it already did for AI providers that can't carry tool calls at all, instead of failing the whole operation. Errors forwarded from an AI provider also now carry that provider's own message, rather than only the backend's generic "upstream AI provider error"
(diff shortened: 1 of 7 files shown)


sergeibbb:
augment review

PR #5704, opened 2026-08-11
Answer MCP callers of Start Work/Review instead of hanging on an interactive prompt (#5679)

### LINKED ISSUE #5679: Start Work and Start Review never answer an MCP caller when no integration is connected
### Summary

`gitlens_start_work` and `gitlens_start_review` never answer an MCP caller when no hosting integration is connected. Instead of reporting that, the wizard opens its interactive _Connect an Integration_ quick pick and waits for a person — but the only callers of these tools are agents, which have no way to answer it. The call stays pending until the client gives up, and the picker is left open in the editor with focus-out disabled, so nothing dismisses it.

This is the default state of a fresh profile and of every CI run.

### Impact

An agent asked to start work on an issue gets no result and no error — indistinguishable from a dead extension — after waiting out its own timeout; the bundled CLI would wait ten minutes before giving up. Meanwhile the user's editor is left showing a modal quick pick they never asked for, on top of whatever they were doing.

The same call from a person is fine: the prompt is exactly right for them, which is why this only shows up on the programmatic path. Both tools are published with a destructive hint, so a cautious client asks the user before calling — and the reward for approving is a hang.

### Validation

1. Start with no hosting integration connected (a fresh profile).
2. From an MCP client, call `gitlens_start_work` with `directory` set to the open workspace and any `issue_url`.
3. Expected: a structured error saying an integration is required. Actual: no response; after a few seconds the _Start Work • Connect an Integration_ quick pick is on screen ("Connect an integration to accelerate your work" / Cancel), and the call is still pending. The client eventually times out.
4. The picker stays open — moving focus away does not dismiss it.
5. Repeat with `gitlens_start_review` and any `pr_url` — same outcome.

### Risk

Low, and the fix is local: on the programmatic path, report that an integration is required instead of entering the connect step, and make sure the pending result is settled on every exit from the wizard. Worth checking the same pattern in the other wizards a programmatic caller can reach.

Investigated on both sides before filing: the bundled CLI cannot prevent this. Its capability gate is derived from the GitLens version rather than from integration state, so at call time it cannot know the wizard will prompt; once the request is sent it can only wait.

### Environment

GitLens `main` @ `eb6b11404`, VS Code 1.132.0, Windows 11. Observed through the bundled gk MCP server (v3.1.71) on a signed-out profile with no integrations connected.

Two E2E cases in #5678 are parked against this and flip on when it is fixed.

Related: #5106

nikolay-1986:
Fixes #5679

### Summary

`gitlens_start_work` and `gitlens_start_review` never answered a programmatic (MCP/agent) caller when the wizard needed to prompt. With no hosting integration connected — the default for a fresh profile and every CI run — the wizard fell back to the interactive **Connect an Integration** quick pick. No agent can answer it, so the awaited deferred (`state.result`) was only ever settled in the wizard's `finally`, which a suspended `yield` never reaches: the CLI waited out its ~10-minute HTTP budget, the client timed out, and a modal picker was left open in the editor.

### Fix

On the programmatic path (`source === 'mcp'`), settle the deferred with a structured error and bail **before every interactive step a caller can't answer** _within the Start Work / Start Review wizards themselves_, instead of yielding a quick pick no one will pick:

- **Connect-an-Integration step** — report that an integration is required.
- **Start Work's issue-pick step** — when an integration *is* connected but the supplied issue URL doesn't resolve, settle and return instead of falling through to the manual picker (mirroring how Start Review already handles an unresolved PR URL).
- **Pro/sign-in access gate** — `ensureAccessStep` gains an opt-out `interactive` flag (defaults to `true`); the two wizards run it non-interactively on the MCP path and settle with a *"GitLens Pro is required"* error when access is gated. All 7 other callers omit the flag, so their interactive gate is byte-for-byte unchanged.

Net effect: no interactive `yield` reachable by a real MCP caller **within these two wizards** can be reached without `state.result` being settled first.

### Known follow-up (out of scope) — #5706

This PR covers the interactive steps inside the Start Work / Start Review wizards. It does **not** cover the *happy path* (integration connected + valid URL) once it reaches the nested `WorktreeCreateGitCommand`, which forces `state.confirm = true` (`worktree/create.ts:147`) and yields its own interactive confirmation + `worktrees` access gate — a **pre-existing** hang on a different trigger than #5679. Suppressing that non-interactively means threading intent through a shared command (~7 callers), so it's tracked separately in #5706. Thanks to the automated review for flagging it.

### Tests

Un-parks the two E2E regression cases that were `fixme`'d against this defect in `mcpGitlensWritePath.test.ts`. Both now pass against a live VS Code instance:

```
ok gitlens_start_work answers a malformed issue URL instead of hanging
ok gitlens_start_review answers a malformed PR URL instead of hanging
```

### Validation

- `pnpm run check` (oxlint type-aware + type-check) — clean
- `pnpm exec playwright test tests/e2e/specs/mcpGitlensWritePath.test.ts -g "instead of hanging"` — 2 passed
- Independent code review across 3 rounds confirmed correctness, that non-MCP callers are unaffected, and that no interactive-hang path remains for an MCP caller *within these two wizards* (the downstream worktree-create path is tracked in #5706).


nikolay-1986:
Confirmed — this is a valid, verified catch. Traced it: `worktree/create.ts:147` forces `state.confirm = true` unconditionally and `canSkipConfirm` is `false`, so `confirm()` (`quickCommand.ts:82`) always enters the interactive Confirm step (`create.ts:277`), plus an interactive `ensureAccessStep('worktrees')` (`create.ts:179`). The nested `worktreeResult` deferred is awaited in branch-create, so a suspended `yield` hangs the top-level `state.result` — same downstream mechanism you described.

It's a pre-existing hang on a different trigger than #5679 (this is the credentialed happy path reaching worktree creation, not the no-integration prompt), and closing it cleanly means threading a non-interactive intent through the shared `WorktreeCreateGitCommand` (~7 callers). To keep this PR focused on #5679, I've scoped the claim in the description and filed it as a follow-up: #5706. Thanks for the flag.

nikolay-1986:
Good catch — fixed in 8426eb026. The two guards no longer hard-code "GitLens Pro is required". They now call a new `getAccessGateErrorMessage(container, feature, repoPath, action)` helper in `steps/access.ts` that mirrors `ensureAccessStep`'s own gate logic and returns the message matching the actual gate:

- **email not verified** → "Verify your email address to {action}."
- **paid subscription gate** (account present, paid plan required) → "GitLens Pro is required to {action}."
- **otherwise** (no account / trial) → "Sign in to GitLens or start a Pro trial to {action}."

`ensureAccessStep`'s contract is unchanged, so the 7 interactive callers are unaffected. `pnpm run check` clean; the two E2E regression cases still pass.

PR #5724, opened 2026-08-12
Fixes dismissed onboarding re-appearing in new windows

sergeibbb:
# Description

The Commit Graph's "Welcome to the Commit Graph" gate (`graph:intro`) kept re-appearing in every new window and after window reloads, even though the flag is already stored globally.

**Root cause:** all onboarding state persists as one aggregated `gitlens:onboarding:state` key in `globalState`, and `OnboardingService` served every read and write from a snapshot cached at first read (which happens during activation in every window, via `migrateLegacyState`). VS Code broadcasts `globalState` writes into the other running windows' extension hosts, but no event reaches the extension for those external changes — so the snapshot in every other open window went stale the moment one window dismissed something. Any later onboarding write from such a window (a coach mark's seen-state, a banner/callout dismissal, the layout prompt) then saved the entire stale aggregate back, silently resurrecting items dismissed elsewhere. This affects every dismissible surface, not just the Graph intro — the intro is just the most visible victim now that the graph writes coach-mark state routinely.

**Fix:**

- `OnboardingService.getOnboarding` now always reads through to the memento (which VS Code keeps fresh across windows), keeping the old snapshot only as the "previous" side of the external-change diff in `onStorageChanged` — the same read-through pattern `UsageTracker` already uses, which is why usage/walkthrough state never had this problem
- `migrateLegacyState` no longer rewrites the whole aggregate on every activation when there is nothing left to migrate
- Also fixes a literal `&mdash;` rendering as text in the welcome body copy (the entity sat in a plain string interpolated into the Lit template, so Lit escaped it)

Regression tests in `src/onboarding/__tests__/onboardingService.test.ts` model the multi-window scenario over a shared backing store with clone-on-read semantics; the "stale window resurrects a dismissal" and "activation rewrites migrated state" tests fail on the previous code and pass with the fix.

# Checklist

- [x] My changes have a descriptive commit message with a short title

https://claude.ai/code/session_01NhWrffdMSoQ4L1v2LtgdH6

---
_Generated by [Claude Code](https://claude.ai/code/session_01NhWrffdMSoQ4L1v2LtgdH6)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,6 +76,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 
 ### Fixed
 
+- Fixes dismissed onboarding &mdash; the _Commit Graph_'s welcome and coach marks, banners, callouts, and other one-time surfaces &mdash; re-appearing in new windows and after window reloads: all onboarding state is saved together, and a save from a window still holding state from before a dismissal (dismissals never reach other open windows) wrote that older state back, silently un-dismissing what was dismissed elsewhere; saves now always apply to the freshest stored state, and activation no longer rewrites the state when there's nothing to migrate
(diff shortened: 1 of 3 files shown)


sergeibbb:
Closing as stopped being able to reproduce.

PR #5740, opened 2026-08-17
perf(integrations): let scoped PR reads request the lightweight row shape

julianmesa-gitkraken:
> Targets `core`, not `main`.
>
> The provider-apis half landed in gitkraken/provider-apis-package-js#300. This PR
> consumes `@gitkraken/provider-apis` 0.57.0, so the repo-scoped `fields` map is honored at runtime; the `searchPullRequestsPage` half is self-contained.

## Why

The **account-wide** branch of `listPullRequestsPage` has always asked its provider
for the lightweight row shape (`summary: true`). The **repo-scoped** branch had no way
to, so every scoped page paid for a per-row build-status rollup regardless of what the
consumer read. On GitHub that rollup is:

```
commits(last: 1) { ... statusCheckRollup { contexts(first: 100) { ... } } }
```

per row — up to 100 check contexts per PR for a page of up to 100 — and consumers
collapse the whole thing into one tri-state (`statusCheckRollupState`).

Separately, `searchPullRequestsPage` always selects the full PR fragment, which
resolves `assignees(25)` + `latestReviews(25)` + `reviewRequests(25)` per node. That
one is *server* time rather than payload, and it is already known to be expensive
here: it is why `defaultPullRequestSearchPageSize` is 30 rather than 100 (GitHub
rejects the wide search with `Resource limits for this query exceeded`).

## Measured

Interleaved to cancel network drift:

**provider-apis path** — `gh api graphql`, 100-PR repo, 4 pairs:

| | with rollup | without |
|---|---|---|
| Payload | 55620 B | 14110 B |
| Wall clock | median ~2000 ms | median ~1070 ms |

**Own-GraphQL path** — 992-PR repo, one 30-node page, 3 rounds:

| Query | Wall clock | Bytes | Per node |
|---|---|---|---|
| full fragment, 30 | ~3090 ms | 155 KB | 103 ms |
| lite fragment, 30 | ~1150 ms | 140 KB | 38 ms |
| lite fragment, 100 | ~1800 ms | 455 KB | 18 ms |

Near-identical bytes at equal node count, so the win is server-side field resolution.
It compounds with the page size, since the default follows the projection: 100 PRs
cost four sequential full pages (~12.4 s) or one lite page (~1.8 s).

**End to end** in a real consumer (Kepler's worktree correlation, which reads neither
reviews nor checks): the scoped PR search went from **~7474 ms to ~3714 ms median** per
request (range 2557-5901 across runs) — its slowest cold-start request.

Re-measured after the `commitCount` fix. The earlier ~5040 ms figure was taken while
the map still kept the subtree, so it reflected only the fragment change on the
own-GraphQL route; the remaining gap is what the provider-apis gate contributes.

## What this does

`summary` is accepted on `listPullRequestsPage` / `searchPullRequestsPage` and threaded
down. Two mechanisms, because the two reads reach GitHub by different routes:

- **Repo-scoped list read** → provider-apis, so it translates to that package's
  `GitHubPullRequestFieldMap` (`summaryPullRequestFields`). The map explicitly drops
  both gated fields because provider-apis treats an absent map as "select everything"
  and a present one as the complete request. The required local map type makes it a
  compile error to omit either gate.

  **`headCommit` AND `commitCount` are both `false`, and both are required.** On GitHub
  the two come from one selection — `commits(last: 1) { totalCount nodes { commit {
  statusCheckRollup { contexts(first: 100) } } } }`, where `commitCount` is the
  `totalCount` — so provider-apis gates the subtree on the pair. An earlier revision of
  this PR left `commitCount: true`, which kept the whole subtree and made the map save
  nothing on this route while still passing every test in both packages. Dropping
  `commitCount` matches what `summary` already means here: `gqlPullRequestLiteFragment`
  selects no `commits` either, and both fields are optional on `PullRequestShape`.
- **`searchPullRequestsPage`** builds its own GraphQL, so it selects
  `gqlPullRequestLiteFragment` — the same projection `searchMyPullRequests` already
  uses — while retaining stack info and omitting review, check, and diff enrichment.
  Its mapper follows the projection too
  (`fromGitHubPullRequestLite`): mapping a lite node with the full mapper would report
  "no reviewers"/"no checks" as facts rather than as unselected.

For `searchPullRequestsPage` only the **default** page size follows the projection,
matching what `searchMyPullRequests` documents. An explicit `pageSize` is still
honoured up to the maximum either way — that contract has a test, which is how the
first version of this change was caught capping it.

## Compatibility

Opt-in throughout: every existing caller keeps today's shape and payload byte for
byte, and a provider with no lightweight projection ignores the flag, so it is a hint
rather than a guarantee about which fields are present. Callers that DO read the
rollup — the Graph sidebar and PR sheet — are unaffected, which is exactly why this is
per-call rather than a removal.

## Testing

- New `summaryPullRequestFields.test.ts`, which is the test whose absence allowed the
  `commitCount` bug: the map lives in this package, the gate that interprets it lives
  in another, and nothing connected them. It asserts the map that actually **reaches**
  provider-apis (captured from the real read) rather than the map's declaration —
  restating the declaration would pass regardless of what the gate does with it — and
  pins both halves: exactly the two gated keys are false, and a non-summary read sends
  no map at all. Verified it fails when `commitCount: true` is restored.
- 3 new tests in `pullRequestSearch.test.ts`: the summary read selects the lite
  fragment and the maximum default page size and drops
  `latestReviews`/`reviewRequests`/`statusCheckRollup` while keeping
  `baseRefName`/`headRefName`/`body`; the non-summary read keeps the full fragment and
  the cost-safe default; an explicit page size is honoured on the summary path too.
- `pnpm test:packages` green across every package (git-github 153, integrations 775,
  git 562, utils 451, git-cli 451, agents 235, ai 113, commit-graph 296, ipc 18).
- `tsc --noEmit` clean for both changed packages.

## Dependency update

- Updated the workspace catalog and lockfile to `@gitkraken/provider-apis` 0.57.0, which includes the runtime support for the scoped `fields` projection.
- Targeted validation after the update: GitHub 153 tests, integrations 775 tests, type-check, build, and consumer fixture all pass.

### DIFF
--- a/packages/plus/git-github/src/api/__tests__/pullRequestSearch.test.ts
+++ b/packages/plus/git-github/src/api/__tests__/pullRequestSearch.test.ts
@@ -219,6 +219,51 @@ suite('GitHubApi.searchPullRequestsPage', () => {
 		assert.match(getCalls()[1].query, /scopeOpen: search\(first: 100,/);
 	});
 
+	test('selects the lite fragment and the maximum default page size for a summary read', async () => {
+		// The full fragment resolves assignees(25) + latestReviews(25) + reviewRequests(25) per node,
+		// which is SERVER time rather than payload: measured against a 992-PR repo, one 30-node page
+		// took ~3090ms full vs ~1150ms lite for near-identical bytes (155KB vs 140KB). It compounds
+		// with the page size, since the default follows the projection: 100 PRs cost four sequential
+		// full pages (~12.4s) or one lite page (~1.8s).
+		const { config, getCalls } = capture();
+		await new GitHubApi(config).searchPullRequestsPage(provider, token, {
+			repos: ['o/a'],
+			summary: true,
+		});
+
+		const query = getCalls()[0].query;
+		assert.match(query, /scopeOpen: search\(first: 100,/);
+		// Review/check/diff selections are what the lite shape drops.
+		assert.doesNotMatch(query, /latestReviews/);
+		assert.doesNotMatch(query, /reviewRequests/);
+		assert.doesNotMatch(query, /statusCheckRollup/);
+		// What its consumers DO read must survive.
+		assert.match(query, /baseRefName/);
+		assert.match(query, /headRefName/);
+		assert.match(query, /body/);
+	});
+
+	test('keeps the full fragment and the cost-safe default when summary is not requested', async () => {
+		const { config, getCalls } = capture();
+		await new GitHubApi(config).searchPullRequestsPage(provider, token, { repos: ['o/a'] });
+
+		const query = getCalls()[0].query;
+		assert.match(query, /scopeOpen: search\(first: 30,/);
+		assert.match(query, /latestReviews/);
+	});
+
+	test('honours an explicit page size on the summary path too', async () => {
+		// Only the DEFAULT follows the projection; an explicit preference is still the caller's.
+		const { config, getCalls } = capture();
+		await new GitHubApi(config).searchPullRequestsPage(provider, token, {
+			repos: ['o/a'],
+			summary: true,
+			pageSize: 10,
+		});
+
+		assert.match(getCalls()[0].query, /scopeOpen: search\(first: 10,/);
+	});
+
 	test('translates the declared mention relationship without widening to commenter', async () => {
 		const { config, getCalls } = capture();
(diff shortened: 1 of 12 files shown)


julianmesa-gitkraken:
augment review

PR #5746, opened 2026-08-18
Tracks each Commit Composer session with an AI conversation ID

ramin-t:
Threads a session-scoped conversation ID through every AI request the _Commit Composer_ makes, so one compose reads as one session instead of one per round trip — and fixes three paths that were mis-resolving a plan's refine continuation. No linked issue.

## Why

Compose never sent a conversation ID. One compose is routinely several AI calls: `compose-tools` retries internally whenever the model returns output its validators reject, and each refine of the returned plan is another call. Nothing tied those requests back to the work that produced them, so `ai/generate` events for compose landed unattributable, and multi-call sessions were indistinguishable from unrelated one-shot runs.

The plumbing already existed — `AIProviderService.sendRequest` takes a `conversationId`, forwards it to the provider, and only the GitKraken provider emits it (`openAICompatibleProviderBase` deliberately drops it so it can never reach a third-party API). Compose simply never passed the option. AI conflict resolution had already solved the same problem with `_resolveConversationIds`; this follows that shape.

## How it's resolved

A session mints one ID lazily and every request in it carries that ID.

| Event | Behavior |
| --- | --- |
| First generate | Mint |
| `compose-tools`' internal validation retries | Same ID — the AI port is built once per call with the ID closed over, so retries inherit it with no retry-aware plumbing |
| Refine | Continues the session that produced the plan |
| Leave the panel / switch rows, return, refine | Continues the same session |
| Start over on top of a live plan | That plan is abandoned; its session ends and a new ID is minted |
| Generate fails or is cancelled, then retried | Continues the same session — a failed attempt is not a new one |
| Regenerating one commit's message | Same ID — a distinct AI action, but made from inside the session |
| Apply, or panel teardown | Session ends; its accumulated usage is reported once |
| Discard, Restart-then-close, repository switch | Session ends, when the webview can name the plan it dropped |

Passing the option also switches usage reporting from per-request to per-session aggregate, which obliges the caller to end the session exactly once — hence the explicit end on apply, on abandon, and on `dispose`.

## Beyond the minimal wiring

**Host state is keyed by compose session, not repository path.** The two are the same string today: compose is only enterable on the working-changes row, so a repository has one compose session. But that invariant is enforced by a single line in the webview (`enterModeForWip` nulling `shas`, which is what keeps `anchorKey` off its `multicommit` branch), and the host now partitions sessions without depending on it. Were compose ever reachable from a second anchor on one repository, two live sessions would stay separate rather than sharing a conversation and evicting each other's plan. Since the key travels beside a repository path and a cache key on these RPCs, it is branded — mirroring the webview's own `AnchorKey` — so a transposed argument is a compile error. That caught 8 real sites when introduced.

Two RPCs (`regenerateProposedCommitMessage`, `reorderProposedCommits`) used `repoPath` for nothing but the old map lookup, so they lost the parameter; both are pure in-memory plan operations now identified by session and cache key.

**Per-commit message regeneration joins the session.** It is a separate AI action (`generate-commitMessage`), but the user invokes it from inside a compose they are already in, so it belongs to that session rather than reading as a task of its own. `generateCommitMessage` now accepts a `conversationId` and the compose panel passes its session's; standalone message generation elsewhere passes nothing and is unchanged. **Consequence for anyone reading the telemetry:** a compose session's IDs now span two `type`s (`commits` and `generate-commitMessage`), so counting sessions means counting distinct IDs across both rather than per `type`. The field's docs say so.

**Sessions the webview abandons are ended eagerly.** Three paths dropped the run's registry entry — the webview's only handle on the plan — without telling the host, leaving the plan and its conversation until the next compose on that session or panel teardown: Discard, the Restart-then-close destroy, and a repository switch, which clears the whole registry and orphans every session it held. A new `discardCompose` releases the plan and closes the conversation. It names the plan the webview believed the session held and the host acts only while that still matches, so a call landing after the user has started over is a no-op instead of tearing down the newer plan and a conversation still in use — the call is fire-and-forget, so that guard is what makes it safe. All three destroy paths funnel through `destroyEngagedOperation`, so one hook covers Discard and Restart-then-close.

The call does nothing when the webview names no plan, and that is deliberate rather than a gap: a session still waiting on its first generate also holds no plan, so there is nothing in "no plan" to tell it apart from an earlier destroyed run. Acting on it would let a late call end the newer session's conversation, after which the in-flight request completes under an already-reported ID — splitting the session and leaving everything after it with nothing to flush it. A cancelled or failed generate instead keeps its conversation for the retry to continue, which is what should happen to it, with dispose as the backstop.

**One AI-layer change came with this.** Usage buckets by provider and model, the aggregate reports whichever bucket holds the most tokens, and each bucket keeps the action it was created with. That was harmless while every request in a compose conversation was the same action — but once message regeneration joined the session, a model switch part-way through could hand the token majority to the regeneration bucket and report the whole session as message generation. A conversation now reports the action that opened it.

## Behavior changes

Three pre-existing defects that only became session-tracking failures once tracking existed. All change what the user sees:

**Leaving compose and returning.** `hideMode` is the preserve-leave path — it keeps the run's registry entry so the plan is projected back on return, and deliberately captures the refine draft so the user can resume — yet it also dropped the cache key that makes refining possible. Toggling out and back, or switching rows and returning, preserved the intent to refine while destroying the ability: the plan on screen was silently discarded and regenerated from scratch. It no longer clears the continuation. `exitMode` and a switch to a different selection are the genuine session ends and still do.

**Retrying a failed refine.** A failed refine leaves an error on the entry, which the refine-vs-cold-start decision read as "no plan to continue", so Retry discarded a plan that was never invalid. That decision no longer requires a result to be on screen — a live cache key alone means a plan exists.

The key now lives on the anchor's registry entry, which makes it per-anchor (a run settling on another worktree row can no longer aim a refine at the wrong plan) and durable across an error. The global `composeCurrentCacheKey` signal it replaces had no readers left and is removed rather than left as state that looks authoritative but is not.

Carrying a possibly-stale key is safe because the host validates it: a refine requires the key to match the plan the host still holds for that session, so a handle that outlived its plan falls through to a fresh generate instead of failing the library's cache lookup — the behavior that code path already documented but did not implement.

That validation compares the webview's key against the host's map, which catches disagreement but not the case where both agree on a key the *library* has dropped. A refine that reaches the library swaps in its replacement and deletes the prior plan, so a throw in the steps after that left the map naming something gone. The catch now releases the registered key when the throw came after such a swap, so the retry cold-starts — while deliberately keeping the conversation, since the user is retrying the session rather than abandoning it.

**Walking Back to the scope picker.** Back parks the entry in `backed` and resets the panel to the idle scope picker, retaining the plan only so Forward can restore it without re-running the AI. Reading the refine handle straight off the entry did not account for that, so a generate submitted from the picker was dispatched as a refine — continuing the plan the user had just walked away from, ignoring the scope they had since chosen, and keeping the abandoned plan's conversation. The handle is now withheld while the entry is `backed`, which is exactly the window where the plan exists but is not on screen; Forward returns it to `complete`, so resuming a plan still resumes its session.

## Tests

Added, all passing:

- **Host session lifecycle** (11, `graphInspectServices.test.ts`) — one per rule: refine continues the generate's session; starting over on a live plan ends the old one and mints a new one; a retry after failure continues; apply is terminal; a repository and its worktree stay independent; two sessions on one repository path stay separate; teardown ends every open session; a refine that fails after swapping the plan lets go of the dead key; a regenerated message continues the session without ending it; a discard ends it immediately; a late discard cannot end a newer session.
- **Webview determination** (8, `detailsWorkflowController.test.ts`) — the anchor round trip dispatches as a refine with the prior key; the session key is the engaged anchor's; no plan cold-starts; a retry after a failed refine dispatches as a refine; a settled error keeps the key on the entry; Back cold-starts and Forward restores the continuation; and a discard actually issues the host call with the right session and cache key. That last one exists because the host tests pass happily with the webview wiring missing entirely.
- **Port forwarding** (2, `compose/integration.test.ts`) — the ID reaches `sendRequest`, and repeated calls through one port all share it, which is what makes the library's internal retries part of the run.
- **Usage accounting** (1, `aiProviderService.telemetry.test.ts`) — a later action with more tokens does not relabel the session.

Every fix above is mutation-verified: reverting it fails its test and nothing else.

Not covered: `compose-tools`' internal retry loop is structural (one port instance per call, ID captured in its closure) and has no seam at this layer without stubbing the library; the port test covers the hop that carries it.

`pnpm run build` is clean — no errors, no warnings. Full unit suite: 2339 passing, 1 failing, which is a pre-existing `discard.test.js` `EPERM` on a temp directory that reproduces in isolation with none of this loaded.

Worth flagging for anyone running the suite: `.vscode-test.mjs` pre-expands its `files` glob with backslash paths that `@vscode/test-cli` then re-globs, so the default `pnpm run test` matches nothing and reports `0 passing` with exit code 0 — a silent green. Runs above used `--run` with forward-slash paths.

## Validation

1. Open the _Commit Graph_ on a repository with uncommitted changes, select the working-changes row, enter _Compose_, and generate a plan.
2. Refine it twice. The initial generate and both refines should report the same `conversationId` on their `ai/generate` events.
3. Toggle out of _Compose_ and back in, then refine. **Previously the plan was discarded and regenerated from scratch; it should now refine the plan on screen**, under the same ID.
4. Force a refine to fail, then click Retry. **Previously this cold-started; it should now retry as a refine**, under the same ID.
5. Walk back to the start of the compose UX and generate again — a different ID.
6. Generate in a worktree of the same repository while the main plan is live — a different ID; returning to either and refining resumes that run's own ID.
7. Regenerate one proposed commit's message — same ID as the session, and the session stays open.
8. Discard a plan, then compose again — the discarded session ends at the Discard, and the new compose gets a new ID.
9. Cancel a generate before it produces a plan, then compose again — the same ID continues, since a cancelled attempt is not a new session.
10. Apply a plan, then compose again — a new ID.

## Deliberately not addressed

- Sessions lost to a crash or a killed window cannot be ended — catching those would mean persisting session state to disk, which is not worth it. One conservative case also remains by choice: if a run completes host-side but the webview has not settled when the user discards, the two keys disagree and the host skips rather than risk tearing down a live plan, leaving that one to the existing backstops.
- A refine ignores a changed scope picker by contract (no re-collection), so changing scope and hitting generate while a plan is displayed silently drops the new scope. Pre-existing and orthogonal.


ramin-t:
Confirmed and fixed in d535935e4 — thanks, this was a real regression.

`compose.back()` calls `enterComposeBacked()`, which resets the resource to the idle scope picker, then `transitionEngagedEntryExecState('compose', 'backed')`, which re-registers via `{ ...entry, execState }` — so the `cacheKey` survived. `composeCacheKeyForAnchor` read it with no state check, so `isRefine` was true and the next generate went out with `mode: 'refine'` and the prior key. The host then took the refine path (the key still matched its tracked plan), skipping both the re-collection of the newly chosen scope and the abandon branch that ends the old conversation. Same for `backFromError()`, which also lands in `backed` via `enterComposeBacked(prev)`.

The handle is now withheld while the entry is `backed` — the window where the plan exists but is not on screen. `forward()` transitions back to `complete`, so resuming a plan still resumes its session.

Two tests added, and I mutation-checked them: dropping the `backed` guard fails `walking Back to the scope picker cold-starts the next generate` and nothing else.

ramin-t:
augment review

ramin-t:
Confirmed and fixed in 3d172b806. This one landed squarely on the gate I added, so thanks for pushing on it.

Traced it: a successful `refinePlanForGraphDetails` sets the new key and does `this._cache.delete(input.priorCacheKey)`, while `_activeComposeCacheKeys.set(sessionKey, cacheKeyToRegister)` happens only after the downstream steps — `signal?.throwIfAborted()` and `svc.commits.getCommit(baseAnchorSha)` (the `getCommit('HEAD')` one is `.catch()`-guarded, that one is not). So a throw in that window leaves the library holding neither key while the map still names the deleted one.

The part that made it bite: my gate compares the webview's key against the map, so it catches disagreement — but here both sides agreed on a key nothing held, the gate passed, and the lookup failed instead of taking the cold-start fallback the gate exists to provide.

Fixed at the root rather than by probing the cache: the catch now drops the registered key when the throw came after a refine had already swapped the plan, tracked by a flag set at the moment `cacheKeyToRegister` is captured. The retry then cold-starts, since the plan really is gone. The conversation is deliberately left alone — the user is retrying this session, not abandoning it, so it continues instead of being ended and re-minted, which also keeps the "retry after a failure continues the same conversation" rule intact.

Added a test that drives the window (refine resolves, a downstream step throws) and asserts the key is released, the session is not ended, and the retry reuses the conversation. Mutation-checked: removing the fix fails that test alone.

ianhattendorf (approved):
Looks pretty solid to me. A couple thoughts:
1. It looks like there are some cases where the conversation isn't flushed (like an explicit discard: `compose.discard` → `destroyEngagedOperation`). I would take a look and see if this could be tightened up a little. Without persisting to disk (probably not worth it) we won't be able to catch them all, but there should be some room for improvement still.
2. Would it maybe make sense to also treat `regenerateProposedCommitMessage` during the compose as part of the same conversation? Right now, from what I can tell regenerating the commit message would trigger a new flat fee.

ramin-t:
Both addressed in 647c8a2d1 — thanks Ian.

**2. Message regeneration** — you were right that it was minting its own session. `generateCommitMessage` spreads its options straight into `sendRequest`, which already accepts `conversationId`, so it was a two-line change: the action takes the option, and the compose panel passes its session's. Standalone message generation elsewhere passes nothing and is unchanged.

One consequence worth flagging for whoever reads the telemetry: a compose session's IDs now span two `type`s (`commits` and `generate-commitMessage`), so counting sessions means counting distinct IDs across both rather than per `type`. I updated the field's docs to say so.

**1. Unflushed conversations** — three paths, all of them the same shape: the webview drops the run's registry entry (its only handle on the plan) without telling the host, so the plan and its conversation sat there until the next compose on that session or panel teardown.

- Discard
- The Restart-then-close destroy
- A repository switch, which clears the whole registry and orphans every session it held

Added a `discardCompose` RPC that releases the plan and closes the conversation. It names the plan the webview believed the session held, and the host acts only while that still matches what it holds — so a call landing after the user has started over is a no-op rather than tearing down the newer plan and a conversation still in use. That guard matters because the call is fire-and-forget. All three destroy paths funnel through `destroyEngagedOperation`, so hooking it there covers Discard and Restart-then-close together; the repository switch closes each session it drops.

Sessions that never produced a plan are ended too, which is how a cancelled or failed generate's conversation gets closed — previously those also waited for teardown.

You're right that some are unreachable: a crash or a killed window takes any open session with it, and I agree persisting to disk isn't worth it. What's left uncaught is now just that class, plus one deliberate conservative case — if a run completes host-side but the webview hasn't settled when the user discards, the keys disagree and the host skips rather than risk a wrongful teardown, leaving that one to the existing backstops.

Four tests added, each mutation-checked (removing the fix fails that test and nothing else): the regen carries the session's conversation and doesn't end it; discard ends it immediately; a late discard cannot end a newer session; and the webview actually issues the call on discard with the right session key and cache key — that last one because the host-side tests would happily pass with the client wiring missing entirely.

ramin-t:
augment review

ramin-t:
Both valid, both fixed in 21b4f1bc0.

**`discardCompose`'s undefined/undefined match.** You're right, and this one is worse than a missed case — my own comment claimed the guard protected an in-flight session, when strict equality only does that while the webview names a non-null key. A teardown call for an earlier destroyed run names nothing, a session still awaiting its first generate holds nothing, and `undefined !== undefined` is false, so the newer session's conversation was ended. Worse, the newer compose had *reused* that ID (`getOrCreateComposeConversationId` returns the existing entry, since the abandon branch only runs when a plan is tracked), so its in-flight request then reported under an already-flushed ID and its refines minted a third — the session split exactly as you describe.

There is nothing in "no plan" that distinguishes those two states, so the call is now a no-op unless it names a plan we still hold. The cost is that a cancelled or failed generate no longer ends eagerly — which turns out to be the correct behavior anyway, since the settled rule for this feature is that a retry after a failure continues the same session. Dispose remains the backstop.

**The dominant-bucket action.** Also confirmed by reading it: buckets key on `provider/model`, keep the action they were created with, and `aggregateBYOKUsage` reports the action of whichever bucket has the most tokens. That was harmless while every request in a compose conversation was `generate-commits`; it stopped being harmless when message regeneration joined the session, because a model switch part-way through can hand the token majority to the regeneration bucket. A conversation now reports the action that opened it — which is also just true, since a conversation is one piece of work.

Both fixes are mutation-verified: reverting either fails its new test and nothing else.

Worth noting for the record that my own review panel raised the first of these last cycle and I talked myself out of it — I read it as "relax the guard", argued against relaxing, and missed that the undefined branch was already too permissive. The finding was right and my rebuttal was answering a different question.

ramin-t:
augment review

ramin-t:
augment review

PR #5761, opened 2026-08-20
Fixes transient Community state when switching organizations

### LINKED ISSUE #3546: When in multiple orgs with the same tier paid license, all but one assigned a free subscription
Applies when multiple paid orgs of the same subscription tier exist.

 

There is an update to the checkin route in dev to address this, but it is not backwards-compatible because client-side changes are needed (we do not do a check-in when changing active org).

sergeibbb:
# Description

Client-side adaptation for the org-scoped `gitlens/checkin` API — refs #3546 (internal: gitkraken/vscode-gitlens-private#92).

Users with two orgs holding the same license type see one of them as Community, because the current checkin response keys `paidLicenses` by license *type* and silently drops the colliding org. The server fix scopes the response by the `gk-org-id` header (live on dev, not yet published). This PR makes GitLens correct with **both** response shapes; the user-visible issue is resolved only once the scoped API is published.

**Root cause of the flicker (scoped API):** `switchOrganization` passed the target org to the checkin HTTP request, but `validateAndUpdateSubscriptions` re-read the *configured* (still old) org id and emitted an intermediate subscription computed for it. With a scoped payload the old org has no license in the response, so every switch briefly showed `community-with-account` (~90 ms observed on dev) before a second, corrective emission re-read from the check-in cache. With the current global payload the intermediate emission happens to be correct, which is what masks the bug today.

**Changes:**

- Forwards the target organization id into `validateAndUpdateSubscriptions`, so a switch emits a single, correct subscription change; removes the now-redundant second emission from `switchOrganization`
- Removes the check-in response cache (the in-memory getter, the stored `gk:{userId}:checkin` blob, and its `StoredGK*` types) — the second emission was its only reader, and under the scoped API the stored blob would no longer be a global snapshot of the account
- Keeps the chosen organization when the organization list fails to load during validation — a transient failure no longer erases the configured org (which silently moved the plan to a different org)
- Keeps the already-known organization list on transient fetch failures, and hydrates it from the stored (24h) cache on forced refreshes — so a failed fetch in a fresh window doesn't clear `activeOrganization`, which would drop the `gk-org-id` header from subsequent requests and, under the scoped API, show a paid account as Community until a later validation

**Verification:** live against dev (scoped API): switching test ⇄ test2 (two orgs with same-type Pro licenses) emits exactly one subscription change carrying the correct target plan — previously two, with a wrong intermediate one. Trial states verified unchanged (in-trial license without a payment source, trial reactivation via `nextOptInDate`, headerless login with no organizations). Against the current global API the behavior is indistinguishable aside from the removed redundant emission. `pnpm run check` clean, unit tests (2356) pass.

CHANGELOG entry is deliberately deferred until the server API is published.

# Checklist

- [x] My changes have a descriptive commit message with a short title

### DIFF
--- a/src/constants.storage.ts
+++ b/src/constants.storage.ts
@@ -139,7 +139,6 @@ type GlobalStorageDynamic = Record<`plus:preview:${FeaturePreviews}:usages`, Sto
 	Record<`provider:authentication:skip:${string}`, boolean> &
 	Record<`gk:promo:${string}:ai:allAccess:dismissed`, boolean> &
 	Record<`gk:promo:${string}:ai:allAccess:notified`, boolean> &
-	Record<`gk:${string}:checkin`, Stored<StoredGKCheckInResponse>> &
 	Record<`gk:${string}:organizations`, Stored<StoredOrganization[]>> &
 	Record<`jira:${string}:organizations`, Stored<StoredJiraOrganization[] | undefined>> &
 	Record<`jira:${string}:projects`, Stored<StoredJiraProject[] | undefined>> &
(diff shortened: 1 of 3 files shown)


sergeibbb:
augment review

PR #5828, opened 2026-09-08
Fixes the two Windsurf-only E2E failures left on main

nikolay-1986:
The last two Windsurf-only E2E failures on `main`. Both passed on VS Code and Positron with the identical bundle, and neither cause is what its CI log first suggested — the diagnosis came from driving a live Windsurf instance at the width CI runs.

### `graphScoping` — the row-action strip was revealed by hover

The strip is `visibility: hidden; pointer-events: none` until its row is hovered, focused or selected, so while inert it is transparent to hit-testing and a click aimed at it falls through to whatever is beneath. Playwright reported the merge-target ref pill intercepting the Undo Commit click for the full budget, and that reads like a layering bug — it isn't.

Measured live at a 299px side bar, where rows render in the two-line **list** layout: `.gl-graph__list-line2` spans x 81..282 and the strip x 225..285, so line2 and the pill inside it do underlie the strip for 57px. But the strip is last among the row's children and positioned, so it paints above them. With the row **selected**, `document.elementFromPoint` at the button's centre returns that button, the strip reports `pointer-events: auto`, and clicking Undo Commit moved the branch back one commit and left the undone commit's file staged. The pill only ever answers the hit test while the strip is inert.

Hover cannot hold that state open: the graph re-renders rows on its own, and Chromium does not re-evaluate `:hover` for a freshly inserted element under a stationary cursor. That is exactly the interleaving the CI log shows — attempts reporting "element is not visible" mixed with attempts where the element was "visible, enabled and stable" and the pill intercepted anyway. `clickGraphRowAction` now reveals the strip by selecting the row and asserting `aria-selected`, which is pointer-independent and survives a re-render. It clicks the row's message span, never the row box, so the click cannot resolve to a ref pill and route elsewhere.

### `rebase` — a notification toast over the Abort button

The abort spec failed with a bare "Test timeout" that named nothing, because the describe's 30s budget equalled the config's `actionTimeout`: a stuck action could never report its own timeout, and the page was torn down before `error-context.md` could capture a snapshot (it came back 398 bytes). Raising the budget to 60s, matching the `Execute rebase` sibling, made the failure name its locator — an announcement toast (`.announcement-actions` inside `.notifications-toasts bottom-right`) parked over the Abort button, shown on Windsurf and not on VS Code. It is cleared with `notifications.clearAll` before the click, which beats the sleep-then-Escape dance the later specs in this file use.

### Validation

Proven on CI, not locally — these failures do not reproduce on a local stand. Dispatch run 34227047697: the **Windsurf job is green as a whole** (231 passed, 0 failed), with `graphScoping:255` and `rebase:184` passing on the first attempt and the scoping family at 9 green / 0 red. No regression: VS Code is 232 passed with no failures and no flakes, and all 13 `rebase.test.ts` specs pass on both editors, so `notifications.clearAll` disturbed none of them. `Unit Tests` green.

### Scope

Tests only — no product code is touched. Left untracked here: `treeView:235` still flakes on Windsurf (it failed and passed on retry), but its cause is the shared `ensureGraphDetailsPanelOpen` helper rather than the spec, and consolidating the three panel-open mechanisms is its own change; `graphConflictResolution:295` still fails on Positron with a cause that is not established.

### DIFF
--- a/tests/e2e/graphScopeHelpers.ts
+++ b/tests/e2e/graphScopeHelpers.ts
@@ -253,6 +253,38 @@ export async function clickSidebarRowAction(row: Locator, label: string): Promis
 	await action.click();
 }
 
+/**
+ * Click a gated action on a GRAPH row, revealing the action strip through SELECTION.
+ *
+ * The strip is `visibility: hidden; pointer-events: none` until its row is hovered, focused or
+ * selected (`graph.scss`), so while inert it is transparent to hit-testing and a click aimed at it
+ * falls through to whatever is beneath. On a narrow host that is real: measured live on Windsurf at a
+ * 299px side bar, rows render in the two-line LIST layout and `.gl-graph__list-line2` (x 81..282)
+ * underlies the strip (x 225..285) for 57px — which is why Playwright names line2, or the merge-target
+ * ref pill inside it, as the interceptor.
+ *
+ * Hover is the wrong key for it, even applied immediately before the click: the graph re-renders rows
+ * on its own (WIP state, avatars), and Chromium does not re-evaluate `:hover` for a freshly inserted
+ * element under a STATIONARY cursor — so the strip can go inert between Playwright's actionability
+ * check and its hit test. That is exactly the shape of the CI failure: attempts reporting "element is
+ * not visible" interleaved with attempts where the element was "visible, enabled and stable" and the
+ * pill intercepted anyway.
+ *
+ * Selection is pointer-independent and sticky, so it holds across a re-render. Verified live on
+ * Windsurf at that same 299px width: with the row selected the strip reports `pointer-events: auto`
+ * and `document.elementFromPoint` at the button's centre returns the button itself; clicking Undo
+ * Commit moved the branch back one commit and left the undone commit's file staged.
+ *
+ * The row's message span is the click target for selecting, never the row box — the box also carries
+ * the inline ref pills, and a click resolved to a ref routes to the ref action instead.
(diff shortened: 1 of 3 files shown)


nikolay-1986:
Addressed in `4f0f38ccb`. The finding is right, and it lands on an existing convention I had missed: `executeCommandIfAvailable` is already how `secondarySidebar` calls its workbench commands, for exactly this reason. `notifications.clearAll` is built in rather than ours, so a fork that renames or omits it would have failed the spec on a missing command instead of on what it was testing. Where the command is absent the clear is now skipped and the original toast interception returns — a legible failure rather than a confusing one.

Re-running the dispatch so the branch's validation matches its head commit; the previous run (34227047697) proved the fixes at `ab3ab9dee` with the Windsurf job green as a whole.

nikolay-1986:
Re-run done — [34230266909](https://github.com/gitkraken/vscode-gitlens/actions/runs/34230266909), on `4f0f38ccb`, so the validation now matches this branch's head rather than the commit before the review fix.

Same result as before the change: the **Windsurf job is green as a whole** (231 passed, 0 failed), with `graphScoping:255` and `rebase:184` passing on the first attempt, and all 13 `rebase.test.ts` specs green on both Windsurf and VS Code. VS Code 232 passed, `Unit Tests` green. The one flake is `treeView:235`, which is out of scope here as noted in the description. Positron still fails only `graphConflictResolution:295`, unrelated to this change.

Issue #5859, opened 2026-09-17
Widening the Graph's side bar does nothing on Windsurf, so fork-only failures surface as unrelated timeouts


nikolay-1986:
## Summary

`widenSideBarForGraph` — the helper every Graph E2E spec calls to give the Graph a panel-like side bar — does nothing on Windsurf. It drives `workbench.action.decreaseViewWidth`, which returns without error there but never grows the primary side bar: the freed width goes to the secondary side bar instead (measured: side bar 300px unchanged across all twelve calls, secondary side bar 300 → 878). Closing the secondary side bar first does not help — it closes fine, and then the command stops moving anything at all (editor stays at 1098px, side bar at 300px).

So every Graph spec runs against a 300px side bar on Windsurf while the same spec gets 872px on VS Code, and nothing says so.

## Impact

Width-sensitive assertions are effectively unexercised on Windsurf, and when they do fail they fail as something else. Two came out of one CI run:

- `localization.test.ts` "Graph sidebar panel headers render pseudo-translated" — hard red, all three attempts. At 300px the Branches panel is 54.8px wide while its actions toolbar alone needs 104px, so the panel title is squeezed to zero width and reported `hidden` even though it is mounted and `checkVisibility()` returns true. Reads as a localization failure; is not one.
- `treeView.test.ts` "should switch back to a previously viewed commit correctly" — flaky. The details file tree is virtualized, so a cramped panel mounts no rows at all and the assertion reports `element(s) not found`.

Neither message points at the side bar, which is why both survived as "fork flakiness".

## Validation

Dragging the side bar's own sash widens it on both editors (measured 300 → 820 on Windsurf, 300 → 870 on VS Code), and with the side bar actually widened the two failures above go green on Windsurf — `localization` and `treeView` pass in full, Positron is unaffected (27 passed).

Two details make the difference between a drag that works and one that works sometimes: several sashes stack on that edge (one live, two `disabled`, same x), so a drag aimed at the coordinate grabs whichever is topmost, and the drag has to start with a short movement or it is consumed as a click.

## Risk

[...]

Issue #5858, opened 2026-09-17
The E2E suite never loads a real translated catalog, so the shipped locales have no coverage


nikolay-1986:
## Summary

The E2E suite can only run in the `qps-ploc` pseudo-locale, so none of the locales GitLens actually ships — `es`, `zh-cn`, `zh-tw` — is ever exercised end to end. Asking the harness for any other locale is refused outright: `E2E locale "es" is not supported — only "qps-ploc" has a fixture language pack (tests/e2e/fixtures/language-packs/qps-ploc)` (`tests/e2e/baseTest.ts:438`).

## Impact

The pseudo-locale smoke test is a good regression net for one thing: it catches a user-facing string that shipped without being wrapped for translation. It cannot catch anything about a real translation, because the catalog it reads is generated from the English source at test time and is git-ignored — no test ever loads `l10n/bundle.l10n.es.json`, `l10n/bundle.l10n.zh-cn.json`, `l10n/bundle.l10n.zh-tw.json` or their `package.nls.*` counterparts.

So these ship unobserved: a catalog that has gone stale against renamed keys, a mis-keyed or partially imported translation, a placeholder dropped in a translated string, and any layout that only breaks at the string lengths a real translation produces. Spanish and Chinese landed as experimental translations and currently have no automated signal at all beyond the static catalog validation in `check:l10n`.

## Validation

- A spec that launches in a real shipped locale and asserts a known label renders that locale's catalog entry — the same shape as the existing pseudo-locale smoke test, pointed at a real catalog instead of the generated one.
- At minimum one such run per shipped locale, so a stale or partially imported catalog fails a test rather than reaching users.

## Risk

The gap is widest where a header or control has a fixed-width neighbour and a title that is allowed to ellipsize. The Graph's sidebar panel header is exactly that shape: the title shares its row with an actions toolbar that does not shrink, so the room the title gets is whatever is left over. Measured on a Windsurf E2E instance, the title had zero width while the actions kept their full width — the layout survives it, but a real translation that is longer than the English source consumes that headroom first, and nothing currently measures it.

The same shape is where a right-to-left locale would break first. That is not reachable today — no RTL locale ships, and text direction is `ltr` in every instance measured — but it becomes reachable the moment one is added, and there would be no test to notice.

## Related

#577 tracks internationalization support itself; this issue is about the absence of test coverage for the catalogs that have since landed, not about adding more languages.

nikolay-1986:
Verified — **Passed**.

[...]

PR #5860, opened 2026-09-17
Gates the Graph sidebar panel header localization spec on its text, not its box

nikolay-1986:
`localization.test.ts` has been red on Windsurf, and the failure names the wrong thing: the panel header resolves and reports `hidden`, so it reads as the Graph sidebar never painting, or the pseudo-locale never applying. Neither is what happens — the string renders, correctly pseudo-translated, the whole time. It simply has no width to occupy.

The header title is a flex item with `min-width: 0` and an ellipsis, sharing its row with an actions toolbar that does not shrink. On Windsurf the primary side bar stays at its 300px default, which leaves the Branches panel 54.8px wide against a 104px toolbar, so the title is squeezed to zero and Playwright calls an empty box `hidden` — while `display` is `block`, `visibility` is `visible`, and `checkVisibility()` returns true. Measured A/B on the revision CI ran (`d9de42de9`), same instrument on both sides: VS Code gets an 872px webview and a 67.3px title; Windsurf gets 300px and a 0px title whose `scrollWidth` is still 67.

So the assertion now gates on the text rather than on the box. That also waits out the panel switch the visibility check used to stand in for, and it does not depend on how much room the title was given, which is all this test is about.

It is a locator assertion rather than a poll over `textContent()` on purpose. The header is not merely narrow when something does go wrong — it is gone: the panel renders `nothing` while no panel is active, so the node is detached. `textContent()` on a detached locator waits out `actionTimeout` (30s, three times this assertion's own budget) and then throws, and `expect.poll` evaluates its callback outside the `try` it retries on, so that throw would end the run with a bare action timeout naming nothing — in exactly the regression this spec exists to catch.

The plain-English guard the previous version carried is kept, moved to its source: a pseudo catalog that came out identical to English would otherwise let the assertion pass while proving nothing was localized.

54 of 54 pass across `localization`, `treeView` and `graphDetails` on VS Code and Windsurf, plus the localization rows on Positron.

Two things this deliberately does not fix, tracked separately. The side bar widening helper does nothing at all on Windsurf (#5859) — which is why that editor runs every Graph spec at 300px, and why `treeView`'s "should switch back to a previously viewed commit correctly" is still flaky there. And the shipped locales have no E2E coverage of their own (#5858).


nikolay-1986:
## E2E dispatched on this branch — the target rows are green everywhere, the run's red is not from here

The E2E matrix only runs on `workflow_dispatch`, so it never ran on this PR's own checks. Dispatched it on the branch: [run 35218440065](https://github.com/gitkraken/vscode-gitlens/actions/runs/35218440065) at `a9b5372b1`. It comes back `failure`, and that needs unpacking, because the two rows this PR is about pass on every editor.

**Both localization rows, first attempt, no retries:**

| Editor | `localization.test.ts:156` sidebar panel headers | `localization.test.ts:217` quick pick placeholders |
| --- | --- | --- |
| VS Code | passed, 3.0s | passed, 684ms |
| Windsurf | passed, 8.9s | passed, 931ms |
| Positron | passed, 4.1s | passed, 452ms |

**Against the last dispatch on `main`, [run 35202794148](https://github.com/gitkraken/vscode-gitlens/actions/runs/35202794148) at `d9de42de9`:**

| Job | main | this branch |
| --- | --- | --- |
| VS Code | success — 235 passed, 2 flaky, 4 skipped, 0 failed | failure — 235 passed, 1 failed, 4 skipped, 1 did not run |
| Windsurf | failure — 234 passed, 1 failed, 1 flaky, 4 skipped, 1 did not run | **success — 236 passed, 0 failed, 1 flaky, 4 skipped** |
| Positron | failure — 226 passed, 3 failed, 5 skipped, 7 did not run | failure — 226 passed, 2 failed, 2 flaky, 5 skipped, 6 did not run |

Windsurf's single failure on `main` was `localization.test.ts:128`, the row this PR fixes, and `localization.test.ts:191` never completed there at all — it sits in the same `serial` describe and was dash-marked on every attempt. Both pass now, which is where the 234 → 236 comes from. Positron loses one of its three failures for the same reason.

**The one new red is `graphHeader.test.ts:199`** ("WIP details show the commit-signing indicator when signing is enabled") on VS Code, failing all three attempts. That row was one of the two flaky rows on VS Code in the `main` run above — it was already unstable before this branch, and this time the retries did not save it. Four more `graphHeader` rows in the same job failed once and passed on retry. Its stack ends in `graphHelpers.ts`, around the side bar widening helper, which is #5859.

It is not reachable from this change: the branch touches exactly one file, `tests/e2e/graphHelpers.ts` is untouched, and `graphHeader.test.ts` has no reference to the changed spec.

**Everything else is unchanged from `main`.** `treeView.test.ts:235` is still flaky on Windsurf (failed at 11.1s, passed on retry at 1.3s) — that one needs the widening helper, so it is #5859 rather than this PR. Positron's `welcomeInEditor.test.ts:61` and `graphConflictResolution.test.ts:295` failed on `main` too.

